PR #2, opened 2015-10-22
Regex check on datatype definitions

zaborowska:


hegner:
Could you add the support for namespaces to this pull request?

andresailer:
Sadly, this is hopelessly outdated, the whole checking was refactored into a separate class.
The pattern also doesn't work for things like "unsigned long long".

Thus I am afraid we will just have to close this PR.

PR #11, opened 2016-04-13
fixed infinite recursion and added buitl in type long

gaede:


hegner:
do you can add a test case for the infinite recursion which failed before your fix? There are a few for that in the standard tests and those didn't lead to problems. So the coverage is incomplete.

gaede:
Here it is: simply added a OneToManyRelation to ExampleCluster with type ExampleCluster (sub clusters) and writing them in write.cpp causes seg fault in read test, if patch 523af8a507b2edd821d6b561a0b52d1b867904d7 is not applied

hegner:
Thanks :-)
I don't understand where the whitespace fixes in the examples come from... you didn't adjust the templates or code generator at the places that would be responsible for it...

hegner:
Phew... this addition of yours triggered one of the python tests to fail... expect another fix related to your added test.

PR #12, opened 2016-04-19
Extracode in components

gaede:


hegner:
Thanks. I think we could just change the syntax for defining components to be the same as for datatypes. Only restricting the keywords to members and extracode...

gaede:
Changing the syntax is perfectly fine with me - just did not want to break existing code, this is why I put the Extracode on the same level as the members.
> On 19 Apr 2016, at 14:05, hegner <notifications@github.com> wrote:
> 
> Thanks. I think we could just change the syntax for defining components to be the same as for datatypes. Only restricting the keywords to members and extracode...
> 
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> 

------------------------------------ 
Dr. Frank Gaede
DESY - FLC
 Phone:  +49 40 8998 4382    
 Fax:    +49 40 8994 4382
 E-mail: frank.gaede@desy.de

hegner:
OK. I see. I will handle it then. Could you add a test to what you did? In unittest.cpp "Extracode"? Then I can merge.

gaede:
done.

hegner:
https://sft.its.cern.ch/jira/browse/PODIO-10

PR #15, opened 2016-04-20
optional get/set syntax; operator<(); acces to reference vector

gaede:
- add option -g [--getSyntax] for createing getters and setters
   as getMember()/setMember()

 - add operator<() for Objects and ConstObjects so that they can be stored in maps

 - add getter for reference vector (additional to begin/end iterators)

 - updated (re-generated) example data model accordingly
   ( no change in any other code)

### DIFF
--- a/python/podio_class_generator.py
+++ b/python/podio_class_generator.py
@@ -31,12 +31,14 @@ _text_ = """
 
 class ClassGenerator(object):
 
-  def __init__(self,yamlfile, install_dir, package_name, verbose = True):
+  def __init__(self,yamlfile, install_dir, package_name, verbose = True, getSyntax = False ):
+
     self.yamlfile = yamlfile
     self.install_dir = install_dir
     self.package_name =package_name
     self.template_dir = os.path.join(thisdir,"../templates")
     self.verbose=verbose
+    self.getSyntax=getSyntax
     self.buildin_types = ClassDefinitionValidator.buildin_types
     self.created_classes = []
     self.requested_classes = []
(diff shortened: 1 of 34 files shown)


hegner:
The getter for the reference vector will not work as you made it now. If you look at the iterators you will see why.

gaede:
yep - that was a bit quick on my side (at least it compiled and ran ;-) ) 
it is removed again, which is a petty as in LCIO we use references to vectors rather than iterators ...

hegner:
For storing things in a set a little test would be nice. So that we don't spoil that usage accidentally on refactoring. 

What concerns getting a vector of the references - let's discuss use cases via email. Unfortunately std::array_view didn't find its way into the C++14 standard. Otherwise that would have been the solution to it.

gaede:
Test case added.
I have written a simple vector_view<T> 
 https://stash.desy.de/projects/IL/repos/plcio/browse/source/include/vector_view.h
which could be beautified and added to podio if need.
Currently used as extra code in plcio, eg:
` vector_view<ConstMCParticle> getParents() const { return vector_view<ConstMCParticle>( parents_begin(), parents_end() , parents_size() ) ; }`

We could think about adding an easy way of configuring which methods beyond the standard ones are created in podio, e.g. by including user_template.py.
This way one could choose whether to create an access to the reference vector.

hegner:
Thanks. 

You could have a look at https://github.com/rhysd/array_view . For user templates we have to see where this would lead us to... feature creep and such ;-)

PR #20, opened 2016-05-27
Use source/build directory paths in testing environment

drbenmorgan:
In building/testing podio, I found that the test environment was set up to use the install locations of the libraries/Python modules rather than the expected local build.source tress. This resulted in test failures as the libraries and python test modules were not the located.

This PR changes the test's run environment to use a dynamic loader path set up that points to the build locations of the podio core/test libraries, plus a python path pointing to the python subdir of the source tree. The dynamic loader path environment variable names is also set based on the platform (though still UNIX only). On OS X, `DYLD_FALLBACK_LIBRARY_PATH` is used as this works on El Capitan.

The last commit removes the install of the tests - this no longer seems necessary given the test run fine from the build directory, but can be reverted if there are other use cases.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -17,16 +17,11 @@ file(GLOB headers_utils utilities/*.h)
 
 add_library(TestDataModel SHARED ${sources} ${sources_utils} ${headers} ${headers_utils})
 target_link_libraries(TestDataModel podio )
-install(TARGETS TestDataModel DESTINATION tests)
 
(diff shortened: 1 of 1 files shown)


hegner:
@drbenmorgan - thanks! In fact it was by intention as to check whether the install is actually complete. But you are right, not the cleanest approach.
Please note that I have some major changes in the pipeline - both cleanup and design. So expect some changes in the next two weeks which may be disruptive.

gaede:
Superseded by already merged #51.

PR #21, opened 2016-06-13
Reference bug test

gaede:
Changed read.cpp to throw an exception if the parent-daughter relations are not read back correctly.
Causes 'make test' to fail until the bug is fixed.


hegner:
Thanks!

PR #23, opened 2016-06-17
Fixing OneToManyRelations in namespaces

jlingema:
I had missed to add a test for OneToManyRelations when introducing namespaces and actually did not strip the namespace in the code generator. This PR takes care of both oversights..


jlingema:
superseded by #24

PR #25, opened 2016-07-28
Expose members of a member POD in the opbject

jlingema:
If an object has a component as member, the members of the component are now exposed through setters and getters to the component. This is optional.
Added an options section to the config (reader) to control things like this exposure and the get syntax.


hegner:
Very nice. However there are two things you may want to improve - adding that option to the docs and adding a check for conflicts between pods. Thanks.

hegner:
Thanks :-)

PR #26, opened 2016-09-30
Add ascii io

gaede:
Cleaned up and fixed creation of ostream<<() operators; added an ASCIIWriter and fixed code generation as needed for LCIO ( using the getXXX syntax); added an id() method based on getObjectID to identify objects (e.g. in dumps)


include/podio/ASCIIWriter.h
```diff
@@ -0,0 +1,89 @@
...
+    void finish();
+
+  private:
+    template<typename T>
+    void writeCollection(const std::string& name);
+    // members
+    std::string m_filename;
+    EventStore* m_store;
+
+    std::ofstream* m_file;
```
hegner: why a pointer and not a simple member?

PR #29, opened 2016-10-28
[WIP] CMake build/install tune up

drbenmorgan:
This is a preliminary tune up of the build scripts generally following the layout in hep-sf/tools. There is a slight overlap with PR #20, though that mostly deals with running tests rather than building them.

This shouldn't be thought of as complete, as I encountered a few issues and wanted to discuss them before going further:
1. The `podio_class_generator.py` program is now available as an "imported" target in CMake, so clients of podio can use it via `execute_process/add_custom_command`. However, it, and the remaining Python programs/modules are not yet installed under appropriate directories. 
   1. Would a standard Python layout of programs in `bin` and packages in `lib{64}/pythonX.Y/site-packages` be acceptable?
   2. The directory of templates is moved under `share`, but this means the Python programs cannot find them as `podio_class_generator.py` hard codes in a relative path. I don't know what the best solution would be here. Thoughts are: i) to keep in `share` and generate the relative path from CMake "@@" variable expansion, but that could make build/testing less straightforward,  ii) If the templates are locked into the Python implementation, put them under the Python package directory with lookup via `__file__`. Any recommendations are welcome!
2. I haven't gone further with seeing if a CMake "podio_class_generate" function could be implemented due to the above issue. However, as discussed offline, I do see the issue here, which is that the list of files output by `podio_class_generator.py` are not known a priori. This means it's difficult to write a CMake `add_custom_command/target` pair because the `OUTPUT` argument of the former cannot be filled in. I _think_ it's possible to work around this if `podio_class_generator.py` can gain a `--list-output-files <datamodelyaml>` argument that would print out names/paths of the files that would be created by the input datamodel. Even this doesn't solve that issue, I think it'd a still be a useful addition.
3. At least on my machine, the built Doxygen HTML does not look right. The main menu is fully expanded, and the (Safari) JS console shows an error:
   
   TypeError: $('#main-menu').smartmenus is not a function. (In '$('#main-menu').smartmenus()', '$('#main-menu').smartmenus' is undefined)
   
   I don't know is this is Safari/Doxygen (my local version is 1.8.12) issue, or some clash between podio's headers/footers...

So I don't consider this fully ready for merge yet, but thought opening the PR worthwhile as a place to discuss/resolve the above points


gaede:
superseded w/ #54.

Issue #33, opened 2018-07-09
Overly aggressive environment setup in init.sh


graeme-a-stewart:
Looking at https://github.com/hegner/podio/blob/a55171677ccd2b2b6de15afd828a6aceee61bd4e/init.sh it very aggressively modifies the environment, setting up an old LCG release with the clang compiler on top on any Linux machine that has cvmfs installed.

I understand that it's useful not to try to compile in a broken environment, but surely one only should help after checking that no suitable environment yet exists, e.g., 

- Check the g++ and llvm versions
- Conditionally modify paths only if they are not yet setup

Although honestly I would be happier if we rather added just some advice on setting up the LCG environment *if required* rather than writing magic to do so. (It's odd that ROOT and yaml are explicit, in the README but not the compiler, right?)

hegner:
Hi Graeme,

ROOT etc are explicit in the README as this refers to MacOS. Seems that sentence referring to it is hard to spot in the "Prerequisites" section. 

What about the following approach: doing what you say and giving the init command line arguments for forcing an LCG setup. I want to have something that works out of the box by everyone. And can be adjusted by experts.

graeme-a-stewart:
Hi @hegner - yes, it's rather unclear as the differences in markdown headers are not so big to be easily able to spot. Let me then try to suggest something in a PR and we can iterate.

BTW, the first stanza in the script is really odd - you test `FCCEDM` but then claim `PODIO` is unset.

Another issue is it's unconditionally setting up an SLC6 release, but at least one should be smart with C7 as well.

hegner:
@graeme-a-stewart - sounds good. waiting for the PR then/

graeme-a-stewart:
I think this can be closed now that #41 was merged.

PR #35, opened 2018-07-24
New smarter init.sh

graeme-a-stewart:
Support not setting LCG release if --nolcg is used (or NOLCG env var). Check
for cvmfs and an already setup g++ release in advance. Switch betwen SLC6
and CentOS7 if a release is set up. Check environmemt variables before
adding PODIO elements.

Add a few things to .gitignore for OS X.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+.DS_Store
+.pydevproject
 build
 install
 *pyc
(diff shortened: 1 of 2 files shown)


graeme-a-stewart:
Note that this addresses #33

PR #41, opened 2019-04-09
Build setup fixes

graeme-a-stewart:
Remove FCC specific code from the `init.sh` script (which breaks things for other users). This was discussed with @JavierCVilla and we agreed that any special FCC setup should be controlled by FCC directly (e.g. in the CI).

Made the convenience scripts more generic and updated the instructions in `README.md`.

Factorised the environment setup to `env.sh` as this allows user to set `PODIO` then use the script to set the requisite environment.

BEGINRELEASENOTES
- Improve convenience setup scripts and build instructions
       - remove FCC specific code from init.sh
       - factorise environment setup to env.sh
       - updated README.md

ENDRELEASENOTES

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,14 +54,25 @@ Check that you can now import the yaml module in python.
 
 ## Preparing the environment
 
-Before building and installing this package, and everytime you need to use it, do:
+Full use of PODIO requires you to set the `PODIO` environment variable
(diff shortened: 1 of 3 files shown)


graeme-a-stewart:
This fixes #33.

I suggest "squash and merge" when accepting to keep the master branch tidy and suppress irrelevant fixup commits.

PR #47, opened 2019-06-05
[WIP] implement HDF5 writer in GSoC project

shrey183:
BEGINRELEASENOTES
-  pip install yaml does not work, pip install pyyaml install yaml. 
-  tried to fix the print_report function in podio/python/podio_class_generator.py

ENDRELEASENOTES


gaede:
Hi Shrey
I am a bit confused by this PR: what is it you actually want to achieve/fix ?
The `print_report()` function works fine as is - at least for me. 
It seems to me from some of your changes that you are using python 3 - please use python 2 for this project !
Is this maybe the reason, you had problems with `print_report()` ?
Cheers, Frank.

graeme-a-stewart:
Hi @gaede - just a quick comment that @shrey183 and I did discuss that and he knows that we are using Python2 for PODIO. That said, there's no reason not to use the dialect of Python2 that's most forward compatible with Python3 (which we will need to do at some point). So anytime we touch the code it's a good idea to start to use the `print` function.

gaede:
@shrey183: nice - worked out of the box for me.
In a next step: please try to not create an array of EventInfoData for reading but use the buffer in the
std::vector<EventInfoData> ( use `EventInfoCollection::_getBuffer()` ) and call `EventInfoCollection::prepareAfterRead()` rather than copying the elements individually.

Can you get the sizes of the datasets from the HDF5 file, rather than using hard coded 2 and 4 ?

gaede:
I have committed some minor fixes that I needed to build your code (with llvm on Mac)

gaede:
Superseded by #53.

Issue #50, opened 2019-07-31
Tests failing before install


vvolkl:
I just tried out podio using the new LCG release. Using `/cvmfs/sft.cern.ch/lcg/views/LCG_96/x86_64-centos7-gcc8-opt/setup.sh` 

I get 

```
    Start 1: generate-edm
1/6 Test #1: generate-edm .....................***Failed    0.13 sec
    Start 2: write
2/6 Test #2: write ............................   Passed    3.97 sec
    Start 3: read
3/6 Test #3: read .............................***Failed    2.62 sec
    Start 4: read-multiple
4/6 Test #4: read-multiple ....................***Failed    2.39 sec
    Start 5: pyunittest
5/6 Test #5: pyunittest .......................***Failed    0.18 sec
    Start 6: unittest
6/6 Test #6: unittest .........................   Passed    0.04 sec
```

The errors are of the type

```
...
2: Error in <TTree::Branch>: The pointer specified for CollectionIDs is not of a class or type known to ROOT
...
3: Error in <TTree::SetBranchAddress>: unknown branch -> CollectionIDs
3: reading event 0
3: 
3:  *** Break *** segmentation violation
```

and for the pyunittest

```
5: ImportError: Start directory is not importable: '/usr/local/python'
5/6 Test #5: pyunittest .......................***Failed    0.09 sec


```

vvolkl:
I was mistaken; this is not connected to LCG_96 but to the way the tests are set up, which require `make install` to be run before the tests. Will be good to change this.

PR #51, opened 2019-08-07
Do not install tests

JavierCVilla:
BEGINRELEASENOTES
Do not install tests
- Allow tests to run after the build phase
- Paths have been modified to point to binary or source directories
- Before, tests had to be run only after running make install
- Test are not installed anymore
- Fail tests if any error is reported (ROOT Interpreter error may not be considered by CMake)

ENDRELEASENOTES

This addresses #50

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -1,5 +1,5 @@
 include_directories(
-	${CMAKE_SOURCE_DIR}/include
+  ${CMAKE_SOURCE_DIR}/include
   ${CMAKE_CURRENT_SOURCE_DIR}/datamodel
 )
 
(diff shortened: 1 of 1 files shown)


gaede:
@JavierCVilla: nice - but let's wait until the CI is running again, to check if it works with the CI as well...

JavierCVilla:
@gaede sure, no problem

gaede:
@JavierCVilla: the tests fail w/ a  CMake issue now - can you have a look ?

JavierCVilla:
@gaede I have just fixed, but now the gcc test fails because it cannot find `clang-format` in the environment, I believe my changes are not related to that issue:

https://travis-ci.org/AIDASoft/podio/jobs/569757159#L1010

JavierCVilla:
@gaede Did you have time to see the error?

gaede:
@petricm our tests fail for gcc because of a missing `clang-format `:
https://travis-ci.org/AIDASoft/podio/jobs/569757159 
Could this be installed in the CI environment or do we need to remove this from the tests for gcc (it works for llvm) ?

petricm:
The gcc LCG view does come with clang tools, on which the CI is based.

I think it would be best to remove this test for a gcc enviroment, otherwise you are telling people/users that just have gcc and no clang that tests don't pass and that something is wrong, but in reality only one python file is not supported in such a environment.

PR #57, opened 2019-09-26
Allow numbers in namespace and class names

### LINKED ISSUE #56: ROOT Writer does not support numbers in namespaces
See https://github.com/HSF/EDM4HEP/issues/9

gaede:
BEGINRELEASENOTES
- allow for numbers in namespace and class names
      - generate type name string for CollectionBase::getValueTypeName()  
      - fixes #56 

ENDRELEASENOTES

### DIFF
--- a/templates/Collection.h.template
+++ b/templates/Collection.h.template
@@ -74,6 +74,9 @@ public:
   /// number of elements in the collection
   int size() const override final ;
 
+  /// fully qualified type name of elements - with namespace
+  std::string getValueTypeName() const override { return std::string("${classname}") ; } ;
+
   /// Returns the const object of given index
   const ${name} operator[](unsigned int index) const;
   /// Returns the object of a given index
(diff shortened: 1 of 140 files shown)


templates/Collection.h.template
```diff
@@ -74,6 +74,9 @@ public:
   /// number of elements in the collection
   int size() const override final ;
 
+  /// fully qualified type name of elements - with namespace
+  std::string getValueTypeName() const override { return std::string("${classname}") ; } ;
```
vvolkl: `${classname}` could also be a made a static member, but since I don't expect many instances of a collection, it is more straightforward like this, I think.

vvolkl:
There are a lot of whitespace changes in this PR. I am not sure where those are coming from. Since it only concerns the auto-generated source files I don't have a strong opinion on that.

gaede:
> There are a lot of whitespace changes in this PR. 
Yes, indeed. I have seen this before - not entirely sure how to fix this easily as this is in the templates and the python code...

Issue #58, opened 2019-09-27
EDM generation broken after install


gaede:
As a result of #54 the EDM egenration does not work anymore for external  clients (EDM4Hep).
The reason is the hardcoded template_dir path "../templates" that is no longer available in the install.  @drbenmorgan  can you have a look at this ?

Current workaround:  
```
cd INSTALL_DIR
ln -s SOURCE_DIR/templates .
```

drbenmorgan:
Hi @gaede, apologies for that, I'll get a fix Pull Request prepared.

drbenmorgan:
Having had a quick look at this, there are a several ways to resolve this so thought better to discuss here before fixing in a Pull Request:

1. Restore original behaviour, so the install tree would look like:
   ```
   +- CMAKE_INSTALL_PREFIX/
      +- python/
      +- templates/
   ```
   The simplest, but will cause problems for packagers - e.g. say the install prefix is `/usr/local`, then `/usr/local/templates` is non-standard and could easily clash (I'm ignoring the `python dir` for now!).

2. Move `templates` as a subdirectory of `python`, so the relative path becomes "./templates". This is probably the best solution if the template files aren't for general access/use, i.e. only podio's python code will read them.

3. Update the relative path on install using known install locations to construct a relative path.

I'll implement whatever solution is preferred for podio. My personal recommendation would be the second one if the template files are only intended for internal use by podio's class generation.

gaede:
Hi @drbenmorgan, I'd also vote for option 2). Can't think of any other use cases for the templates than from our python generators.

gaede:
PS: shouldn't the `INSTALL_DIR/python` have a `podio` subdirectory in case someone wants to install to `/usr/local/` ?

drbenmorgan:
Thanks @gaede, I'll go ahead and implement 2). 

You're right on the `INSTALL_DIR/python` issue, but `python/podio` also seems odd. Looking at the files in `python/`, there are some programs *and* modules

- programs: `podio_class_generator.py`, `podio_create_package.py`
- modules: `podio_config_reader.py`, `podio_templates.py`, `EventStore.py`

The former should probably go under a `bin/` subdirectory, but am not sure about the latter. I'd guess `lib/pythonX.Y/site-packages/podio`. Maybe just install them all to `bin` for now?

drbenmorgan:
Ah scratch that install to `bin`, since that would stick `templates` under that which wouldn't be good.

gaede:
[...]

PR #59, opened 2019-09-27
Move templates under python

drbenmorgan:
The changes introduced by #54 caused installs to fail due to the change in location of the templates directory relative to python/.

Following discussion in #58, implement simplest possible change to fix by moving templates inside the python directory. Retain this layout at install time, adjusting the `template_dir` variable in
podio_class_generator.py accordingly.

Fixes: #58

BEGINRELEASENOTES
- move `templates` inside the `python` directory when installing
      - fixes: #58

ENDRELEASENOTES


gaede:
Thanks for fixing this. Please write in the future some short comment inside the BEGINRELEASENOTES/ENDRELEASENOTES. These will show up in the ReleaseNots.md when we make release tags.

PR #63, opened 2019-10-07
Fix for C++17

peremato:
BEGINRELEASENOTES
- Fix to avoid deprecated STL classes in C++17  

ENDRELEASENOTES

### DIFF
--- a/tests/catch.hpp
+++ b/tests/catch.hpp
@@ -3450,7 +3450,7 @@ namespace Catch {
     };
 
     class DebugOutStream : public IStream {
-        std::auto_ptr<StreamBufBase> m_streamBuf;
+        std::unique_ptr<StreamBufBase> m_streamBuf;
         mutable std::ostream m_os;
     public:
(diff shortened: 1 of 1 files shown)


vvolkl:
Hi Pere,
I just checked and indeed podio includes a hardcopy of an old version (1.4) of the `catch` unit testing library (https://github.com/catchorg/Catch2), that does not claim to support anything beyond C++03. Interestingly it builds fine with C++17 on centos7/gcc8.3 for me. The proper thing would probably be to update `catch` (or check if  a library included in the lcg releases would be good replacement). But if this patch works we should merge it I think, Frank @gaede!

vvolkl:
Note also #64 which allows you to skip building the tests all together.

peremato:
@vvolkl The fact that it works with C++17 on centos7/gcc8.3 is known. GCC has not removed C++17 deprecated STL functions and classes, while Clang on MacOS has done it.

PR #64, opened 2019-10-14
Add CMake option for building tests

vvolkl:
BEGINRELEASENOTES
- Add `BUILD_TESTS` CMake option for building tests 

ENDRELEASENOTES

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,7 +53,6 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CPP_STANDARD_FLAGS}")
 
 #--- Declare options -----------------------------------------------------------
 option(CREATE_DOC "Whether or not to create doxygen doc target." OFF)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -52,7 +52,8 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CPP_STANDARD_FLAGS}")
 
 
 #--- Declare options -----------------------------------------------------------
-option(CREATE_DOC "Whether or not to create doxygen doc target.")
+option(CREATE_DOC "Whether or not to create doxygen doc target." OFF)
+option(BUILD_TESTS "Whether or not to build unit and integration tests." ON)
```
andresailer:
```
include(CTest)
```
already defines an option `BUILD_TESTING`, which should be used for this purpose

vvolkl: Wasn't aware of this variable, that is better indeed, thanks.

PR #65, opened 2020-02-12
podio_class_generator: only write files if their content changed

andresailer:
Avoids spurious re-compilation
BEGINRELEASENOTES
- podio_class_generator:  only write files if their content changed

ENDRELEASENOTES

### DIFF
--- a/python/podio_class_generator.py
+++ b/python/podio_class_generator.py
@@ -51,8 +51,7 @@ class ClassGenerator(object):
             print "       Please make sure it is in the PATH."
             self.clang_format = []
             return
-        self.clang_format = [cformat_exe, "-i",  "-style=file", "-fallback-style=llvm"]
-
+        self.clang_format = [cformat_exe, "-style=file", "-fallback-style=llvm"]
 
     def process(self):
         self.reader.read()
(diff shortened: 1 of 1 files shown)


python/podio_class_generator.py
```diff
@@ -1034,17 +1033,22 @@ def prepare_vectorized_access(self, classname,members ):
...
-        open(fullname, "w").write(content)
         if self.clang_format:
-          subprocess.call(self.clang_format + [fullname])
+          cfproc = subprocess.Popen(self.clang_format, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
+          content = cfproc.communicate(input=content)[0]
+        existing_content = ''
+        if os.path.exists(fullname):
+          existing_content = open(fullname, 'r').read()
+        if existing_content != content:
+          open(fullname, 'w').write(content)
```
vvolkl: These lines could be slightly more "pythonic" with a `with` statement: 
```python
with open(fullname, "r+") as _f:
  existing_content = _f.read()
  if existing_content != content:
    _f.write(content)
```

andresailer:
```python
#!/usr/bin/env python
content ='foo bar'
fullname = 'temp.txt'
with open(fullname, "r+") as _f:
  existing_content = _f.read()
  if existing_content != content:
    _f.write(content)
```
Errors if temp.txt does not exist
```
$ python testRead.py
  File "testRead.py", line 6, in <module>
    with open(fullname, "r+") as _f:
IOError: [Errno 2] No such file or directory: 'temp.txt'
```
So we would still need one path for exists, one without.

vvolkl: That's true, I overlooked that in one case the file won't exist yet. Taking that in consideration I think the code is best as it is!

andresailer: There are many things to be cleaned up, especially in view of python3 compatibility (#66)

PR #67, opened 2020-02-14
Preserve order of members from yaml

andresailer:
Thanks to https://stackoverflow.com/questions/5121931/in-python-how-can-you-load-yaml-mappings-as-ordereddicts/21048064#21048064

BEGINRELEASENOTES
- Change Yaml loader to preserve order of members as given in the yaml

ENDRELEASENOTES


andresailer:
This should fix that -Wreorder in edm4hp ObjectID

PR #68, opened 2020-02-14
[CMake] remove need for directory properties set includes, links opti…

andresailer:
…ons directly for targets, add PODIO_GENERATE_DICTIONARY macro to allow this

This is somewhat temporary until root supports the use of targets directly.
These changes are necessary to allow one to use properties of targets instead of directory properties. It is also necessary if one has include_directories properties that contain a list of include directories, not just individual entries.

BEGINRELEASENOTES
- Add podioMacros.cmake to contain PODIO_GENERATE_DICTIONARY

ENDRELEASENOTES

### DIFF
--- a/cmake/podioMacros.cmake
+++ b/cmake/podioMacros.cmake
@@ -1,0 +1,95 @@
+#---------------------------------------------------------------------------------------------------
+#---PODIO_GENERATE_DICTIONARY( dictionary headerfiles SELECTION selectionfile OPTIONS opt1 opt2 ...
+#                              DEPENDS dependency1 dependency2 ...
+#                              USES target1 target2 ...
+#                            )
+#
+#  USES takes a list of targets. From the targets the properties INCLUDE_DIRECTORIES
+#     and COMPILE_DEFINITIONS are used to create the appropriate -I and -D flags
+#
+#---------------------------------------------------------------------------------------------------
+macro(PODIO_GENERATE_DICTIONARY dictionary)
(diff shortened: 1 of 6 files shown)


cmake/podioMacros.cmake
```diff
@@ -0,0 +1,95 @@
+#---------------------------------------------------------------------------------------------------
+#---REFLEX_GENERATE_DICTIONARY( dictionary headerfiles SELECTION selectionfile OPTIONS opt1 opt2 ...
```
vvolkl:
```suggestion
#---PODIO_GENERATE_DICTIONARY( dictionary headerfiles SELECTION selectionfile OPTIONS opt1 opt2 ...
```
This comment refers to `PODIO_GENERATE_DICTIONARY` or should remind the use of `REFLEX...`?

andresailer: Thanks, I modified this line as well. Or did you mean something else?

gaede:
Should this be merged ?

cmake/podioMacros.cmake
vvolkl: No that's all I meant. Thanks!

vvolkl:
Yes, looks good to me!

hegner:
@andresailer - do you know whether the point is still relevant or whether it would be safe to go back to default ROOT?

andresailer:
Give the ROOT REFLEX_GENERATE_DICTIONARY a try and see what happens...

hegner:
seems I have to move to ROOT_GENERATE_DICTIONARY anyhow now :-(

PR #69, opened 2020-02-19
Split root dependencies into separate library

andresailer:
BEGINRELEASENOTES
-  Move all Root dependencies (RootReader, RootWriter) of the podio Library into podioRootIO, rename podioDict to podioDict

ENDRELEASENOTES

### DIFF
--- a/.podio-ci.d/compile_and_test.sh
+++ b/.podio-ci.d/compile_and_test.sh
@@ -6,7 +6,7 @@ cd /podio
 source init.sh
 mkdir build install
 cd build
-cmake -DCMAKE_INSTALL_PREFIX=../install -DCMAKE_CXX_STANDARD=17 .. && \
-make -j4 && \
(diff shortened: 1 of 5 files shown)


andresailer:
Will try to split the Root and non root dependent code into separate libraries

andresailer:
Should we have: podio (anything not linking against root), podioRootIO (ROOT Reader and Writer), and podioDict (basically rootDict.cxx file created from rootcling/genreflex)?
Then in the future we might have in addition podioHDF5IO etc.

hegner:
Sounds good to me.

andresailer:
I split the libraries into podio, podioDict and podioRootIO, updated the PODIO_GENERATE_DICTIOARY now that the PR in ROOT was merged (https://github.com/root-project/root/pull/5019)

andresailer:
Review please

Issue #49, opened 2019-07-12
Support for collection-level metadata


vvolkl:
As discussed in the context of EDM4HEP support for some metadata, or any information that is saved together with a collection would be useful. For example, this could be used to save a readout string (`system:4,layer:5,module:18,x:-15,z:-15`) along with a dd4hep cellID. My first idea was to simply add a member to the collection template. @gaede , maybe you can comment how this is done in LCIO? Cheers

andresailer:
LCIO collections have a LCParameters member to hold arbitrary collection parameters
http://lcio.desy.de/v02-09/doc/doxygen_api/html/classEVENT_1_1LCCollection.html#ac60df52bb014939521d2cc72ffe60223

http://lcio.desy.de/v02-09/doc/doxygen_api/html/classEVENT_1_1LCParameters.html

andresailer:
Having a look at HepMC3: there are Attributes that can be attached to GenEvent (or other types)
http://hepmc.web.cern.ch/hepmc/group__attributes.html

PR #71, opened 2020-03-02
[WIP] Add "RunParameters" branch to metadata tree

vvolkl:
This adds a branch with a  `std::map<std::string, std::string>` to the metadata tree and adds the necessary methods to the EventStore in order that MetaData can be saved. Would be a straightforward,  but a bit minimalistic solution to #49. 
BEGINRELEASENOTES
- Add RunParameters Branch to Metadata Tree
ENDRELEASENOTES

### DIFF
--- a/include/podio/EventStore.h
+++ b/include/podio/EventStore.h
@@ -5,6 +5,7 @@
 #include <vector>
 #include <type_traits>
 #include <set>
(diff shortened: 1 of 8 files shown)


vvolkl:
Superseded by #92

PR #73, opened 2020-03-12
Always test the up-to-date datamodel

andresailer:
This adds logic to cmake to always create the TestDataModel.

Edited:
There is the execute_process which is run in the configuration step (when you call `cmake ..`) and the `add_custom_target` call is run during the build step (`make`), this is always run, and before the datamodel is compiled. Because the class_generator only writes files if their content changed neither make nor ninja will superfluously compile objects. Of course if an additional class is added to the yaml `cmake` has to be called explicitly, because of the use of "GLOB" (but this was also the case before).

As far as I understand, the difference to the behaviour of make when adding ${headers} and ${sources} as OUTPUT to add_custom_command is the lack of dependency of these files on datalayout.yaml, so the age of that file does not matter, only the age of the created header or cc files.


tests/CMakeLists.txt
```diff
@@ -5,6 +5,24 @@ foreach( _conf ${CMAKE_CONFIGURATION_TYPES} )
...
+  COMMAND python ${CMAKE_SOURCE_DIR}/python/podio_class_generator.py tests/datalayout.yaml tests datamodel
+  WORKING_DIRECTORY ${CMAKE_SOURCE_DIR})
+
+# This ensures that we re-run the class generator if the yaml file is touched
+# the command is run if datalayout.yaml is newer then selection.xml
+add_custom_command(OUTPUT ${CMAKE_SOURCE_DIR}/tests/src/selection.xml  # we specify one output file so this doesn't always run
+  COMMENT "Re-Creating TestDataModel"
+  MAIN_DEPENDENCY datalayout.yaml
+  # This removal ensures selection.xml is newer than datalayout.yaml, other wise this command always runs
+  COMMAND ${CMAKE_COMMAND} -E remove ${CMAKE_SOURCE_DIR}/tests/src/selection.xml
```
andresailer: While this re-runs the class generator when the yaml file changes, it comes to late in the process.
Changing (bringing the GLOB lines above)
```suggestion
  COMMAND ${CMAKE_COMMAND} -E remove ${CMAKE_SOURCE_DIR}/tests/src/selection.xml ${sources} ${headers}
```
Will make `make` recompile everything because the object files would be older than the yaml file. `ninja` is better in this regard.

hegner: Great! Looking forward to testing it, once the WIP flag is gone

andresailer: No, it is not great at all :) My goal was to only re-compile what is needed...

andresailer:
Better implementation with add_custom_command instead of add_custom_target, see the opening post of this PR (https://github.com/AIDASoft/podio/pull/73#issue-387375326)
This PR includes the commits from #74 to reduce the verbosity at rebuilding

andresailer:
Rebased

hegner:
I think this will work nicely for the testing.
Would it make sense to provide a cmake macro w/ equivalent feature for users? Which e.g. edm4hep could use directly?

andresailer:
Moved things into a macro,
Had to add the python install location to pocioConfig to avoid things like
``` cmake
    COMMAND python ${podio_DIR}/../../../python/podio_class_generator.py ../edm4hep.yaml ./ edm4hep
```

hegner:
Great! Thanks for the quick addition

andresailer:
You're welcome! 
I added something to the first post for the release notes as well.

PR #74, opened 2020-03-17
ClassGenerator: Fix exception when called with --quiet

andresailer:
BEGINRELEASENOTES
- Fix exception when podio_class_generator is called with -q/--quiet
- Printout about includes only in verbose mode
- Do not print warnings in quiet mode
- Print each warning only once

ENDRELEASENOTES

### DIFF
--- a/python/podio_class_generator.py
+++ b/python/podio_class_generator.py
@@ -83,13 +83,14 @@ class ClassGenerator(object):
             # self.create_PrintInfo(name, components)
 
     def print_report(self):
-        if self.verbose:
-            pkl = open(os.path.join(thisdir, "figure.txt"))
-            figure = pickle.load(pkl)
-            text = _text_ % (self.yamlfile,
-                             len(self.created_classes),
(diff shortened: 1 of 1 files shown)


python/podio_class_generator.py
```diff
@@ -489,7 +490,8 @@ def create_class(self, classname, definition):
         if "includes" in extra:
             extraIncludes = set(extra["includes"].split('\n'))
             datatype["includes"].update(extraIncludes)
-            print " ***** adding includes : " ,  extraIncludes , "to" ,  datatype["includes"]
+            if self.verbose:
```
andresailer: Should this just be dropped?

andresailer:
These changes make the class generator completely quiet if called with -q/--quiet
Review please.

python/podio_class_generator.py
hegner: I think this was just debugging during development. It should be dropped indeed

PR #75, opened 2020-03-24
2-3-Compatibility of Python Code

vvolkl:
BEGINRELEASENOTES
- Fixes in python code for 2-3 compatibility

ENDRELEASENOTES

### DIFF
--- a/python/EventStore.py
+++ b/python/EventStore.py
@@ -1,3 +1,8 @@
+
+
+from __future__ import absolute_import, unicode_literals, print_function
+from six.moves import range
+
 from ROOT import gSystem
 gSystem.Load("libpodioRootIO")
 from ROOT import podio
@@ -6,7 +11,7 @@ from ROOT import podio
 def iterator(self):
     '''dynamically added iterator'''
     entries = self.size()
-    for entry in xrange(entries):
+    for entry in range(entries):
         yield self.at(entry)
 
 
@@ -99,7 +104,7 @@ class EventStore(object):
         '''
         for nev, store in self.stores:
             self.current_store = store
-            for entry in xrange(store.getEntries()):
+            for entry in range(store.getEntries()):
                 yield store
                 store.endOfEvent()
 
--- a/python/EventStore.py
+++ b/python/EventStore.py
@@ -41,7 +41,7 @@ class EventStore(object):
                       files matching a wildcard pattern.
            treename: not used at the moment
         '''
-        if isinstance(filenames,str):
+        if isinstance(filenames, str):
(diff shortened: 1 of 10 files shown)


python/EventStore.py
```diff
@@ -1,3 +1,6 @@
+
+from past.builtins import xrange
```
petricm: I would rather not import the past to the future but the future to the past. So not give python 3 features from python 2, but rather assume it's python 2 and give it python 3 features. The code should read as Python 3 (so no visible xrange) so it's "easier" to drop python 2.

petricm:
Please check the gihub actions for python, you still have several problems

python/EventStore.py
vvolkl: Good point. Done.

petricm:
Also it would be good to install the python stuff under `lib[64]/python${Python_VERSION_MAJOR}.${Python_VERSION_MINOR}/site-packages`

vvolkl:
> Also it would be good to install the python stuff under `lib[64]/python${Python_VERSION_MAJOR}.${Python_VERSION_MINOR}/site-packages`

Agreed, as there are actually a few issues to fix with the python code organization. I'd propose I do it in a separate PR.

python/podio_class_generator.py
```diff
@@ -27,1083 +32,1107 @@
...
+    ostream_implementation += '  << std::endl;\n'
+    ostream_implementation = ostream_implementation.replace("{header_string}", ostream_header_string)
+
+    # ----------------------------------------------------------------------
+
+    refmembers = definition["OneToOneRelations"]
+    refvectors = definition["OneToManyRelations"]
+    vectormembers = definition["VectorMembers"]
+    nOfRefVectors = len(refvectors)
+    n
```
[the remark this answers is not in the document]
vvolkl: Interesting bug indeed. Something about the diff pattern matching?

setup.cfg
```diff
@@ -21,6 +23,10 @@ ignore =
        E125
        # continuation line over-indented for visual indent
        E127
+       # line break after binary operator (contradicts W503) 
+       W504
+       # module level import not at top of file (poblem with ROOT)
+       E402
```
petricm: E402 what if you just flag the lines of ROOT as false positives

vvolkl: Done!

vvolkl:
@petricm good point, I just switched everything to 4 spaces now.

vvolkl:
As discussed in the EDM4hep meeting, I changed this to 2 spaces indent. @gaede should be good to merge now!

PR #80, opened 2020-03-27
Add macOS workflow to tests

petricm:
BEGINRELEASENOTES
- add a macOS test to GitHub actions

ENDRELEASENOTES
But for a proper implementation I need https://github.com/actions/virtual-environments/issues/598 now it downloads root from the root webpage

### DIFF
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -11,6 +11,7 @@ jobs:
       run: |
         curl -O https://root.cern/download/root_v6.20.02.macosx64-10.15-clang110.tar.gz
         tar xf root_v6.20.02.macosx64-10.15-clang110.tar.gz
+        pip install -r requirements.txt
     - name: Compile and test
       run: |
         source root/bin/thisroot.sh
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -1,4 +1,4 @@
-name: compile
+name: mac
 
 on: [push, pull_request]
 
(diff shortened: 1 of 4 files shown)


andresailer:
missing `yaml` python package

andresailer:
Also this clearly points that a check for the return value of podio_class_generator should be added to abort if this fails.

.github/workflows/mac.yml
```diff
@@ -0,0 +1,23 @@
...
+jobs:
+  test:
+    runs-on: macos-10.15
+    steps:
+    - uses: actions/checkout@v2
+    - name: Download ROOT
+      run: |
+        curl -O https://root.cern/download/root_v6.20.02.macosx64-10.15-clang110.tar.gz
+        tar xf root_v6.20.02.macosx64-10.15-clang110.tar.gz
+        sudo pip install yaml
```
andresailer:
```suggestion
        sudo pip install pyyaml
```

petricm: I am checking what else is missing, this needs a requirements.txt file

petricm: ok all imports 
```py
set(['string', 're', 'optparse', 'ROOT', 'unittest', 'subprocess', 'yaml', 'podio_config_reader', 'collections', 'EventStore', 'pickle', 'podio_templates', 'copy', 'os'])
```
only `yaml` is not in the basic python or ROOT or podio

PR #91, opened 2020-05-25
RootIO: ensure podioDict is linked also when linker has as-needed

### LINKED ISSUE #90: libpodioDict.so is not always properly linked
When building podio `libpodioDict.so` is not always properly linked into executables or other shared libraries, like `libpodioRootIO.so` or the test executables. I am not sure whether it is a podio problem or something that is more closely related to some subtle cmake issue, as it e.g. occurs on Ubuntu 18.04, but not on Fedora 29 (or on lxplus).

I have prepared an [Ubuntu docker image](https://hub.docker.com/repository/docker/tmadlener/ubuntu-podio-build-problem) where the problem occurs and a [Fedora docker image](https://hub.docker.com/repository/docker/tmadlener/fedora-podio-build-problem), where it does not.

To reproduce the problem in the Ubuntu container pull it and run it interactively, then do
```bash
bash build_podio.sh
```

This will build podio and run `tests/write > /dev/null` as well as `ldd tests/write` in the podio build directory. And result in output along the following lines

```
[...]
Error in <TTree::Branch>: The class requested (vector<podio::ObjectID>) for the branch "WithNamespaceRelationCopy#1" is an instance of an stl collection and does not have a compiled CollectionProxy. Please generate the dictionary for this collection (vector<podio::ObjectID>) to avoid to write corrupted data.
Error in <TTree::Branch>: The pointer specified for CollectionIDs is not of a class or type known to ROOT
	

        linux-vdso.so.1 (0x00007ffdfc9e1000)
	libTestDataModel.so => /home/hsf/podio/build/tests/libTestDataModel.so (0x00007f0b7867d000)
	libpodioRootIO.so => /home/hsf/podio/build/src/libpodioRootIO.so (0x00007f0b78472000)
	libpodio.so => /home/hsf/podio/build/src/libpodio.so (0x00007f0b78269000)
	libCore.so.6.20 => /home/hsf/root/lib/libCore.so.6.20 (0x00007f0b77b69000)
[...]
```

Repeating the same steps in the Fedora container leads to the following output, where only `ldd tests/write` is now present, since `tests/write > /dev/null` runs without problems.
```
[...]

        linux-vdso.so.1 (0x00007ffe1b542000)
	libTestDataModel.so => /home/hsf/podio/build/tests/libTestDataModel.so (0x00007f925746c000)
	libpodioRootIO.so => /home/hsf/podio/build/src/libpodioRootIO.so (0x00007f925745f000)
	libpodioDict.so => /home/hsf/podio/build/src/libpodioDict.so (0x00007f9257454000)
	libpodio.so => /home/hsf/podio/build/src/libpodio.so (0x00007f9257449000)
	libTree.so.6.20 => /home/hsf/root/lib/libTree.so.6.20 (0x00007f92572b5000)
[...]
```

The most important difference is that now `libpodioDict.so` is present.

To me it is currently unclear from where this difference arises, but it seems to be unrelated to the exact `gcc` version that is used on Ubuntu and also on how `root` has been built.

The only hint towards a problem can be seen when using `cmake-3.10` ~on Ubuntu~, where the following warning appears at the `cmake` stage

```
You have called ADD_LIBRARY for library podioDict without any source files. This typically indicates a problem with your CMakeLists.txt file
```

This message does not appear with newer cmake versions (tested with 3.14 and 3.17). It is also inconsequential as the outcome remains unchanged.

andresailer:
@tmadlener, can you test this, please

BEGINRELEASENOTES
- Ensure podioRootIO is linked against podioDict even when linker uses as-needed by default, fixes #90 

ENDRELEASENOTES

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -46,9 +46,38 @@ PODIO_GENERATE_DICTIONARY(podioDict ${headers} SELECTION selection.xml
 # prevent generating dictionary twice
 set_target_properties(podioDict-dictgen PROPERTIES EXCLUDE_FROM_ALL TRUE)
 
+# check if the linker supports as needed or not
+set(STORE_REQUIRED_FLAGS ${CMAKE_REQUIRED_FLAGS})
+include(CheckCCompilerFlag)
+set(CMAKE_REQUIRED_FLAGS "-Wl,--no-as-needed")
+check_c_compiler_flag("" TEST_NO_AS_NEEDED)
+if(TEST_NO_AS_NEEDED)
+    message(STATUS "Linker supports: -Wl,--no-as-needed")
+else()
+    message(STATUS "Linker does not support: -Wl,--no-as-needed")
+endif()
+set(CMAKE_REQUIRED_FLAGS "-Wl,--as-needed")
+check_c_compiler_flag("" TEST_AS_NEEDED)
+if(TEST_AS_NEEDED)
+    message(STATUS "Linker supports: -Wl,--as-needed")
+else()
+    message(STATUS "Linker does not support: -Wl,--as-needed")
+endif()
+set(CMAKE_REQUIRED_FLAGS ${STORE_REQUIRED_FLAGS})
+
 target_sources(podioDict PRIVATE podioDict.cxx)
 # ROOT IO without the Dict does not work, so we link RootIO against the Dict
-target_link_libraries(podioRootIO PUBLIC podioDict)
+# need to disable as needed if linker supports it
+if(TEST_NO_AS_NEEDED AND TEST_AS_NEEDED)
+  target_link_libraries(podioRootIO
+    PRIVATE
+      -Wl,--push-state,--no-as-needed podioDict -Wl,--pop-state
+    INTERFACE
+      podioDict
+     )
+ else()
+   target_link_libraries(podioRootIO PUBLIC podioDict)
+ endif()
 
 # Install the Targets and Headers
 install(TARGETS podio podioDict podioRootIO


tmadlener:
Thanks @andresailer . I have tested it in the two containers mentioned in #90 and additionally, locally on my Ubuntu machine and on `lxplus. For me it works everywhere.

andresailer:
push-state/pop-state might be better to not affect the behaviour, but for this line here, I don't think it matters?

src/CMakeLists.txt
```diff
@@ -46,9 +46,38 @@ PODIO_GENERATE_DICTIONARY(podioDict ${headers} SELECTION selection.xml
...
+set(CMAKE_REQUIRED_FLAGS ${STORE_REQUIRED_FLAGS})
+
 target_sources(podioDict PRIVATE podioDict.cxx)
 # ROOT IO without the Dict does not work, so we link RootIO against the Dict
-target_link_libraries(podioRootIO PUBLIC podioDict)
+# need to disable as needed if linker supports it
+if(TEST_NO_AS_NEEDED AND TEST_AS_NEEDED)
+  target_link_libraries(podioRootIO
+    PRIVATE
+      -Wl,--no-as-needed podioDict -Wl,--no-as-needed
```
andresailer, line 74: Of course I meant to write
```suggestion
      -Wl,--no-as-needed podioDict -Wl,--as-needed
```
But then it probably wouldn't work, because cmake re-arranges the order of flags and libraries...
So, just set no-as-needed and don't bother with resetting?
(push-state/pop-state has the same problem)

andresailer: Nevermind, the problem was using PUBLIC instead of INTERFACE in the line below, then podioDict was twice in the list of libraries and de-duplication made it look like re-arranging was happening.

tmadlener:
The latest version with `push-state` and `pop-state` still works for me.

PR #92, opened 2020-05-27
Add meta data for collections, events and runs

gaede:
BEGINRELEASENOTES
- implement reading/writing of meta data for runs, events and collections
     - based on GenericParameters that hold named parameters of type `int, float, string` or vectors if these (copied from `lcio::LCParameters`)
     - meta data for the three types is always written 
     - it is read only on request
- example for writing:
```
   auto& evtMD = store.getEventMetaData() ;
    evtMD.setValue( "UserEventWeight" , (float) 100.*i ) ;
```
- example for reading:
```
  auto& evtMD = store.getEventMetaData() ;
  float evtWeight = evtMD.getFloatVal( "UserEventWeight" ) ;

```
- addresses #49 

ENDRELEASENOTES

### DIFF
--- a/src/ROOTWriter.cc
+++ b/src/ROOTWriter.cc
@@ -41,6 +41,11 @@ namespace podio {
     m_metadatatree->Branch("CollectionIDs",m_store->getCollectionIDTable());
     m_metadatatree->Fill();
 
+    m_colMDtree->Branch("colMD", "std::map<int,podio::GenericParameters>", m_store->getColMetaDataMap() ) ;
+    m_colMDtree->Fill();
+    m_runMDtree->Branch("runMD", "std::map<int,podio::GenericParameters>", m_store->getRunMetaDataMap() ) ;
+    m_runMDtree->Fill();
+
     m_file->Write();
     m_file->Close();
   }
--- a/src/ROOTWriter.cc
+++ b/src/ROOTWriter.cc
@@ -21,7 +21,7 @@ namespace podio {
     m_runMDtree(new TTree("run_metadata", "Run metadata tree"))
   {
 
-    m_evtMDtree->Branch("evtMD", "GenericParameters", m_store->getEventMetaData() ) ;
+    m_evtMDtree->Branch("evtMD", "GenericParameters", m_store->eventMetaDataPtr() ) ;
   }
 
   ROOTWriter::~ROOTWriter(){
--- a/src/ROOTWriter.cc
+++ b/src/ROOTWriter.cc
@@ -15,8 +15,14 @@ namespace podio {
     m_store(store),
     m_file(new TFile(filename.c_str(),"RECREATE","data file")),
     m_datatree(new TTree("events","Events tree")),
-    m_metadatatree(new TTree("metadata", "Metadata tree"))
-  {}
+    m_metadatatree(new TTree("metadata", "Metadata tree")),
+    m_evtMDtree(new TTree("evt_metadata", "Event metadata tree")),
+    m_colMDtree(new TTree("col_metadata", "Collection metadata tree")),
+    m_runMDtree(new TTree("run_metadata", "Run metadata tree"))
+  {
+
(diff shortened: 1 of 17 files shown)


hegner:
Hi Frank! Thanks for that! I have two major comments:

1. In case there is no meta data tree - do we want to return some empty object instead when getting the metadata? That would be more foolproof for users and allow reference semantics
2. Different experiments have different ideas whether an event weight is data or meta data. Thus I would not put event weight as meta data example.

gaede:
Hi Benedikt,
1. I am using effectively value semantics, i.e. the meta data objects always exist but are empty when nothing has been set/written. Only returning a pointer in EventStore to deal with the ROOT I/O.
2. This is why I had called it `UserEventWeight` - but we can change it to another example if you think it is too confusing...

gaede:
... thinking about it, I think it would indeed better to return references. This is really an implementation detail that should be hidden.  We really need to rethink about the design of the EventStore-Reader-Writer interface as I find it rather unclear....

hegner:
Yap, store.get[Collection,Event,Run]MetaData() should return references.

On second thought about the event weight - maybe just leave it there. The experiments not considering it meta data anyhow do not have real event meta data. The difference between data and metadata is blurry and choices can be different

hegner:
And for the general interface I agree. We are in the situation where we (mostly I) mixed prototype and production quality. And code shows some inconsistencies

hegner:
Thanks! :-) Let me know once it is not WIP any more

gaede:
From my side this could be merged - unless @vvolkl or @andresailer have any comments or suggestions ?

src/ROOTWriter.cc
```diff
@@ -15,8 +15,14 @@ namespace podio {
     m_store(store),
     m_file(new TFile(filename.c_str(),"RECREATE","data file")),
     m_datatree(new TTree("events","Events tree")),
-    m_metadatatree(new TTree("metadata", "Metadata tree"))
-  {}
+    m_metadatatree(new TTree("metadata", "Metadata tree")),
```
hegner, line 18: Created issue #93 to discuss/review the implicit attachment of the created TTrees to the data file.

vvolkl:
I'm also thinking about how this could be used for usecases like saving the delphes output, where the "metadata" is mostly one value per particle in the collection, for example flavor tagging flags or isolation variables. With this implementation I guess I could add a vector of  collectionmetadata  with the same size of the collection. As Benedikt said, this is really blurring the line between data and metadata - so is this a usecase we want to support?

hegner:
> I'm also thinking about how this could be used for usecases like saving the delphes output, where the "metadata" is mostly one value per particle in the collection, for example flavor tagging flags or isolation variables. With this implementation I guess I could add a vector of collectionmetadata with the same size of the collection. As Benedikt said, this is really blurring the line between data and metadata - so is this a usecase we want to support?

To me that is clearly output of some sort of reco (even if in fast sim way) and thus event data.

vvolkl:
I agree, but I wonder if we will ever be able to cover all possible extra informations that will pop up in sim/reco/analysis in edm4hep. Maybe (ab-)using this mechanism to slightly extend the datamodel is better than the alternative of being too restrictive and having people circumvent the EDM altogether. In my opinion in FCCSW this did not work well at all, there were at least two separate authors who started writing to their private TFiles in Algorithms instead of using the DataService. @gaede I think you also mentioned that the metadata in LCIO were used this way?

An alternative could be to encourage developers to write their own "extended" classes and compile them in extra dictionaries. So that there might be a composition of the standard `ReconstructedPartice` and extra information (this is used much more in FCC-EDM, with a `core` type and additions), which could still be pruned back to only the standard types and used in applications that don't know about the extended class.

gaede:
@hegner I have added sth. to the example.md documentation on meta data. @vvolkl as discussed in the last key4hep meeting, I'll merge this PR now and then we can iterate on potential improvements of the interface for accessing it.

hegner:
Thanks!

PR #95, opened 2020-06-09
Add non-const getter for Collections in EventStore

### LINKED ISSUE #94: EventStore: better way to keep track of collections used for writing
The EventStore currently has two methods to access it's collections by name:
```
    /// access a collection by name. returns true if successful
    template<typename T>
    bool get(const std::string& name, const T*& collection);
    /// access a collection by name
    /// returns a collection w/ setting isValid to true if successful
    template<typename T>
    const T& get(const std::string& name );
```

That's not really ideal when trying to create collections from an external datasource where it is not known how many / which collections there are at compile time. In this case, ideally I would create branches according to what the external datasource provides and keep track of the collections in a map: branchname -> collectionptr.

The const pointer cannot be used for this purpose, as I cannot call `create` in the next step of filling the branches in the event loop. And dealing with maps involving references is quite painful - maybe it's possible but I haven't figured it out. I managed to implement the same thing with a vector of references, but would then have to do a lot more bookkeeping. Simply adding a getter for non-const pointers seems like the easiest way.

vvolkl:
Fixes #94 

BEGINRELEASENOTES
- add non-const getter for Collections in EventStore

ENDRELEASENOTES

### DIFF
--- a/include/podio/EventStore.h
+++ b/include/podio/EventStore.h
@@ -53,6 +53,11 @@ namespace podio {
     template<typename T>
     bool get(const std::string& name, const T*& collection);
 
+    /// access a collection by name. returns true if successful
+    template<typename T>
+    bool get(const std::string& name, T*& collection);
+
+
     /// fast access to cached collections
     CollectionBase* getFast(int id) const{
       return ( m_cachedCollections.size() > id ? m_cachedCollections[id] : nullptr ) ;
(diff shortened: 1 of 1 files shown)


hegner:
I find such syntax quite dangerous. If mixing a const T& and T* interface of the otherwise same signature, I can see sloppiness and usage of the T* version raise drastically. Which is bad for thread-safety. 
Anyhow, I just realised that somehow the const-handling of the interface does not seem consistent at other places. Should we make that a discussion topic?

gaede:
As discussed in the EDM4hep meeting, we postpone this PR until we have a better understanding of the needed semantics of the EventStore API.

PR #97, opened 2020-06-10
Throw an exception if a non-existent product is required

### LINKED ISSUE #96: podio silently skips over non-existant products
When getting a product from the `EventStore`, e.g. via `store.get<edm4hep::TrackCollection>("tracks")`, podio currently does not even issue any warning that the product is not available. Instead a default constructed object of the desired type is returned. The most trivial way this can happen is a simple typo. However, it might actually be the case that a user requires a product that is simply not present in the `EventStore`.

I think podio should at least print a warning statement that the desired product is not found. ~~Arguably, it should be possible for the user to check whether the object is present in the EventStore or not. (I know this is in principle possible with `bool EventStore::get(const std::string&, const T*&)`, but I think this should also be possible using the more ergonomic `EventStore::get(const std::string&)`).~~ I just realized that this is actually easily possible by asking the returned collection whether it is valid via `collection.isValid()`. The question then becomes do we want to force the user to check whether the collection is valid or not?

This would be an excellent use case for `optional`, but that would require at least c++17, unless we want to pull in `boost`.

tmadlener:
Minimal solution to fix #96

### DIFF
--- a/include/podio/EventStore.h
+++ b/include/podio/EventStore.h
@@ -1,10 +1,12 @@
 #ifndef ALBERS_EVENTSTORE_H
 #define ALBERS_EVENTSTORE_H
 
+#include <stdexcept>
 #include <string>
 #include <vector>
 #include <type_traits>
 #include <set>
+#include <iostream>
 
 // podio specific includes
 #include "podio/CollectionIDTable.h"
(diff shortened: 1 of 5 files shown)


gaede:
It would be better to throw an exception in this case ...

PR #99, opened 2020-06-16
PodioDict: rootmap should point to dictionary library

andresailer:
As was pointed out in the meeting today, Root should probably still auto-load the required libraries.
I think the problem is that the rootmap files point to the base libraries instead of the dictionaries.

A similar PR for edm4hep is in the works. More testing also still needed

BEGINRELEASENOTES

- Point to libpodioDict.so instead of libpodio.so in rootmap file

ENDRELEASENOTES


tmadlener:
This solves the issue that has triggered this in the first place for me on Ubuntu 18.04. I have also ran the tests successfully on Ubuntu 20.04.

Issue #61, opened 2019-09-27
Add possibility to use range based for loops on relations and vector members


gaede:
Add a helper class `vector_view` and code generation such that on can write range based for loops, e.g.
```
auto mcps = store.get<MCParticleCollection>("MCParticles") ;
for( auto& p : mcps ){
    for( auto& d : p.getDaughters() ) {
        printDaughter(d) ;
    }
}

```

tmadlener:
This can be closed, as it was addressed by #107 (Wasn't closed automatically, because at the time of merging, the issue was not yet linked to the PR).

Issue #105, opened 2020-06-24
ostream operator can easily enter infinite loop


tmadlener:
It is easily possible to enter an infinite loop (and a subsequent core dump) using the ostream operator if `OneToManyRelations` are used like they are e.g. in `ExampleMC` (only the important parts here):

```yaml
ExampleMC:
  OneToManyRelations:
    - ExampleMC parents
    - ExampleMC daughters
```

Properly setting mother/daughter relations in this case
```cpp
ExampleMC mcp1;
ExampleMC mcp2;

mcp1.adddaughters(mcp2);
mcp2.adddaughters(mcp1);
```

will lead to an infinite loop in the ostream operator, since that is generated to be

```cpp
std::ostream& operator<<( std::ostream& o,const ConstExampleMC& value ){
  o << " id : " << value.id() << std::endl ;
  o << " parents : " ;
  for(unsigned i=0,N=value.parents_size(); i<N ; ++i)
    o << value.parents(i) << " " ; // the problem is here
  o << std::endl ;
  o << " daughters ): " ;
  for(unsigned i=0,N=value.daughters_size(); i<N ; ++i)
    o << value.daughters(i) << " " ; // and also here
  o << std::endl ;
  return o ;
}

```
and is thus calling itself over and over again if such a circular relation is found.

PR #107, opened 2020-06-24
Introduce range wrapper for OneToManyRelations and VectorMembers

tmadlener:
BEGINRELEASENOTES
- Introduce range wrapper for OneToManyRelations and VectorMembers to allow range-based for loops

ENDRELEASENOTES

Requires #106 

Currently to loop over related objects one has to do something like this

```cpp
ExampleMC mcp;
// fill with some daughters
for (auto it = mcp.daughters_begin(); it != mcp.daughters_end(); ++it) {
  // ...
}
```

A `podio::RelationRange` is introduced here. It provides `begin` and `end` and allows for range-based for loops:

```cpp
ExampleMC mcp;
// fill with some daughters
for (const auto& p : mcp.daughters()) {
  // ...
}
```
The tests that come with this break, because #105 needs to be fixed first (by #106)

### DIFF
--- a/include/podio/RelationRange.h
+++ b/include/podio/RelationRange.h
@@ -1,0 +1,30 @@
+#ifndef PODIO_RELATIONRANGE_H
+#define PODIO_RELATIONRANGE_H
+
+#include <vector>
+#include <iterator>
+
+namespace podio {
+  /**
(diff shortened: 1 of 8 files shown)


tmadlener:
Fixes #61

Issue #114, opened 2020-07-06
clangformat option is broken with python3


tmadlener:
When trying to invoke `podio_class_generator.py` using python3.8 with the `--clangformat` option the following error occurs

```
Traceback (most recent call last):
  File "/home/tmadlener/work/podio/python/podio_class_generator.py", line 1189, in <module>
    gen.process()
  File "/home/tmadlener/work/podio/python/podio_class_generator.py", line 71, in process
    self.process_components(self.reader.components)
  File "/home/tmadlener/work/podio/python/podio_class_generator.py", line 79, in process_components
    self.create_component(name, components["Members"])
  File "/home/tmadlener/work/podio/python/podio_class_generator.py", line 948, in create_component
    self.fill_templates("Component", substitutions)
  File "/home/tmadlener/work/podio/python/podio_class_generator.py", line 1136, in fill_templates
    self.write_file(filename, content)
  File "/home/tmadlener/work/podio/python/podio_class_generator.py", line 1088, in write_file
    content = cfproc.communicate(input=content)[0]
  File "/usr/lib/python3.8/subprocess.py", line 1024, in communicate
    stdout, stderr = self._communicate(input, endtime, timeout)
  File "/usr/lib/python3.8/subprocess.py", line 1846, in _communicate
    input_view = memoryview(self._input)
TypeError: memoryview: a bytes-like object is required, not 'str'
```

This seems to be one of those cases where python3 behaves differently from python2. However, I have to setup a python2 environment first to see if it actually works there to confirm that it is indeed only the difference between the two versions.

If that is confirmed I would try to make this work in both versions.

Issue #116, opened 2020-07-09
Discussion about python code generator update / refactoring


tmadlener:
I have started to work on making the python code that fills the templates to generate the full classes a bit more readable. Ideally this would also result in better maintainability. While doing this I have stumbled upon some things for which I didn't really find anything in the documentation (or equally likely, I have missed it).

- Some fields in the data model syntax are not really used, respectively after reading them in, they are not used for the class generation. E.g.:
  - `const_declaration` in the `ExtraCode` field of components (#115)
  - `ConstExtraCode`, `TransientMembers`, `Typedefs` for the datatypes

- On the other hand it seems that `const_declaration` and `const_definition` should have been placed under the `ConstExtraCode` but are now generally placed in `ExtraCode`. It is not entirely clear from the documentation what the difference between `declaration` and `const_declaration` is or should be. As far as I have gathered the former goes only into the `DataType`, while the latter additionally also goes into the `ConstDataType`(?)

I am not sure what the original design idea was, but it seems to me that it is no longer rigorously followed.

----

I have also been playing with a python [schema validation library](https://pypi.org/project/schema/), and from a few [quick tests](https://gist.github.com/tmadlener/6f97cc77aabbe7a6b821b6bd74344c27#file-podio_schema_proto-py-L155-L176) it looks to be feasible to apply to the podio datamodel files. In the gist I already do some very basic parsing of the members of the datatypes. I think to get the very basic errors that can happen in the yaml file this would be enough. It might even be possible to do a more thorough validation, but that would need to be tested. The main question here is, is it OK to pull in another dependency for this or should we aim to do this without?

joequant:
One idea that I'd like to throw out is whether it would be a good idea to include templated classes so instead of merely having

class ExampleHitObj y{
}

you also have

template<>
class Obj<ExampleHit> {
}

typedef ExampleHitObj Obj<ExampleHit>

My suspicion is that if the classes were templatized that some of the code duplication could be removed
and you could write generic functions for specificing the reltationship between Obj and Data, and you 
can do things like write generic functions to move between const and non-const objects and convert to and
from STL and other frameworks.

Thoughts?

tmadlener:
[...]

PR #117, opened 2020-07-20
Code generator refactoring

### LINKED ISSUE #115: const_declaration code is not generated for components
Extra code that is declared as `const_declaration` is not generated for components. 

Since there really is no difference between `const_declaration` and `declaration` for components, I think it is OK to only consider the `declaration` for components. However, since then `const_declaration` is most likely a user input error, the validation should get this.

Do I miss something, that would make a differentiation between `const_declaration` and `declaration` necessary for components?

tmadlener:
BEGINRELEASENOTES
- Refactoring of the podio class generator to improve its readability. This refactoring is mainly focusing on reducing the size of some methods to make them easier to grasp as well as trying to more cleanly separate the validation of the inputs and the c++ code generation.  **The validation is stricter than the current version and will potentially break things. Specifically, the handling of `ExtraCode` and `ConstExtraCode` is now implemented according to the documentation.** In the previous version this has somewhat diverged from there. **The yaml declaration of `components` has been made equivalent to the ones of the `datatypes` but "old-style" definitions are still handled properly.** In the latter case a deprecation warning is emitted. **The generated c++ code is functionally equivalent, but not equal character by character.** The differences are mainly whitespace, but at some places also some of the generated code has been updated to more recent c++ capabilities. Additionally, some tests for the validation are added. They are not really exhaustive and should be considered to serve mainly documenting purposes for the moment.

ENDRELEASENOTES

- Fixes: #114, fixes #115 

Regarding the stricter validation: e.g. #115 and key4hep/EDM4hep#71 will now be caught by the validation and not passed on to code generation.

Some work is still left to do (**updated after the meeting on July 21**):
- [x] Some methods need to still be looked at to see if they can be reasonably split
- [x] Check of templates to see if they can / should be updated to newer c++ versions
- [x] Possibly unify the implementation of the supported syntax and the documentation (See #116), resp. update documentation to match with implementation again
  - [x] Have a consistent way of defining components and datatypes. Deprecate the current style of defining components.
  - [x] Handling of `ExtraCode` and `ConstExtraCode` as advertised in the documentation.
- [x] Check the possibility of using a template engine

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,9 +34,9 @@ endif()
 
 # Set up C++ Standard
 # ``-DCMAKE_CXX_STANDARD=<standard>`` when invoking CMake
-set(CMAKE_CXX_STANDARD 14 CACHE STRING "")
+set(CMAKE_CXX_STANDARD 17 CACHE STRING "")
 
-if(NOT CMAKE_CXX_STANDARD MATCHES "14|17")
+if(NOT CMAKE_CXX_STANDARD MATCHES "17")
   message(FATAL_ERROR "Unsupported C++ standard: ${CMAKE_CXX_STANDARD}")
 endif()
 
--- a/include/podio/CollectionBase.h
+++ b/include/podio/CollectionBase.h
@@ -53,7 +53,7 @@ namespace podio {
     virtual bool isValid() const = 0;
 
     /// number of elements in the collection
-    virtual int size() const = 0;
+    virtual size_t size() const = 0;
 
     /// fully qualified type name of elements - with namespace
     virtual std::string getValueTypeName() const = 0;
--- a/python/generator_utils.py
+++ b/python/generator_utils.py
@@ -52,24 +52,25 @@ def _format(string, **replacements):
 
 def get_extra_code(classname, definition):
   """Get the extra code from a datatype definition"""
-  extra_code = {'decl': "", 'const_decl': "", 'code': "", 'const_code': "", 'includes': set()}
+  extra_code = {'decl': "", 'const_decl': "", 'impl': "", 'const_impl': "", 'includes': set()}
 
   extra = definition.get('ExtraCode', {})
-  if not extra:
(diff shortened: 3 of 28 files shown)


tmadlener:
The currently failing `python` tests are because `flake8` (and also `pylint`) have non-zero exit codes if they encounter any warning. The ones that `flake8` is currently complaining about are mainly about style (when run locally):

I have now addressed all but `E121` (continuation line under-indented for hanging indent) and `E133` (closing bracket is missing indentation). 

Could we consider adding these (or at least `E121`) to the ignore list for `flake8`? Since we are using a 2 spaces indent, strictly following `E121` means that lists or dicts that are defined over more than one line have to be indented by 4 additional spaces, which breaks the flow a bit and and is not really consistent with the rest in my opinion.

For `E133` it is mainly personal taste that I like the closing bracket at the same indentation level as the variable.

tmadlener:
After consulting the [DD4hep python configuration](https://github.com/AIDASoft/DD4hep/blob/master/setup.cfg), I have adapted to the style mandated by `flake8` in order to be consistent in the grand scheme of things.

I think introducing a template engine can be done in a separate PR and this one is ready for review and merge.

hegner:
Hi Thomas! Can you mark this as WIP until finished? Thanks.

tmadlener:
Hi Benedikt, if we do not want to have a template engine, I think this is ready for review and can be considered finished (barring any larger problems detected during review). The commit from today fixed a small discrepancy in the generated code that has escaped detection until now because ti has not been used.

I have started an implementation using `jinja2` as template engine, but that will still take a few days until all the templates are ported. I could put that into another PR, or mark this as WIP again and add it here. Personally I have no preference of either of the two options.

hegner:
Maybe do a second PR. Then we have a working version of the code creation in the repo to test against.

gaede:
@tmadlener can you please rebase this PR - I have just merged a few older PRs...

PR #120, opened 2020-08-03
Switch to jinja2 template engine to generate c++ code

tmadlener:
BEGINRELEASENOTES
- Use `jinja2` template engine to generate c++ code. This makes it possible to more cleanly separate the generation of the c++ code and all the necessary pre-processing. Now only the pre-processing is done in python, while the complete generation is done using `jinja2`. This should make it much easier to make changes to the generated c++ code.

ENDRELEASENOTES

This builds on top of #117 ~(and until that is merged also includes all changes introduced there)~

While discussing the changes introduced in #117 the topic of using a "proper template engine" for generating the c++ code was briefly discussed. This was mainly motivated by the fact that the previous implementation did a lot of special case handling and corresponding string building and formatting in the python generator part. Using `jinja2` as template engine allows to more clearly separate the actual generation of the code and the necessary pre-processing. Now code generation is done entirely in `jinja2` templates, and no string building or formatting is done in the python part. The class generator now only collects all the necessary includes and forward declarations and puts them into a structure which is easily usable in the templates.

~A diff containing between #117 and this PR can be found [here](https://github.com/tmadlener/podio/compare/update-python...jinja2-templates).~

The commits are not cleanly separated in this PR and I don't think all of them build individually, so I would suggest to squash them into one (really) big commit on merge. Since this is in principle a rewrite of a large fraction of the code generation, I don't think that keeping a finely grained history makes much sense here.

A few things are still open at this point
- [x] Printing of all members in Collection ostream overload
- [x] re-introduce report printing (has fallen victim to very eagerly deleting code)


tmadlener:
I have done the following to clean up the commit history a bit for this PR:
- Merged all the commits that introduce the `jinja2` engine into one very large commit. 
- Made a dedicated separate commit where the "old" generator is replaced by the new `jinja2` based generator.

In this way both generators co-exist for one commit and all commits in this PR can be built and the tests run successfully. 

There are some smaller commits that address smaller fixes that I haven't absorbed into the two large commits mentioned above.

From my side this is ready for review and merge.

hegner:
Great work!
Which kind of testing did you do so far?

tmadlener:
> Which kind of testing did you do so far?

- I checked that the `podio` and `edm4hep` tests build and run. For edm4hep I used the current master, i.e. after key4hep/edm4hep#75.
- I have also looked at diffs of the generated c++ code before and after this PR. Similar to #117 there are again some whitespace changes, and some smaller changes that don't affect the functionality of the generated code. These changes include:
  - some cleanup of forward declarations and includes (e.g. previously it could happen that `Obj` class was forward declared while also including the `Obj.h` header. Now only the header include is present).
  - Switching typedef declarations from `typedef someClass someOtherClass` to `using someOtherClass = someClass`.

I think this should cover our bases, but if there are more tests or use cases that are not covered by this, I would happily test them.

hegner:
Thanks. I think that should cover the bases indeed. Your code looks nice (at least as much as is possible with this task ;-) ). Squashing is OK with you?

tmadlener:
> Thanks. I think that should cover the bases indeed. Your code looks nice (at least as much as is possible with this task ;-) ).

Thanks :)

Squashing is fine for me. This will in any case be a pretty large and pretty opaque change from outside this PR.

hegner:
And included. Now things should be easier in the future...

hegner:
Just merged and now I remembered there was something missing. Can you update the top-level README? About prerequisites?

tmadlener:
Ah yes that is a very good point. I will do that immediately and open a new PR.

PR #127, opened 2020-08-25
Make internal maps non-mutable

tmadlener:
BEGINRELEASENOTES
 - cleanup of `GenericParameters` for meta data
      - remove mutable from internal maps
ENDRELEASENOTES

By using a map<K, V>::const_iterator as return type from map::find()
there is no longer a need to mark the internal maps as mutable.

With these changes in place a refactoring to not have so much code repetition is almost trivial. I could do it in this PR, or it could be put up as a "good first issue" as we have discussed some time ago, if we still want to compile such a list to point newcomers to.

### DIFF
--- a/include/podio/GenericParameters.h
+++ b/include/podio/GenericParameters.h
@@ -32,86 +32,82 @@ namespace podio {
     using FloatMap = MapType<float>;
     using StringMap = MapType<std::string>;
 
-  public: 
-    
-    GenericParameters() = default; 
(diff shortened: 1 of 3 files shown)


tmadlener:
Rebased this PR onto master and implemented some of the trivial refactoring for reducing code duplication. I have also made all member functions non-virtual, since we are currently not inheriting from this, nor do I think that we will do so in the future.

PR #133, opened 2020-09-03
Fix meta data handling in ROOTReader and unify tests

### LINKED ISSUE #132: ROOTReader doesn't handle meta data properly when reading from multiple files
When reading from more than one file the reading of the meta data breaks after the first file has been read. The `read-multiple` test doesn't actually cover this so it has gone undetected until now.

The reason seems to be that in
https://github.com/AIDASoft/podio/blob/6a1e506efbd4d998beb05b07de8d38898fde4308/src/ROOTReader.cc#L18-L20

the call to `m_chain->GetFile()` possibly still returns the "old" file and has not necessarily switched to the "new" file yet. For this to happen a call to `m_chain->LoadTree(m_eventNumber)` needs to happen, which is happening only in `readCollection`:
https://github.com/AIDASoft/podio/blob/6a1e506efbd4d998beb05b07de8d38898fde4308/src/ROOTReader.cc#L84

This issue is easily solved for meta data that is expected to change with every event / run. However, it is a bit harder, also from a conceptual point of view, for meta data that are read at the beginning of the file which is done in `openFiles`, but only for the first file. It is possible that the stored meta data in the next file are different, but currently this is not taken into account when switching files.

tmadlener:
BEGINRELEASENOTES
- Make `ROOTReader` handle file switches properly for meta data reading.

ENDRELEASENOTES

Fixes #132 

In principle #132 is not yet completely fixed by this, because only the run, collection and event meta data are handled properly. Each file also has another metadata collection which currently is not yet updated during a file switch.

Unifying the testing code by making `read` and `read-multiple` use the same `processEvent` function (which lead to the discovery #132). Additionally making `write` a template function to easily switch out different writers. This significantly reduces code duplication, increases test coverage and also (hopefully) makes maintenance a bit easier in the future.


tmadlener:
Rebased to fix the merge conflict. 

Can the two commits here be squashed into one when merging? Then the `read-multiple` test will not be broken intermediately, which it currently is after the first commit.

tmadlener:
Since #97 changed the behavior of the EventStore (exception on non-existant collections), the tests needed some changes as well.

hegner (approved):

PR #134, opened 2020-09-07
Fix memory leaks in EventStore and Reader/Writer interplay

### LINKED ISSUE #139: reading and writing events leaks memory
It is not really visible with the 2000 events that are currently used for the tests, but simply increasing that number and monitoring the memory consumption of the corresponding process shows that it continuously increases with the running time of the process. I am not yet sure where / what memory is leaked, but it is not negligible. For me the memory footprint for `tests/write` increases from about 120M at startup to 1150M when writing 1e6 events. Similarly, `tests/read` starts off at around 220M and increases to 1000M at the end.

Repeating the tests with different readers / writers, e.g. the `SIOReader` and `SIOWriter` from #130, shows similar behavior, so it seems that the interplay between the readers / writers and the EventStore could be the culprit here.

tmadlener:
BEGINRELEASENOTES
- Make the EventStore actually take ownership of the metadata passed to it by the readers. (see #139)
- Make the collections properly clean up data of VectorMembers (see #139)
- Fix small memory leak in the EventStore for the CollectionIDTable.

ENDRELEASENOTES

The `m_table` pointer to the collection ID table is not deleted in the deconstructor even though it is acquired via `new` in the constructor. Using a `unique_ptr` to make clear that the `EventStore` owns this table (which it seems to do at least following the logic in `setCollectionIDTable`)

Fixes #139

### DIFF
--- a/include/podio/EventStore.h
+++ b/include/podio/EventStore.h
@@ -4,9 +4,9 @@
 #include <stdexcept>
 #include <string>
 #include <vector>
-#include <type_traits>
 #include <set>
 #include <iostream>
+#include <memory>
 
 // podio specific includes
 #include "podio/CollectionIDTable.h"
@@ -80,7 +80,7 @@ namespace podio {
     /// set the reader
     void setReader(IReader* reader);
 
-    CollectionIDTable* getCollectionIDTable() const {return m_table;};
+    CollectionIDTable* getCollectionIDTable() const { return m_table.get(); }
 
     virtual bool isValid() const final;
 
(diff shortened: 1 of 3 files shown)


hegner:
Thanks! Nice work. Let's wait for the tests for the final review

hegner:
expected that the Mac test fails?

tmadlener:
> expected that the Mac test fails?

Not really, but this one is still related to #137 and is only related to the fact that github changed the mac virtual environment a few days ago and this no longer plays nicely with the root version we currently use.

hegner (approved):

PR #135, opened 2020-09-09
Read the collectionID table directly

tmadlener:
No need to first read it into a temporary object and then re-building it with information from the temporary object.

### DIFF
--- a/src/ROOTReader.cc
+++ b/src/ROOTReader.cc
@@ -123,17 +123,10 @@ namespace podio {
     for (const auto& filename:  filenames) {
       m_chain->Add(filename.c_str());
     }
-    CollectionIDTable* l_table = new CollectionIDTable();
+    m_table = new CollectionIDTable();
     auto metadatatree = static_cast<TTree*>(m_chain->GetFile()->Get("metadata"));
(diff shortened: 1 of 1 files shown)


hegner:
Hi! is the ownership and responsibility for delete well defined?

tmadlener:
Hi @hegner, as far as I can tell the ownership doesn't change here, right? Prior to this, and also now, `m_table` is allocated via `new` and then populated. The main change here is to "cut out the middleman" of `l_table`, which was also allocated here (and then also leaked as far as I can tell).

From
https://github.com/AIDASoft/podio/blob/d4a02651332e2fe3ffbac4e99fa61b34600a16fa/include/podio/EventStore.h#L106
it looks like the `EventStore` currently takes ownership of this, although there is still a memory leak there, see #134 that tries to fix this.

hegner:
OK. Thanks. Just wanted to be sure.
(it just indicates how important it is to get rid of raw pointers wherever possible to get rid of this uncertainty)

hegner:
Any open issue blocking from merge?

tmadlener:
No this should be fine, as the change doesn't affect any outside behavior, apart from fixing the leaking of the temporary variable.

Issue #137, opened 2020-09-16
mac CI workflow is broken


tmadlener:
The CI workflow for mac is currently always failing because of a 
```
Fatal Python error: PyThreadState_Get: no current thread
```
occurring during the `pyunittest`. From what I could gather up until now this is most likely related to some cross talk between different python versions on macOS. However, I haven't really been able to diagnose this any further.

tmadlener:
After some more debugging, I think the most likely culprit is ROOT. I have adapted the mac workflow to print some more information about the environment and tried to see what the minimal example for python to fail is. See, e.g.: https://github.com/tmadlener/podio/actions/runs/258003293 (also the workflow file should be accessible from there).

Basically, I have checked which environment variables change by `source root/bin/thisroot.sh` and then disabled all of them again and enable them one by one to see which ones are necessary to trigger the error.

Setting `PYTHONPATH` and trying to `import ROOT` leads to the above described error, while the `jinja2` and `yaml` imports work fine. In the other cases the ROOT import fails with the obvious `ImportError`, because python doesn't know where it is.

Maybe the recent update of the github macos virtual environment: [commit](https://github.com/actions/virtual-environments/commit/865b3bf4cd0c6d0b24dbab6afbc693a2263d2627) leads to a mismatch in the python versions? Is there a ROOT version compiled against XCode 11.7 available somewhere to check this?

PR #138, opened 2020-09-16
update CI to use use newer LCG views

petricm:
BEGINRELEASENOTES
- Update CI actions to use LCG 96, 97, 98 for mac, centos7 and ubuntu1804

ENDRELEASENOTES
Should resolve the issue caused by the update of the virtual enviroment #137

### DIFF
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -3,21 +3,38 @@ name: mac
 on: [push, pull_request]
 
 jobs:
-  test:
+  compile:
     runs-on: macos-10.15
+    strategy:
+      matrix:
+        LCG: ["LCG_98python3/x86_64-mac1015-clang110-opt","LCG_97apython3/x86_64-mac1015-clang110-opt"]
     steps:
     - uses: actions/checkout@v2
-    - name: Download ROOT
+    - name: Install CVMFS
       run: |
-        curl -O https://lcd-data.web.cern.ch/lcd-data/ROOT/root_v6.20.02.macosx64-10.15-clang110_Xcode113_19D76.tar.gz
-        tar xf root_v6.20.02.macosx64-10.15-clang110_Xcode113_19D76.tar.gz
-        pip install -r requirements.txt
+        brew install ninja
+        brew cask install osxfuse
+        curl -L -o cvmfs-2.7.3.pkg https://ecsft.cern.ch/dist/cvmfs/cvmfs-2.7.3/cvmfs-2.7.3.pkg
+        sudo installer -package cvmfs-2.7.3.pkg -target /
+        wget --no-check-certificate https://lcd-data.web.cern.ch/lcd-data/CernVM/default.local
+        sudo mv default.local /etc/cvmfs/default.local
+        sudo cvmfs_config setup
+        brew cask install gfortran
+        brew cask install xquartz
+    - name: Mount CVMFS
+      run: |
+        mkdir -p /Users/Shared/cvmfs/sft.cern.ch
+        mkdir -p /Users/Shared/cvmfs/geant4.cern.ch
+        sudo mount -t cvmfs sft.cern.ch /Users/Shared/cvmfs/sft.cern.ch
+        sudo mount -t cvmfs geant4.cern.ch /Users/Shared/cvmfs/geant4.cern.ch
+        ls /Users/Shared/cvmfs/sft.cern.ch
+        ls /Users/Shared/cvmfs/geant4.cern.ch
     - name: Compile and test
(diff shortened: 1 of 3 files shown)


petricm:
Why the `pyunittest` fails for ROOT 6.22 and not for 6.20 I don't know.

petricm:
Seems that `pyunittest` does also not work on linux with ROOT 6.22, just wait that the PR CI finishes.

tmadlener:
Hi @petricm,
The failure of the `pyunittest` is looking like an "actual" failure this time and not the one described in #137. I am not entirely sure why this seems to fail just now, because the tests ran succesfully under python3 before, but the failure is consistent for `linux` and `mac` and the original issue seems to be resolved.

petricm:
@tmadlener this is not a issue of python versions but of ROOT versions. LCG_98=ROOT 6.22 and LCG_97=ROOT 6.20. You did no test 6.22 before.

tmadlener:
Ah, I see. Thanks for clarifying.

tmadlener:
I am not entirely sure why this breaks with root 6.22.00, but didn't with 6.20.04. The problem is that when we try to loop over a collection in python, the `__getitem__` method gets called and that usually expects two arguments: `self` and an index. It seems that in our case the `__getitem__` method gets called with only one argument and, hence, complains about a missing positional argument.

In any case, not overriding the `__getitem__` method as we currently do, but letting `cppyy` seems to get rid of the problem. I have opened #141 with the necessary changes.

tmadlener:
The ubuntu build seems to be failing because ROOT defines it's own `string_view`, resp. aliases the one from `experimental/string_view`. Could it be that the dictionary generation picks up the wrong c++ standard? I couldn't find the c++ standard with which the root release used in the docker image is built, but is it possible that it is not c++17? At least locally on Ubuntu 20.04 I can build podio against root 6.22, when I also use c++17 for building root.

tmadlener:
With the changes from #141 applied on top of this one all workflows look good, apart from the ubuntu one, which still fails during the build:
[mac](https://github.com/tmadlener/podio/actions/runs/259629585)
[linux](https://github.com/tmadlener/podio/actions/runs/259629587)
[python](https://github.com/tmadlener/podio/actions/runs/259629576)
[ubuntu](https://github.com/tmadlener/podio/actions/runs/259629577)

petricm:
@tmadlener you are right, the ubuntu root has the flags `-pthread -std=c++11 -m64 -I/opt/root/include`, but I would advice that you inherit the CXX standard from ROOT and not set it explicitly like here 

https://github.com/AIDASoft/podio/blob/d4a02651332e2fe3ffbac4e99fa61b34600a16fa/CMakeLists.txt#L37

petricm:
I have removed the ubuntu image update

tmadlener:
> @tmadlener you are right, the ubuntu root has the flags `-pthread -std=c++11 -m64 -I/opt/root/include`, but I would advice that you inherit the CXX standard from ROOT and not set it explicitly like here


The problem is that podio is using some features that are only available in c++17. The switch to c++14 was done 4 years ago. However, it is possible that compiling it with a lower standard was still possible for some time. The decision to switch to c++17 for podio (and edm4hep) was made a few months back. so in that regard, switching back to c++14 would probably be possible, but I would really not like to do so, because c++17 has some really nice features, and it is 2020 after all ;)

petricm:
I completely agree. It would be however still good for the user that you check the ROOT libs for their `target_compile_features` and check if they have the same standard as you are building podio and if not break at the cmake step.

vvolkl:
@petricm agreed - I usually build podio with spack, where the recipe already is updated with the correct root dependency, but it's a good idea nonetheless.

tmadlener:
That is a very good idea indeed. I have added it to #141 

CI runs with the changes of this PR and #141 can be found here:
- [ubuntu](https://github.com/tmadlener/podio/actions/runs/260861294)
- [python](https://github.com/tmadlener/podio/actions/runs/260861293)
- [linux](https://github.com/tmadlener/podio/actions/runs/260861292)
- [mac](https://github.com/tmadlener/podio/actions/runs/260861291)

tmadlener:
Should we merge this PR with #141, so that it is easier to see that the combination of the two fixes the current issues with the CI?

petricm:
Replace the ubuntu root image with the ubutu LCG view

petricm:
@tmadlener cherry-pick my PR into yours and copy the release notes.

tmadlener:
> @tmadlener cherry-pick my PR into yours and copy the release notes.

Done

PR #141, opened 2020-09-17
Fix python bindings for root 6.22 and update CI

### LINKED ISSUE #137: mac CI workflow is broken

tmadlener:
BEGINRELEASENOTES
- Update CI actions to use LCG 96, 97, 98 for mac, centos7 and ubuntu1804
- Make python bindings work with root 6.22 (and onwards)
- Make sure that root has been built with c++17 at the cmake stage
- Require at least CMake 3.12

ENDRELEASENOTES

For some reason cppyy seems to be calling the `__getitem__` method with
only one argument instead of the usual two. Starting from root 6.22 this
breaks our python access to the collections. By only overriding it if it
is not already defined we can get it to work with root <= 6.20 and root
> 6.22.

Also removed the `__iter__` overload, because that is not necessary to
loop over a collection, specifically because the iterator protocol would
also require a `__next__` (or `next` in python2) method for actually
advancing.

Fixes #137

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,6 +73,22 @@ option(CREATE_DOC "Whether or not to create doxygen doc target." OFF)
 list(APPEND CMAKE_PREFIX_PATH $ENV{ROOTSYS})
 set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 find_package(ROOT REQUIRED COMPONENTS RIO Tree)
+
+# Check that root is compiled with a modern enough c++ standard
+get_target_property(ROOT_COMPILE_FEATURES ROOT::Core INTERFACE_COMPILE_FEATURES)
+if (NOT "cxx_std_17" IN_LIST ROOT_COMPILE_FEATURES)
+  message(FATAL_ERROR "You are trying to build podio against a version of ROOT that has not been built with a sufficient c++ standard. podio requires c++17")
+endif()
+# Also verify that it has been built against the same python version as we are
+# using (that is if ROOT gives us the necessary tools)
+if (${ROOT_VERSION} VERSION_GREATER_EQUAL 6.19)
+  set(Python_FIND_FRAMEWORK LAST)
+  find_package(Python COMPONENTS Development)
+  if (NOT ${Python_VERSION} EQUAL ${ROOT_PYTHON_VERSION})
+    message(WARNING "Python version used for buildig ROOT is different than the current python version: " ${ROOT_PYTHON_VERSION} " vs " ${Python_VERSION})
+  endif()
+endif()
+
 # ROOT only sets usage requirements from 6.14, so for
 # earlier versions need to hack in INTERFACE_INCLUDE_DIRECTORIES
 if(ROOT_VERSION VERSION_LESS 6.14)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 # HSF recommends 3.3 to support C/C++ compile features for C/C++11 across all
 # platforms
-cmake_minimum_required(VERSION 3.8)
+cmake_minimum_required(VERSION 3.12)
 
 #--- Project name --------------------------------------------------------------
 project(podio)
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -6,6 +6,7 @@ jobs:
   compile:
     runs-on: macos-10.15
     strategy:
+      fail-fast: false
       matrix:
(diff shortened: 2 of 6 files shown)


tmadlener:
See [here](https://github.com/AIDASoft/podio/pull/138#issuecomment-694332573) for links to the results of the updated CI workflows.

tmadlener:
CI runs with the updates from #138 and the fixes from this one can be found here:
* [ubuntu](https://github.com/tmadlener/podio/actions/runs/260861294)
* [python](https://github.com/tmadlener/podio/actions/runs/260861293)
* [linux](https://github.com/tmadlener/podio/actions/runs/260861292)
* [mac](https://github.com/tmadlener/podio/actions/runs/260861291)

With both PRs merged we should have a working CI again.

petricm:
It would be good that you also add a python check, to verify that the python used in the current env is the same as the one used to build root.

Look at the snipets from DD4hep

https://github.com/AIDASoft/DD4hep/blob/a55c5098badc0123ec09b4815b87c600231a59a4/CMakeLists.txt#L115-L118

and

https://github.com/AIDASoft/DD4hep/blob/8ecaf5428dfe7072813a98fa457fc945041b6d0d/cmake/DD4hepBuild.cmake#L651-L660

tmadlener:
@petricm I have implemented the check without checking the ROOT version first. Is there a reason why dd4hep only does the check for root >= 6.19? I also only emit a warning if I detect a mismatch, since this doesn't seem to be a strict requirement.

petricm:
Because the python version variables does not exist prior to that version [see here](https://github.com/root-project/root/commit/43e0abba286df8fa61d0adc279bf7511ff05129a)

tmadlener:
Ah right, that makes sense. Thanks

CMakeLists.txt
```diff
@@ -73,6 +73,22 @@ option(CREATE_DOC "Whether or not to create doxygen doc target." OFF)
 list(APPEND CMAKE_PREFIX_PATH $ENV{ROOTSYS})
 set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 find_package(ROOT REQUIRED COMPONENTS RIO Tree)
+
+# Check that root is compiled with a modern enough c++ standard
+get_target_property(ROOT_COMPILE_FEATURES ROOT::Core INTERFACE_COMPILE_FEATURES)
+if (NOT "cxx_std_17" IN_LIST ROOT_COMPILE_FEATURES)
```
vvolkl, line 79: While we need cxxstd17 for podio, I believe root may be compiled with cxxstd14 for this to work also. And root builds with cxxstd14 are still very common, see for example the discussion here: https://github.com/root-project/root-docker/issues/35 But I'll double-check!

vvolkl (changes requested):

CMakeLists.txt
andresailer: I am in favour of insisting on a consistent standard.

tmadlener:
Rebased this PR on to #138 as suggested there.

tmadlener:
It looks like all the tests that are actually run in the CI are now running successfully, but github seems to still expects some of the old workflows, which no longer exist and, thus no longer run. 

I am not sure when or from where github picks up the list of checks that are necessary for a PR. I thought it was the list of .yml files that are in the .github/workflows folder at the HEAD of the PR, but it seems that also the ones that have been there at the start of a PR are still present.

I will try and see if closing and re-opening this makes any difference, unless there are strong arguments against that.

andresailer:
@tmadlener There is a list of "required" status checks that need to be successful. Under settings -> Branches, Branch Orotection Rules A few hours ago, I made those (by name) that are no longer part of the checks being run no longer required, so this PR is :heavy_check_mark: now.

tmadlener:
Thanks @andresailer 

Good to know for future reference, but I don't have the rights to actually see the repository settings for podio.

tmadlener:
Following this [comment in #141](https://github.com/AIDASoft/podio/pull/130#discussion_r493493462), I have bumped the minimum cmake version to be 3.12. Alternatively, the python compatibility check could be removed.

andresailer (approved):

petricm:
@tmadlener I suggest you also pick up https://github.com/AIDASoft/DD4hep/commit/f6e4fc4df7bf43175882bb661c0e7d2621a20424
> jobs.<job_id>.strategy.fail-fast
When set to true, GitHub cancels all in-progress jobs if any matrix job fails. Default: true

tmadlener:
@petricm thanks for suggestion. Done.

PR #142, opened 2020-10-02
Formally introduce the BUILD_TESTING switch

tmadlener:
Very trivial change that just documents that the option exists

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -126,6 +126,8 @@ install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/LICENSE
 #--- project specific subdirectories -------------------------------------------
 add_subdirectory(python)
 add_subdirectory(src)
+
+option(BUILD_TESTING "Enable and build tests" ON)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -126,6 +126,8 @@ install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/LICENSE
 #--- project specific subdirectories -------------------------------------------
 add_subdirectory(python)
 add_subdirectory(src)
+
+option(BUILD_TESTING "Enable and build tests" ON)
```
andresailer, line 130: This option is already set by `include(CTest)` https://cmake.org/cmake/help/v3.12/module/CTest.html
If this is just have the option explicitly in the File it should be moved to before the `include(CTest)`, so that at least its default value would be used if someone where to change it.

tmadlener: Ah right. I didn't know that. In that case this is completely unnecessary and I will close it. Sorry for the noise.

PR #143, opened 2020-10-13
Generate an additional cmake lists file containing the generated source files

tmadlener:
BEGINRELEASENOTES
- Generate an additional `podio_generated_files.cmake` file containing all generated source files as a `header` and `sources` list and make the code generation macro include this file to get the headers and source files.
  - Now only the files generated for the current settings are picked up by cmake
  - Makes it possible to have additional files in the folders where the generated files are placed, since these are no longer globbed over.

ENDRELEASENOTES

This now makes it possible to seamlessly switch between branches of podio that generate different files :)

The contents of the generated file are the following

```cmake
SET(headers
  # all header files in the format
  ${ARG_OUTPUT_FOLDER}/${datamodel}/FooObj.h
)
SET_PROPERTY(SOURCE ${headers} PROPERTY GENERATED TRUE)

SET(sources
  # All source files in the format
  ${ARG_OUTPUT_FOLDER}/src/FooObj.cc
)
SET_PROPERTY(SOURCE ${sources} PROPERTY GENERATED TRUE)


SET(selection_xml
  ${ARG_OUTPUT_FOLDER}/src/selection.xml
)
SET_PROPERTY(SOURCE ${selection_xml} PROPERTY GENERATED TRUE)
```

The `selection_xml` variable is not used at the moment, but since it is a generated file I put it there.

### DIFF
--- a/cmake/podioMacros.cmake
+++ b/cmake/podioMacros.cmake
@@ -147,12 +147,17 @@ function(PODIO_GENERATE_DATAMODEL datamodel YAML_FILE RETURN_HEADERS RETURN_SOUR
     ${podio_PYTHON_DIR}/podio_config_reader.py
   )
 
+  message(STATUS "Creating '${datamodel}' datamodel")
   # we need to boostrap the data model, so this has to be executed in the cmake run
   execute_process(
-    COMMAND ${CMAKE_COMMAND} -E echo "Creating \"${datamodel}\" data model"
     COMMAND python ${podio_PYTHON_DIR}/podio_class_generator.py ${YAML_FILE} ${ARG_OUTPUT_FOLDER} ${datamodel} ${ARG_IO_BACKEND_HANDLERS}
     WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+    RESULT_VARIABLE podio_generate_command_retval
     )
+
+  IF(NOT ${podio_generate_command_retval} EQUAL 0)
+    message(FATAL_ERROR "Could not generate datamodel '${datamodel}'. Check your definition in '${YAML_FILE}'")
+  ENDIF()
 
   # Get the generated headers and source files
   include(${ARG_OUTPUT_FOLDER}/podio_generated_files.cmake)
--- a/cmake/podioMacros.cmake
+++ b/cmake/podioMacros.cmake
@@ -142,6 +142,9 @@ function(PODIO_GENERATE_DATAMODEL datamodel YAML_FILE RETURN_HEADERS RETURN_SOUR
     DIRECTORY APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS
     ${YAML_FILE}
     ${PODIO_TEMPLATES}
+    ${podio_PYTHON_DIR}/podio_class_generator.py
+    ${podio_PYTHON_DIR}/generator_utils.py
+    ${podio_PYTHON_DIR}/podio_config_reader.py
   )
 
   # we need to boostrap the data model, so this has to be executed in the cmake run
--- a/cmake/podioMacros.cmake
+++ b/cmake/podioMacros.cmake
@@ -134,6 +134,7 @@ function(PODIO_GENERATE_DATAMODEL datamodel YAML_FILE RETURN_HEADERS RETURN_SOUR
     # At least build the ROOT selection.xml by default for now
     SET(ARG_IO_BACKEND_HANDLERS "ROOT")
   ENDIF()
+
   # we need to boostrap the data model, so this has to be executed in the cmake run
   execute_process(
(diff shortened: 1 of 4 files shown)


cmake/podioMacros.cmake
```diff
@@ -121,6 +121,18 @@ function(PODIO_GENERATE_DATAMODEL datamodel YAML_FILE RETURN_HEADERS RETURN_SOUR
   IF(NOT ARG_OUTPUT_FOLDER)
     SET(ARG_OUTPUT_FOLDER ${CMAKE_CURRENT_SOURCE_DIR})
   ENDIF()
+
+  # Make sure that no remnants from previous runs are still present in the
+  # output directory, to avoid running into unexpected problems with "old"
+  # generation files
+  IF(NOT __STARTED_GENERATION_FROM_CLEAN_SLATE)
```
andresailer, line 128: Does this really work as intended?

1.) cmake .. # cleans and sets __STARTED... True
2.) git pull  # e.g.,
3.) cmake .. # doesn't clean

Right?

tmadlener: Yes, you are right. However, in that case you should still have a "valid" build file and can do something along the lines of `make clean` to force the cleanup of the generator output directory. This fix should make starting in a clean build directory more predictable. E.g.
```bash
mkdir build && cd build
cmake ..
```
Then do whatever needs to be done here. Suppose you then want to start from a clean slate, but you forgot to `make clean` before doing the usual `rm -rf build` and then start the process from the beginning. In this case it is likely that the generated files from the previous run are still around and possibly pollute the new run, e.g. when some files are no longer generated, because the class has been removed from the yaml file in between.

tmadlener: However, maybe it would be more intuitive if each call to `cmake` cleans up the generator output directories? This would cover my intended use case for free basically.

Stupid question: Everything in `PODIO_GENERATE_DATAMODEL` is only run when `cmake` is invoked, right? The possible regeneration of the datamodel via a call to `make` is handled by the custom target that we add? In that case I think it would be possible to simply remove the generator output directories unconditionally without the usage of a cache variable.

tmadlener: > I would be using `git clean` at this point. 

Ah I hadn't thought about that possibility in fact. But that would probably also clean a lot of things it is not supposed to, e.g. files that are currently being worked on but not yet tracked.

> Or the podio output folder should go to CMAKE_BINARY_DIR , a.k.a. the "build" folder.

The problem with that is that it might be desirable to have the generated code somewhere else, e.g. for documentation purposes, or possibly to put it under version control (for some reason).

> If one wants no-longer needed generated files to be removed (or ignored), maybe it would be better for podio to spit out the created files instead of using the cmake globbing of files?

Yes, that is probably the most robust approach. It would also allow users to put other files into the generator output directories if they desire to do so.

tmadlener:
Seems like ninja is not really satisfied with including a generated file here and enters an infinite loop. This doesn't happen with make.

andresailer:
While googling:
Found: https://stackoverflow.com/questions/52556785/code-generator-generating-its-own-cmake-files-and-targets

We should add the datamodel yaml file to the CMAKE_CONFIGURE_DEPENDS property (https://cmake.org/cmake/help/latest/prop_dir/CMAKE_CONFIGURE_DEPENDS.html)
Then Cmake re-runs when the yaml file changes. That maybe simplifies many things.

That cmake file should probably only written if its content changes to avoid triggering re-running?

The header and source files should have the GENERATED property, having them in BYPRODUCTS was causing this to happen.

tmadlener:
Thanks for the `CMAKE_CONFIGURE_DEPENDS` hint, I think that solves pretty much all of our problems, and I could remove the custom target for the generation. I have added the `GENERATED` property to all generated files.

I have added the templates to that list as well, as changes to the templates should also trigger a re-configuration. Now these changes should make it possible to have "dirty" directories into which the files are generated, as only those headers and source files that are actually generated by the podio code generator will be used to populate the corresponding cmake variables. This in turn makes the following possible:
- Switch between different settings seamlessly, e.g. `ENABLE_SIO=ON` and switching to `ENABLE_SIO=OFF` now works, without having to clear all the generated files first
- Switch between different branches in podio which have other templates without having to clear all the generated files first

tmadlener:
Rebased onto the laster master to resolve conflicts.

hegner:
@tmadlener - just to x-check. we are complete here and I can merge?

tmadlener:
Yes, for me this is complete. Maybe @andresailer wants to have another look.
I have been using this as basis for all my podio developments locally, and it has been working smoothly for me (at least after I have found all the dependencies for retriggering a generation ;) )

hegner (approved):

PR #144, opened 2020-11-03
Decouple writers and EventStore

### LINKED ISSUE #103: cannot write out event data previously read from file
Currently we cannot write out event data previously read in from a file to another file.
This is a standard use case, e.g. in reconstruction or any other event data processing job.
See example ./tests/read_and_write.cpp (see #102 ).
- the main reason is the different memory layout of relations and vector members in events that have been read from a file.

tmadlener:
BEGINRELEASENOTES
- Decouple the writers and the EventStore to allow to write collections that have previously been read from a file.

ENDRELEASENOTES

~Builds on top of #130, [diff without changes already in there](https://github.com/tmadlener/podio/compare/new-sio-backend...tmadlener:allow-read-write)~

Fixes #103 

Previously the writers assumed that the collections are re-used throughout the lifetime of the EventStore. However, since the readers actually create new collections for every event, this also made it impossible to write events to a new file after they have been read from a file. Simply, forcing the writers to update the collection information (buffer addresses, etc.) for every event makes this possible. The collections themselves acquire a new flag that is necessary to steer what has to be done in `prepareForWrite`.


tmadlener:
I removed the [WIP] label, because I think this could actually be merged as is, since other work in this direction will probably still take some time and since this fixes a long standing and pretty serious issue it might be better to have the fix for that first.

Also: Closing and re-opening this because the failing test failed in CVMFS setup.

PR #147, opened 2020-11-13
fix for macos: link edm-core library to edm-sioBlocks

gaede:
BEGINRELEASENOTES
- fix for MacOs when using SIO I/O with podio
      - need to link edm-core library to edm-sioBlocks library

ENDRELEASENOTES


tmadlener:
I just realized that we have only workflow in the CI that actually uses the sio backend. Should we add at least one for all the platforms we aim to support? That would mean adding one for mac and ubuntu, I suppose.

gaede:
I guess we cannot have too many tests, so yes, absolutely.

PR #152, opened 2020-11-23
make EventStore non-copyable

joequant:
BEGINRELEASENOTES
- Make EventStore non-copyable

ENDRELEASENOTES


joequant:
It looks like it failed for system reasons, can you rerun?

joequant:
Seems to pass all of the tests.  Can you review and check in?

PR #153, opened 2020-11-23
rpath, linker, compiler flags and find python fixes

### LINKED ISSUE #76: rpath setup
There is a need to setup some rpath handling, something along these lines would be good

https://github.com/root-project/root/blob/master/cmake/modules/RootBuildOptions.cmake#L401-L422

same for EDHM4hep

### LINKED ISSUE #145: Find python issues
There are some issue with finding python on ubuntu and with certain cmake versions (check your pipeline, it finds the wrong python) see problem explanation and solutions

https://github.com/AIDASoft/DD4hep/pull/725
https://github.com/AIDASoft/DD4hep/pull/728

petricm:
resolves #145 
resolves #76
BEGINRELEASENOTES
- Set rpath for macOS and externalize compiler and linker flags
- Search for the same version of python as was used for building ROOT

ENDRELEASENOTES

the compiler flags are only a suggestion

### DIFF
--- a/cmake/podioBuild.cmake
+++ b/cmake/podioBuild.cmake
@@ -50,7 +50,7 @@ macro(podio_set_compiler_flags)
   endif()
 
   FOREACH( FLAG ${COMPILER_FLAGS} )
-    ## meed to replace the minus or plus signs from the variables, because it is passed
+    ## need to replace the minus or plus signs from the variables, because it is passed
     ## as a macro to g++ which causes a warning about no whitespace after macro
     ## definition
     STRING(REPLACE "-" "_" FLAG_WORD ${FLAG} )
--- a/cmake/podioBuild.cmake
+++ b/cmake/podioBuild.cmake
@@ -1,0 +1,92 @@
+#---------------------------------------------------------------------------------------------------
+#---podio_set_rpath
+#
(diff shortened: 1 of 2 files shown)


gaede (approved):
Worked nicely on my macbook. Also, activating the compiler warnings is really helpful.

cmake/podioBuild.cmake
```diff
@@ -0,0 +1,92 @@
...
+
+  SET(COMPILER_FLAGS -Wshadow -Wformat-security -Wno-long-long -Wdeprecated -fdiagnostics-color=auto -Wall -Wextra -pedantic -Weffc++)
+
+  # AppleClang/Clang specific warning flags
+  if(CMAKE_CXX_COMPILER_ID MATCHES "^(Apple)?Clang$")
+    set ( COMPILER_FLAGS ${COMPILER_FLAGS} -Winconsistent-missing-override -Wheader-hygiene )
+  endif()
+
+  FOREACH( FLAG ${COMPILER_FLAGS} )
+    ## meed to replace the minus or plus signs from the variables, because it is passed
```
vvolkl, line 53:
```suggestion
    ## need to replace the minus or plus signs from the variables, because it is passed
```

PR #154, opened 2020-11-24
disable explicit assignment operator - implement #111

### LINKED ISSUE #111: (Non)-Copyable Containers
Can we make containers non-copyable?

(a follow up on #109)

joequant:
This disables the explicit assignment operator to enforce copy
constraints.  The copy constructor is not disabled since this
interferes with root adapters.

BEGINRELEASENOTES

- Disable operator = for collections so that it maintains one copy of collections, fixes #111

ENDRELEASENOTES


tmadlener:
This looks good to me and would also get rid of a few `Weffc++` warnings.

PR #156, opened 2020-12-06
make readers/writers non-copyable

joequant:
Since the readers/writers have pointers to internal state,
they should not be copied.

BEGINRELEASENOTES
- Readers/writers are now noncopyable
ENDRELEASENOTES


gaede:
Can you please rebase to the master. Otherwise I think this should be merged...

andresailer:
Please rebase

PR #157, opened 2020-12-06
make array indices and counts unsigned

joequant:
BEGINRELEASENOTES
- fix array indices and counts to be unsigned
ENDRELEASENOTES


gaede:
Should be merged with #158.

PR #158, opened 2020-12-06
more fixes to unsigned and shadowed variables

joequant:
BEGINRELEASENOTES
- Changed various loops to use unsigned rather than int

ENDRELEASENOTES

### DIFF
--- a/include/podio/ObjBase.h
+++ b/include/podio/ObjBase.h
@@ -11,7 +11,7 @@ namespace podio {
   public:
 
     /// Constructor from ObjectID and initial object-count
-    ObjBase(ObjectID id, int i) : id(id) , ref_counter(i) {};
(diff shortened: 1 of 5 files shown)


gaede:
@joequant could you please combine the changes in #157 within this PR ? I think this should remove some of the warnings of type `comparing ints w/ different sign`. Also there are errors with narrowing on macos, e.g. here: https://github.com/AIDASoft/podio/pull/158/checks?check_run_id=1507010819#step:5:145.
In the end we have to see if this makes things cleaner/less error prone given the fact that we need the object ids to stay `int` in order to allow for marking `invalid` with negative numbers ....

tmadlener:
All of the changes here have been adapted in #171. Directly cherry-picking the commit from here was not possible due to different approaches that are chosen to fix the `Wsign-compare`.

This can be closed.

PR #162, opened 2020-12-09
Add cmake 3.16 to check for which python version to require

### LINKED ISSUE #161: cmake finds "unsuitable" python version
Just pulled the latest master and I am running into the following Problem on Ubuntu 20.04 with cmake 3.16.3

```
CMake Error at /usr/share/cmake-3.16/Modules/FindPackageHandleStandardArgs.cmake:146 (message):
  Could NOT find Python: Found unsuitable version "3.8", but required is
  exact version "3.8.5" (found /usr/lib/x86_64-linux-gnu/libpython3.8.so)
  /usr/share/cmake-3.16/Modules/FindPackageHandleStandardArgs.cmake:391 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake-3.16/Modules/FindPython/Support.cmake:2214 (find_package_handle_standard_args)
  /usr/share/cmake-3.16/Modules/FindPython.cmake:304 (include)
  CMakeLists.txt:86 (FIND_PACKAGE)
```

I haven't tried yet with other versions of cmake, but it seems that the changes introduced in #153 are also having problems on when used with cmake 3.16.3 (at least on Ubuntu 20.04)

tmadlener:
BEGINRELEASENOTES
- Fix cmake problem #161 on Ubuntu

ENDRELEASENOTES

Fixes #161 

For now simply adding 3.16.3 to the list of "exceptional" cmake versions.


tmadlener:
After doing some more testing I think the problem has been introduced in 3.16.0 and fixed again in 3.17.3. I haven't tested all the versions between 3.14.3 and 3.17.3, but the ones I have tested seem to suggest this. The versions I tested are:
```
3.14.3
3.14.7
3.15.5
3.15.6
3.16.0
3.16.1
3.16.2
3.16.3
3.16.7
3.16.9
3.17.1
3.17.3
```

gaede (approved):

PR #171, opened 2021-01-19
Fix compiler warnings

### LINKED ISSUE #170: Discussion about enabled compiler flags
Currently we have the following list of compiler flags enabling different warnings:
https://github.com/AIDASoft/podio/blob/bcb18dafd494abef5d4fc94dc157a209734194a2/cmake/podioBuild.cmake#L45

Since these have been added with #153 there is quite a bit of noise in the build of podio. Some of these warnings are definitely useful and should be properly fixed, but others seem to lead to more false positives than actually useful warnings. 

### `Weffc++`
`Weffc++`, this one seems to lead to the most noise. It mainly warns about member variables that are not initialized in the member initialization list and classes that have pointer members but no copy constructor or assignment operator. The latter concern mainly our `Obj` classes, the `CollectionIterator`s and the `Collection`s themselves. The collections are partially covered by #154, but due to constraints from root we will probably not be able to get rid of all of these. The ones concerning the `Obj` and `CollectionIterator` can be addressed by changing the templates accordingly and explicitly deleting the non-present copy c'tor and assignment operators.

The warnings about members that should be initialized in the member initialization list partially occur in core classes, but also are visible a lot in generated classes. I am at the moment not entirely sure if they can all be fixed.

In the end I would suggest to fix all possible instances of these warnings and then disable this flag, as we will always have "false positives" in the generated `Collection`s, unless we generate guards around these classes as well. Some core classes are already covered in #156, but a few still need to be covered.

### `Wshadow`
There are a lot of `Wshadow` warnings occurring in member initialization lists of constructors with empty bodies. Hence, these are most likely false positives as well in the majority of cases. The most prominent cases are `podio::ObjBase`, which is covered by #158.

### `Wsign-compare`
Mostly related to the fact that `podio::ObjectID` uses two integers to keep track of the objects, but this is used directly in comparison to `vector::size`. I am not sure what the best fix would be in this case. Having the `invalid` and `untracked` states be negative is a nice feature, on the other hand using `unsigned` members could fix a few of these warnings. (Also proposed in #158 and #157). There are a few things to consider: A `static_cast<unsigned>(-2)` would give us an equally usable "magic" value, corresponding to `std::numeric_limits<unsigned>::max() - 1`. The problem that arises with this approach is that each user facing class has a
```cpp
unsigned int id() const { return getObjectID().collectionID * 10000000 + getObjectID().index; }
```
which would also need to be changed in order to not have a result that overflows, once in the multiplication and then again in the final addition. Even though the result is still well defined, its much harder to interpret than one we would get with negative numbers for the `collectionID` and `index`.

The rest of the warnings are mostly `Wunused-parameter` in some generated classes, some `Wunused-variable` in test code as well as `SIOBlock` as well as a few others of different types scattered around the code base. However, all of these should be easy enough to fix.

In the end I think the goal should be to have a build without warnings, potentially also enforcing that with a `-Werror` in CI.

tmadlener:
BEGINRELEASENOTES
- Fix compiler warnings, that were uncovered by #153 and described in #170. Fix them in the core classes and also in the generated ones.
- Enforce no new warnings with `Werror` in the CI builds.

ENDRELEASENOTES

Fixes #170 

These changes should fix in the core classes:
- all of the `Weffc++` warnings that are not yet covered by #156
- all of the `Wshadow-[parameter,variable]` ~not yet covered by #158 (mainly `podio::ObjBase`)~ (26.01.2021, copied relevant changes)
- the one `Wnon-virtual-dtor` warning in `ColWriterBase`
- the two `Wcatch-value=` warnings in `unittest.cpp`
- the `Wreorder` warnings in the `ROOTWriter`
- the `Wpedantic` warnings due to unnecessary semi colons

In the generated classes they should fix:
- All `Weffc++` warnings for the `Collection` classes, ~except the ones that are not fixable due to the root constraints (see #170)~(28.01.21)
- All `Weffc++` warnings for the `Obj` classes ~except for the ones covered by #158~ (26.01.2021, copied relevant changes)
- All `Weffc++` warnings about uninitialized members in the `Data` classes and in the components. (26.01.2021)
- All `Wshadow` warnings ~(partially overlapping with #158 due to a slightly different approach)~ (26.01.2021, copied the relevant changes)

~These changes **do not yet fix**~
- ~`Weffc++` warnings arising in the generated components. Here, I think we first have to think about how to best handle these.~
- ~The `Wsign-compare` warnings in the core as well as in the generated classes (see also the discussion in #170). Also this needs a bit more discussion on how to proceed.~ (26.01.2021)

Some simple statistics (via `make | sed -nr 's/.+ \[-(.*)\].*/\1/p' | sort | uniq -c`)

#### gcc statistics
Before
```
      2 Wcatch-value=
   2338 Weffc++
      1 Wnon-virtual-dtor
      2 Wpedantic
      3 Wreorder
    366 Wshadow
     97 Wsign-compare
     37 Wunused-parameter
     18 Wunused-variable
```
After
```
   468 Weffc++
```

~All of the `Wshadow` warnings and about a third of the `Weffc++` warnings are fixed by some of the changes that are present in #158 and #156.~ (26.01.2021) All of the remaining `Weffc++` warnings are fixed by #156 and #154.

#### clang statistics
(probably the easier to parse and more reliable output):
Before
```
     15 Wdeprecated-copy-dtor
     16 Winconsistent-missing-override
     22 Wmismatched-tags
      1 Wnon-virtual-dtor
      1 Woverlength-strings
      1 Wreorder-ctor
     23 Wshadow
    101 Wsign-compare
     37 Wunused-parameter
      1 Wunused-private-field
     18 Wunused-variable
```
After
```
      1 Woverlength-strings
```

These are mostly the same warnings under a slightly different name, only one or two are actually warning about things that gcc didn't warn about yet.

- [x] Check `Wsign-compare` appearance in edm4hep (i.e. user-facing code)
- [x] Fix remainging `Weffc++`
  - [x] Default initialize component and datatype members
  - [x] Disable for `Collection`s 
- [x] Fix warnings that appear in builds using llvm

### DIFF
--- a/.github/scripts/compile_and_test.sh
+++ b/.github/scripts/compile_and_test.sh
@@ -13,7 +13,7 @@ cd /Package
 source init.sh
 mkdir build install
 cd build
-cmake -DCMAKE_INSTALL_PREFIX=../install -DCMAKE_CXX_STANDARD=17 -DENABLE_SIO=${ENABLE_SIO} -DCMAKE_CXX_FLAGS=" -fdiagnostics-color=always " -G Ninja .. && \
+cmake -DCMAKE_INSTALL_PREFIX=../install -DCMAKE_CXX_STANDARD=17 -DENABLE_SIO=${ENABLE_SIO} -DCMAKE_CXX_FLAGS=" -fdiagnostics-color=always -Werror" -G Ninja .. && \
 ninja -k0 && \
 ninja install && \
 ctest --output-on-failure
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -24,7 +24,7 @@ jobs:
         source /Users/Shared/cvmfs/sft.cern.ch/lcg/views/${{ matrix.LCG }}/setup.sh
         mkdir build install
         cd build
-        cmake  -GNinja -DCMAKE_INSTALL_PREFIX=../install -DCMAKE_CXX_STANDARD=17  -DCMAKE_CXX_FLAGS=" -fdiagnostics-color=always " ..
+        cmake  -GNinja -DCMAKE_INSTALL_PREFIX=../install -DCMAKE_CXX_STANDARD=17  -DCMAKE_CXX_FLAGS=" -fdiagnostics-color=always -Werror" ..
         ninja -k 0
         ninja install
         ctest --output-on-failure
--- a/cmake/podioMacros.cmake
+++ b/cmake/podioMacros.cmake
@@ -80,6 +80,11 @@ function(PODIO_GENERATE_DICTIONARY dictionary)
     target_sources(${dictionary} PRIVATE ${gensrcdict})
   ENDIF()
   set(gensrcdict ${dictionary}.cxx PARENT_SCOPE)
+  set_source_files_properties(${gensrcdict}
+    PROPERTIES
+    GENERATED TRUE
+    COMPILE_FLAGS "-Wno-overlength-strings"
+    )
 
   #---roottest compability---------------------------------
   if(CMAKE_ROOTTEST_DICT)
--- a/include/podio/ASCIIWriter.h
+++ b/include/podio/ASCIIWriter.h
@@ -24,6 +24,7 @@ namespace podio {
 
   struct ColWriterBase {
     virtual void writeCollection(CollectionBase*, std::ostream& )=0 ;
+    virtual ~ColWriterBase() = default;
   } ;
   
   template <class T>
--- a/include/podio/ASCIIWriter.h
+++ b/include/podio/ASCIIWriter.h
@@ -45,6 +45,10 @@ typedef std::map< std::string, ColWriterBase* > FunMap ;
     ASCIIWriter(const std::string& filename, EventStore* store);
     ~ASCIIWriter();
 
+    // non-copyable
+    ASCIIWriter(const ASCIIWriter&) = delete;
+    ASCIIWriter& operator=(const ASCIIWriter&) = delete;
+
     template<typename T>
     bool registerForWrite(const std::string& name);
     void writeEvent();
--- a/include/podio/ASCIIWriter.h
+++ b/include/podio/ASCIIWriter.h
@@ -14,7 +14,6 @@
 namespace podio {
 
   class CollectionBase;
-  class Registry;
 
   //std::function<void(CollectionBase*)> fun ;
   //std::map< std::string, std::function<void(CollectionBase*)>* > FunMap ;
@@ -59,9 +58,9 @@ typedef std::map< std::string, ColWriterBase* > FunMap ;
 
     std::ofstream* m_file;
 
-    std::vector<CollectionBase*> m_storedCollections;
-    std::vector<std::string> m_collectionNames ;
-    FunMap m_map ;
+    std::vector<CollectionBase*> m_storedCollections{};
+    std::vector<std::string> m_collectionNames{};
+    FunMap m_map{};
   };
 
   // int main () {
--- a/include/podio/BenchmarkRecorder.h
+++ b/include/podio/BenchmarkRecorder.h
@@ -22,6 +22,7 @@ public:
   // Avoid some possible issues that could arise from copying by simply
   // disallowing it
   BenchmarkRecorderTree(const BenchmarkRecorderTree&) = delete;
+  BenchmarkRecorderTree& operator=(const BenchmarkRecorderTree&) = delete;
 
   BenchmarkRecorderTree(TFile* recFile, const std::string& name, const std::vector<std::string>& steps) :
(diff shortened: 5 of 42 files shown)


petricm:
@tmadlener you have to still fix the `brew cask install` call in the mac workflow

tmadlener:
> @tmadlener you have to still fix the `brew cask install` call in the mac workflow

Opened #172 for that, since this one might need a bit longer for review.

tmadlener:
I have implemented the fixes for `Wsign-compare`. In the end this only required very few changes, mainly in the tests and in two places in the `EventStore`.

I have also copied some changes that are currently in #158. I think that can be closed, as all the relevant changes are here now. Some other changes that are present there do not apply here, due to the slightly different approach there (making the members of `podio::ObjectID` unsigned).

With the changes from #154 and #156 only the "unfixable" `Weffc++` in the `Collections` and the uninitialized component and datatype members remain. I could merge those PRs into this one to have this in only one PR, rather than having to deal with it in several different ones.

tmadlener:
Now all members in the `Data` classes are default initialized and `Weffc++` is disabled for the copy-constructor of the `Collection`s. This get's rid of all the warnings in the generated code. The unittests now check for `std::is_standard_layout` and `std::is_trivially_copyable` instead of the deprecated `std::is_pod`. I think these should be the two cases we are most interested in.

CI is currently failing because of `-Werror`. I deliberately put this in, so that #170 can be closed when this is merged.

tmadlener:
Looking at the CI build outputs, there are a few more warnings that appear in the builds using clang.

tmadlener:
I have now also fixed all but one warning in the llvm/clang builds. Additionally, I have significantly reduced the contents of the generated `selection.xml` (see https://github.com/AIDASoft/podio/issues/111#issuecomment-769249615), so that it is now possible to also delete the copy constructor of the generated collections. This in turn means that there is no need for ignoring `Weffc++` for the generated collections any longer, so I have removed that again.

The one warning that is still present for llvm/builds is `Woverlength-strings`, which is triggered by the root generated dictionary: https://github.com/AIDASoft/podio/pull/171/checks?check_run_id=1786088699#step:5:164

I am not entirely sure if and how that could be fixed, but this will only get worse the larger the generated datamodel gets, as basically all generated headers are just dumped into that `DICTPAYLOAD` string. This is basically the only "unfixable" warning left for now.

tmadlener:
I have no added `-Wno-overlength-strings` for the root generated dictionary file, in order to suppress it in all builds (not only the CI ones).

tmadlener:
I just realized that ubuntus gcc7 and gcc8 are *very* picky in their shadowing warnings: https://github.com/AIDASoft/podio/pull/171/checks?check_run_id=1861872804#step:5:182

They complain about the fact that a local loop variable `data`, shadows a class method `data()`. For me this is a false positive, and I don't see a reason to alter the templates in this case. I have instead renamed the member of the corresponding class from `data` to `component`. This will at least make the CI work, but doesn't really solve the underlying problem. Starting from gcc9 this is no longer flagged, and so also users of podio should not run into similar problems with newer compilers.

petricm:
@tmadlener there is a LCG_99 ubuntu2004 gcc9 build

tmadlener:
> @tmadlener there is a LCG_99 ubuntu2004 gcc9 build

That is good to know. Thanks. However, at least for now it seems that these older Ubuntu builds are also surfacing some more legitimate unused variable warnings. Unfortunately they also now have one that is definitely a false positive here: https://github.com/AIDASoft/podio/pull/171/checks?check_run_id=1872871018#step:5:94 both of these are used.

The more legitimate case is the following in ROOTReader:
```cpp
std::pair<TTree*, unsigned> getLocalTreeAndEntry(const std::string&);
auto [tree, dummy] = getLocalTreeAndEntry("evt_metadata");
// only use tree here, but not dummy
```
where `dummy` is not used. Unfortunately "ignoring" parts of the structured binding is not (yet) possible in c++. The easiest solution in this case is to simply not do the structured binding, and instead access the pair with `.first` directly.

Alternatively we could switch to newer versions of Ubuntu and gcc. That should hopefully also fix the false positive.

tmadlener:
> Unfortunately they also now have one that is definitely a false positive here: https://github.com/AIDASoft/podio/pull/171/checks?check_run_id=1872871018#step:5:94 both of these are used.

Just for completeness, these were actually accurate warnings, but they were attributed to the wrong line in the source file. They were flagging https://github.com/AIDASoft/podio/blob/72e472cd3460c0b486a6514bcedbfdc32f0a2056/src/ROOTWriter.cc#L94
but should instead have flagged L99 in the same file.

Now also these are fixed, and I think this PR is now ready for review and eventual merging.

hegner (commented):
thanks!

PR #172, opened 2021-01-19
Fix deprecated brew cask install

tmadlener:
BEGINRELEASENOTES
- Fix deprecated `brew install` commands in mac workflow

ENDRELEASENOTES

`brew cask install` has been deprecated in 2.6.0 and removed in 2.7.0. See https://brew.sh/2020/12/01/homebrew-2.6.0/ and Homebrew/brew/pull/8899

Together with a similar fix in the cvmfs-github-action (cvmfs-contrib/github-action-cvmfs#10) this should fix the mac CI workflows.

### DIFF
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -17,8 +17,8 @@ jobs:
     - name: Install LCG dependencies
       run: |
         brew install ninja
-        brew cask install gfortran
-        brew cask install xquartz
+        brew install --cask gfortran
+        brew install --cask xquartz
     - name: Compile and test
       run: |
         source /Users/Shared/cvmfs/sft.cern.ch/lcg/views/${{ matrix.LCG }}/setup.sh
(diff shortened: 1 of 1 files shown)


tmadlener:
I am not sure what the difference is between `brew install --cask gfortran` and simply `brew install gfortran` for the mac workflows, but the latter has worked on the first try, while the former has worked in one workflow but not the other. However, I don't see any difference in the workflows up until that point, so I am not entirely sure whether it was really the change that made the difference or if there was another problem causing this.

petricm:
As far as I understood the cask was resolving [this](https://github.com/fxcoudert/gfortran-for-macOS/releases/) and the other is resolving the whole gcc suite which is at it seems anyway installed. This was needed in the past to satisfy the view dependency at initialize step (dependency of R), but in reality this is not relevant for podio at all....

petricm (approved):

Issue #168, opened 2020-12-28
 Parse multiple yaml files at the same time


liqiyun:
As described above, does podio support parsing multiple files at the same time ? Or say, can a yaml file be included by another yaml file ?

tmadlener:
Hi @liqiyun,

At the moment podio only supports parsing one yaml file, which cannot include other yaml files.
In principle such a use case could be solved, by first merging all the different yaml files into one and then process that with podio.

What would be your use case for which parsing multiple yaml files would be necessary?

liqiyun:
It will be convenient to switch between different design scheme of data models.

SanghyunKo:
Hi @liqiyun,

I was trying similar attempts, with a slightly different motivation from yours - in my case what I was trying to do is adding several additional event data to already existing (compiled) Podio data model.

I assume feeding multiple yaml files would be possible even without 'concatenating' yaml files into a single one if you simply build Podio data models into separated CMake package and link them to your application simultaneously. 

In the case, what you need to take care is package namespace and including headers properly (if either of data model is not independent each other), since the first argument of `PODIO_GENERATE_DATAMODEL` will be the namespace, include directory name and so on - so you need to include headers by yourself if you build your model into separated packages. 
```
ExtraCode:
  includes:
    "#include <namespace/header1.h>\n #include <namespace/header2.h>"
```
But I'm not sure whether authors would encourage/discourage this kind of use case...

tmadlener:
Hi @SanghyunKo,

That is definitely an interesting approach, thanks for pointing it out. I think this should be a supported use case, but maybe there are some things that still could be done to refine this.

You say that you put the second datamodel yaml into a separate CMake package, but I don't think that is strictly necessary. Have you tried to generate the second datamodel in the same CMake package and then simply link it against the first? Something along the lines of:
```cmake
# main datamodel
PODIO_GENERATE_DATAMODEL(datamodel1 datamodel1.yaml dm1_headers dm1_sources)
PODIO_ADD_DATAMODEL_CORE_LIB(datamodel1 "${dm1_headers}" "${dm1_sources}")
# similar for the ROOT dicts and SIO blocks if necessary

# second datamodel using parts of the first one. Generation should be unaffected
PODIO_GENERATE_DATAMODEL(datamodel2 datamodel2.yaml dm2_headers dm2_sources)
PODIO_ADD_DATAMODEL_CORE_LIB(datamodel2 "${dm2_headers}" "${dm2_sources}")
[...]

PR #175, opened 2021-02-01
Fully qualify relation return types in generated .cc files

tmadlener:
BEGINRELEASENOTES
- Fully qualify return types for `OneToOneRelation` getters in generated .cc file for objects and `Const` objects. This fixes a bug described in https://github.com/AIDASoft/podio/issues/168#issuecomment-770751871 and now allows to mix different namespaces in the generated code. This allows to more easily extend already existing datamodels by compiling and linking against them.

ENDRELEASENOTES

Additionally, cleaned up the templates for the `OneToOneRelation` setters.


tmadlener:
CI seems to be having problems with the `github-action-cvmfs`, which seems to be caused by problems accessing the cvmfs package repository at cern: https://github.com/AIDASoft/podio/pull/175/checks?check_run_id=1806087593#step:3:60

petricm:
only a transient connection problem

tmadlener:
triggering CI again.

Issue #136, opened 2020-09-16
Broken constness in meta data handling in EventStore


hegner:
The recently added methods for meta data handling violate the concept of constness and thread-safety. Despite being declared as mutable internally, the return values of the meta data getters have to be const as well. Otherwise, multithreaded users of PODIO do not have a chance to implement concurrency. 

Maybe the design of the meta data handling needs to be rethought?

PR #177, opened 2021-02-10
[WIP] Implement reference collections

### LINKED ISSUE #146: Idea: True reference collections
I think podio should support _true_ reference collections. Currently something similar can be achieved by defining a helper data type, a la
```yaml
DataType:
  Members:
    - int member // just some exemplary member

DataTypeReference:
  Members: {}
  OneToOneRelations:
    - DataType referenced // reference to the actual datatype
```

However, this is slightly awkward in its usage, as there is always an additional indirection to get to the actual information

```cpp
auto& referenceCollection = eventStore.get<DataTypeReferenceCollection>("refCollection");
auto ref  = referenceCollection[0];
auto data = ref.getReferenced();
int i = data.getMember();
```

Also setting this is slightly cumbersome
```cpp
auto& dataCollection = eventStore.create<DataTypeCollection>("dataCollection");
auto& referenceCollection = eventStore.create<DataTypeReferenceCollection>("refCollection");

auto data = dataCollection.create();
auto ref = referenceCollection.create();
ref.setReferenced(data);
```

Ideally, reference collections behave the same as normal collections when they are `const`, i.e. reference collections do not allow to alter the objects they reference in any way, but can be used for _read only_ access to them. So something along the lines:

```yaml
DataType:
  Members:
    - int member // just some exemplary member
```

for which podio would generate a `DataTypeCollection` and a `DataTypeRefCollection` (naming is just a proposal here, also to distinguish it from the above example more clearly). So that in the end the usage would look something like this

```cpp
// when writing
auto& dataCollection = eventStore.create<DataTypeCollection>("dataCollection");
auto& referenceCollection = eventStore.create<DataTypeRefCollection>("refCollection");

auto data  = dataCollection.create();
referenceCollection.push_back(data);
// changes to data should still be properly propagated to the reference object


// when reading
auto& referenceCollection  = eventStore.get<DataTypeRefCollection>("refCollection");
auto data = referenceCollection[0];
int i = data.getMember();
```

I am not yet completely sure about the internals, but I think something like this should be possible. From a persistency point of view it should in principle work like any other `OneToOneRelation` and storing a `vector<ObjectID>` should be enough to rebuild the reference collection after reading it in. One of the major changes that would be necessary is probably to change the layout of the `Obj` classes from storing a `Data` object to only having a pointer to a `Data` object. I have already briefly mentioned this earlier in https://github.com/AIDASoft/podio/pull/129#issue-474854578 . However, since then I haven't really checked what would actually be necessary to make that change work.

This is at the moment just an idea and an early draft of an implementation, so any input is very welcome.

### LINKED ISSUE #176: Indexed collection access is not const-correct
Currently the indexed access to collection looks like this
```cpp
class FooCollection {
public:
// ...
const Foo operator[](unsigned) const;
Foo operator[](unsigned);
//  same again for at(unsigned)
// ...
}
```

This is not const-correct, as it allows to get a mutable handle from a collection, e.g.
```cpp
auto store = podio::EventStore();
const auto& collection = store.get<FooCollection>("collection");
auto foo = collection[0];
foo.setValue(12345); // this should not work, but does at the moment
```

Here we are breaking const-correctness, because we are allowed to change an element inside a collection that is `const`.

tmadlener:
BEGINRELEASENOTES
- Implement true reference collections, that can collect objects that are already tracked by another collection. Each datatype now gets a `RefCollection` class that behaves similar to a `Collection` with the major difference, that it does not own its elements, and does not offer `create` functionality. It can only store elements that are already tracked by a "proper" `Collection`.

ENDRELEASENOTES

~~**Contains all changes of #180 !**~~

### reference collections
The basic idea is to generate an additional `RefCollection` class, which only holds a container of `Obj*` and gives similar access as a normal `Collection`. All objects that are stored in a `RefCollection` have to be already tracked by a `Collection`, as the `RefCollection` does no management of resources in any form. Since it doesn't manage any resources, it is also lacking the `create` functionality. To persist it's contents it simply fills a vector of `podio::ObjectID`, following the same approach as `OneToOneRelation`s.

~~### const correct collection access~~
~~Introduce appropriate overloads for `operator[]` and `at` functions. Additionally introduce a new `ConstCollectionIterator` alongside the `CollectionIterator`. The former only gives access to `Const` classes, while the latter gives access to the normal classes.~~ Moved to #193

~~### const correct meta data handling in EventStore~~
~~Since the `EventStore` is never const at the moment, I have removed all the `const` qualifiers from the methods giving access to any kind of metadata. This makes it impossible to use metadata from a `const EventStore`, but we do not have that anywhere at the moment. I think this is the better solution that giving the wrong impression that we have actually handled metadata in a const-correct way.~~ Moved to #192

fixes #146

- [x] generate additional `RefCollection` classes.
- [x] Enable all readers / writers to read and write these collections again

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,7 @@ spack*
 *.exe
 *.out
 *.app
+
+# clangd cache and other LSP related files
+/.clangd/
+compile_commands.json
--- a/include/podio/CollectionBase.h
+++ b/include/podio/CollectionBase.h
@@ -51,6 +51,9 @@ namespace podio {
     /// fully qualified type name of elements - with namespace
     virtual std::string getValueTypeName() const = 0;
 
+    /// Is this a reference collection
+    virtual bool isReferenceCollection() const = 0;
+
     /// destructor
     virtual ~CollectionBase() = default;
 
--- a/include/podio/ROOTReader.h
+++ b/include/podio/ROOTReader.h
@@ -80,6 +80,8 @@ class ROOTReader : public IReader {
 
 
   private:
+    void createCollectionBranches(const std::vector<std::tuple<int, std::string, std::string>>& collInfo);
+
     std::pair<TTree*, unsigned> getLocalTreeAndEntry(const std::string& treename);
     // Information about the data vector as wall as the collection class type
     // and the index in the collection branches cache vector
--- a/include/podio/SIOBlock.h
+++ b/include/podio/SIOBlock.h
@@ -47,7 +47,7 @@ namespace podio {
     virtual SIOBlock* create(const std::string& name) const=0 ;
 
     // create a new collection for this block
-    virtual void createCollection() = 0;
+    virtual void createCollection(const bool referenceCollection=false) = 0;
 
   protected:
 
@@ -74,11 +74,13 @@ namespace podio {
 
     podio::CollectionIDTable* getTable() { return _table; }
     const std::vector<std::string>& getTypeNames() const { return _types; }
+    const std::vector<short>& getRefCollectionBits() const { return _isRefColl; }
 
   private:
     podio::EventStore* _store{nullptr};
     podio::CollectionIDTable* _table{nullptr};
     std::vector<std::string> _types{};
+    std::vector<short> _isRefColl{};
   };
 
 
@@ -130,7 +132,7 @@ namespace podio {
     std::shared_ptr<SIOBlock> createBlock( const podio::CollectionBase* col, const std::string& name) const;
 
     // return a block with a new collection (used for reading )
-    std::shared_ptr<SIOBlock> createBlock( const std::string& typeStr, const std::string& name) const;
+    std::shared_ptr<SIOBlock> createBlock( const std::string& typeStr, const std::string& name, const bool isRefColl=false) const;
 
     static SIOBlockFactory& instance() {
       static SIOBlockFactory me ;
--- a/include/podio/SIOReader.h
+++ b/include/podio/SIOReader.h
@@ -85,6 +85,7 @@ namespace podio {
     int m_eventNumber{0};
     int m_lastEventRead{-1};
     std::vector<std::string> m_typeNames{};
+    std::vector<short> m_refCollectionBits{};
 
     std::shared_ptr<SIOEventMetaDataBlock> m_eventMetaData{};
     std::shared_ptr<SIONumberedMetaDataBlock> m_runMetaData{};
--- a/python/podio_class_generator.py
+++ b/python/podio_class_generator.py
@@ -207,6 +207,7 @@ class ClassGenerator(object):
     self._fill_templates('ConstObject', datatype)
     self._fill_templates('Obj', datatype)
     self._fill_templates('Collection', datatype)
+    self._fill_templates('RefCollection', datatype)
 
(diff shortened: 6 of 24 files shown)


tmadlener:
Currently some of the changes in this break with root 6.20/04. Will have to check what needs to be done to make the changes that fix the const-correctness problems in the collections, work with this version of ROOT (and presumably also earliers), unless we drop support for it.

hegner:
Hi Thomas! Will merging #180 fix the mentioned issues? Or is there more needed?

tmadlener:
#180 will not fix any of the issues that are mentioned here. Instead this PR is started on top of #180, because some of the work done there proved to be going into the right direction already for the reference collections (#146). Since I had to touch the templates I found that some of our collection access methods are not const-correct (#176) and fixed them right away as well. Additionally the fixes for const-correct metadata (#136) are not really fixes, but at least we are no longer making false promises. It has become a bit convoluted in the end, but at least parts could probably be lifted into separate and easier to digest PRs.

hegner:
@tmadlener - where are we with this now? are there still things can one pull out from this?

tmadlener:
Hi @hegner, 

I think the commit for making the meta data access const-correct could be salvaged trivially. However, as already mentioned above it is less of a fix and more a "removal of false promises".

Potentially the const-correct collection access, both index and iterator based, could be lifted from here as well, but I would have to check if that is as trivial as cherry picking a few commits. In any case, I am not entirely sure now how to best deal with the python side of the whole thing, but that is something that could be checked in a different pull request.

Finally, the reference collections are technically working as they are done here, but I don't like the approach I have chosen here very much any longer, because it introduces a separate collection class, and we might be able to do better with the separation of storage and collection interface that we have planned.

Shall I try and pull this apart in two smaller PRs for the const-correct meta data handling, and the const correct collection access for easier review?

hegner:
Yes, try to split it up and then we take it from there. Thanks

tmadlener:
I have rebased this onto the current master to make it possible to have a look at this. 

In its current form the main usage from k4SimDelphes would change from something like:
```cpp
auto& recos = store.create<ReconstructedParticleCollection>("recos");
// fill ...

// reference objects of the former
auto& refCollection = store.create<RecoParticleRefCollection>("muons");
auto ref = refCollection.create();
ref.setParticle(recos[0]);
```

to 
```cpp
auto& recos = store.create<ReconstructedParticleCollection>("recos");
// fill ...

// reference objects of the former
auto& refCollection = store.create<ReconstructedParticleRefCollection>("muons");
refCollection.push_back(recos[0]);
```

On the reading side things would change from
```cpp
auto& muons = store.get<RecoParticleRefCollection>("muons");
auto muon = muons[0].getParticle();
```
to
```cpp
auto& muons = store.get<ReconstructedParticleRefCollection>("muons");
auto muon = muons[0];
```

The main thing I am not yet fully satisfied with yet is the fact that the reference collections and the normal collections are distinct types, even though they offer basically the same interface. This is slightly inconvenient in usage, because functions that want to use collections and reference collections have to be template functions, and additionally because the user needs to know the type of the collection he wants to get from the EventStore. Ideally both would have the same type, at least when reading the collections.

tmadlener:
superseded by #197

PR #180, opened 2021-02-19
Root performance degradation fixes

### LINKED ISSUE #179: Performance degradation for root 6.22/06 compared to 6.20/04
There seems to be a quite serious performance degradation between root 6.20/04 and 6.22/06.

Running one of our benchmark cases using different versions of root yields the following report: https://github.com/tmadlener/podio_benchmarking/tree/c5aebc017b99f572a84082f505958dafbd78b79b/results/k4SimDelphes/compare_root_versions

It can clearly be seen that in total time root 6.22/06 is about a factor 2 slower when reading and writing compared to root 6.20/04. This seems to be entirely driven by much longer per event times (for the easiest comparison look at the plots).

Will have to investigate if this is due to some changed defaults in roots I/O or if this is due to some other reason.

tmadlener:
BEGINRELEASENOTES
- Improve the branch look-up logic in ROOTReader and ROOTWriter. Triggered by a performance degradation in v6.22/06, where this logic was changed inside ROOT and our use case was affected badly. All ROOT versions profit from these changes as it is in general more efficient than the previous implementation.

ENDRELEASENOTES

Need to discuss which of the proposed solutions in https://root-forum.cern.ch/t/serious-degradation-of-i-o-performance-from-6-20-04-to-6-22-06/43584/10 we would like to have in the end. ~Currently this PR only implements the first proposal, but this is something that probably needs to be touched again in the redesign, so this might be fine.~

~Also contains changes from #171. **Do not merge before that**~

Fixes #179

### DIFF
--- a/include/podio/CollectionBranches.h
+++ b/include/podio/CollectionBranches.h
@@ -1,0 +1,22 @@
+#ifndef PODIO_COLLECTION_BRANCHES_H
+#define PODIO_COLLECTION_BRANCHES_H
+
+#include "TBranch.h"
+
+#include <vector>
+
+namespace podio::root_utils {
+/**
+ * Small helper struct to collect all branches that are necessary to read or
+ * write a collection. Needed to cache the branch pointers and avoid having to
+ * get them from a TTree/TChain for every event.
+ */
+struct CollectionBranches {
+  TBranch* data{nullptr};
+  std::vector<TBranch*> refs{};
+  std::vector<TBranch*> vecs{};
+};
+
+}
+
+#endif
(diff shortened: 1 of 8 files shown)


tmadlener:
Added caching of the branch pointers for a marginal speed up compared to the "improved" look up of the branches.

hegner (commented):
Hi Thomas! Thanks. What I am not too sure about is the naming part. But I guess it is too little of work to to justify a podio::root namespace. What do you think?

tmadlener:
I had similar thoughts actually, as you said I thought that opening an extra namespace is probably too much. However, I realize that I do actually open a small extra namespace for some utilities.

In general, I think we could separate the different I/O backends better from each other and from the core parts of podio. E.g. have a dedicated directory where everything for ROOT goes and another one for SIO. I am not sure if we should make those changes here or if we just address them in the upcoming developments, where we have to touch these things in any case.

hegner:
Yes, I think we do it after the re-design. And initially I thought it was less scattered.

hegner (approved):

PR #183, opened 2021-02-25
Use SIO::sio target in cmake

tmadlener:
BEGINRELEASENOTES
- Use SIO targets in cmake, which are exported starting with v00-01 (iLCSoft/SIO#15)

ENDRELEASENOTES


vvolkl (approved):
Very nice, thanks Thomas.

PR #184, opened 2021-03-12
Clean up AsciiWriter comments

vvolkl:
BEGINRELEASENOTES
- Clean up AsciiWriter comments

ENDRELEASENOTES

### DIFF
--- a/include/podio/ASCIIWriter.h
+++ b/include/podio/ASCIIWriter.h
@@ -15,13 +15,6 @@ namespace podio {
 
   class CollectionBase;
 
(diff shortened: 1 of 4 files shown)


tmadlener:
The write_ascii.cpp seems to still be using code that has slipped through in #171. It can probably also reuse the `write` common functionality in `read_tests.h` (with some minor adjustments to only process 1 event).

vvolkl:
Thanks Thomas, as discussed offline, we skip the `write` refactoring, and just fix the error for now.

PR #186, opened 2021-04-06
Add support for fixed width integers

tmadlener:
BEGINRELEASENOTES
- Add support for [fixed width integer type](https://en.cppreference.com/w/cpp/types/integer) members in components and datatypes.
  - Now possible to use `int16_t`, `int32_t`, `int64_t`, `uint16_t`, `uint32_t` and `uint64_t` as members. Other fixed width integer types that are potentially defined in `<cstdint>` are not considered valid as the intended use case is really only fixed width integers for now. These are rejected at the datamodel validation step.
  - Fixed width integers are considered to be "builtin" types for podio.

ENDRELEASENOTES

See key4hep/EDM4hep#112 for a possible use case.

The actually necessary changes are confined to the `generator_utils.py`, the rest is adding a bunch of tests that should cover the major use (and non-use) cases.

### DIFF
--- a/doc/datamodel_syntax.md
+++ b/doc/datamodel_syntax.md
@@ -28,10 +28,11 @@ A component is just a flat struct containing data. it can be defined via:
     components:
       # My example component
       MyComponent:
-        x : float
-        y : float
-        z : float
-        a : AnotherComponent
+        Members:
+          - float x
+          - float y
+          - float z
+          - AnotherComponent a
 ```
 
 ## Definition of custom data classes
@@ -113,7 +114,8 @@ Some customization of the generated code is possible through flags. These flags 
     components:
       # My simple component
       ExampleComponent:
-        x : int
+        Members:
+          - int x
     datatypes:
       ExampleType:
         Description: "My datatype with a component member"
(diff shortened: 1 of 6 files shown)


hegner (changes requested):
Thank you for these changes. Can you along with this feature change update our docs? So that this is properly documented?
Thanks!

hegner (approved):
Thanks!

hegner:
The failing builds seem a problem of the machines and need follow up...

tmadlener:
> The failing builds seem a problem of the machines and need follow up...

Yes, I think that is related to problems with some cvmfs publisher node. This leads to a broken latest release for key4hep.

tmadlener:
And for the `dev4` workflow we seem to pick up another cmake that is not the one that *should* come with `dev4`. Testing it on lxplus it seems that indeed the setup script does not set up any cmake at all and instead we still get the system cmake.

andresailer:
There is nothing in `dev4...gcc10-opt` at the moment besides the setup.sh so not only cmake is missing. Something must have gone wrong during the publication today.

PR #188, opened 2021-04-27
[WIP] [discussion] tooling integration using podio as example

tmadlener:
As discussed during the [Key4hep meeting today](https://indico.cern.ch/event/1032840/), this is a pull request that should serve mainly as a discussion platform to arrive at a sensible selection of tools that we want to use in the end and. Additionally, to arrive at reasonable configurations for tools that require configuration.

The changes are:
- switching to Catch2 v3 and building it on demand as a static library
- Adding a few sanitizers to the cmake setup to make it easier to use them as well as adding github action workflows that build podio with them enabled.
- Adding a basic clang-tidy setup with a .clang-tidy inspired by CMSSW
  - Plus necessary changes
- Adding a basic clang-format setup with a .clang-format inspired by Gaudi
  - Plus necessary changes 
  - Plus commit to make the code generation cmake macro look for a .clang-format in projects that use it
- Adding a basic configuration for pre-commit and a workflow that runs it in CI

The commits should be sufficiently well separated to see which changes would be necessary to make a given tool happy on the podio code base. 

**Tools that are not yet integrated but were mentioned during the discussion** (and will arrive here as well)
- [ ] [include-what-you-use](https://include-what-you-use.org/)
- [ ] clang coverage and other sanitizers also available
    - [ ] https://clang.llvm.org/docs/SanitizerCoverage.html
    - [ ] https://clang.llvm.org/docs/SafeStack.html
    - [ ] https://llvm.org/docs/LibFuzzer.html
    - [ ] https://clang.llvm.org/docs/ControlFlowIntegrity.html


andresailer (commented):
my 2 cents

petricm:
You can ditch the `caniusepython3.pylint_checker` in favor of running python3 pylint

include/podio/BenchmarkRecorder.h
```diff
@@ -35,32 +34,27 @@ class BenchmarkRecorderTree {
...
   void recordTime(const std::string& stepName, const ClockT::duration time) {
     const auto it = std::find(m_stepNames.cbegin(), m_stepNames.cend(), stepName);
     const auto index = std::distance(m_stepNames.cbegin(), it);
     m_stepTimes[index] = std::chrono::duration_cast<TimingResolution>(time).count();
   }
 
-  void Fill() {
-    m_recordTree->Fill();
-  }
+  void Fill() { m_recordTree->Fill(); }
```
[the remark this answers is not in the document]
tmadlener: I agree and this controlled by the `AllowShortFunctionsOnASinlgeLine` option ([all options](https://clang.llvm.org/docs/ClangFormatStyleOptions.html), unfortunately there seems to be no way to link directly to one, so a Ctrl+F is still necessary).

There is the possibility to have this enabled if it appears inside class definitions (`InlineOnly` or `Inline`) but have it disabled outside, which would make it possible to have "trivial" single line calls on the same line class definitions. If more than that is happening in such a call one could make an argument to move it to a dedicated implementation in any case. It can also be disabled everywhere (`None`). I have no strong preference here on whether we want to allow this inside class definitions.

src/EventStore.cc
```diff
@@ -1,158 +1,144 @@
...
-    } else {
-      // collection already requested in recursive call
-      // do not set the references to break collection dependency-cycle
-      auto name = m_table->name(id);
-      success = doGet(name, collection,false);
+}
+
+bool EventStore::get(int id, CollectionBase*& collection) const {
+  // see if we have a cached collection
+  if ((collection = getFast(id)) != nullptr) return true;
```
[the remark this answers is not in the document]
tmadlener: Yes, can be done with `clang-tidy` via [`readability-braces-around-statements`](https://clang.llvm.org/extra/clang-tidy/checks/readability-braces-around-statements.html).

include/podio/BenchmarkRecorder.h
```diff
@@ -35,32 +34,27 @@ class BenchmarkRecorderTree {
     }
   }
 
-  template<typename TimingResolution=std::chrono::nanoseconds>
+  template <typename TimingResolution = std::chrono::nanoseconds>
```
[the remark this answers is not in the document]
tmadlener: It seems that in this case we are in an "all or nothing" situation. I haven't found a way to control the behavior of assignment operator spacing depending on the context where it appears.

The space between `template` and `<` can be controlled via `SpaceAfterTemplateKeyword`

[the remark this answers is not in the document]
andresailer: Thanks for checking! Then totally fine with things as the `template <... = ...>`

.pre-commit-config.yaml
```diff
@@ -19,7 +19,7 @@ repos:
         language: system
       - id: pylint
         name: pylint
-        entry: 'pylint --rcfile=.github/scripts/podio.pylint.py3k.rc --py3k --msg-template="{path}:{line}: [{msg_id}({symbol}), {obj}] {msg}"'
+        entry: 'pylint --py3k --msg-template="{path}:{line}: [{msg_id}({symbol}), {obj}] {msg}"'
```
petricm, line 22: `--py3k` disables all the "normal" pylint warnings, and it would be good to have a .rc file someline along the lines of [this](https://github.com/AIDASoft/DD4hep/blob/master/.github/scripts/DD4hep.pylint.rc)

tmadlener: I see. I have never actually looked to closely into the pylint call to be honest. However, simply removing the `--py3k` already makes pylint complain a lot more ;) On the other hand the configuration from DD4hep you linked seems to be incompatible with the default `flake8` options, mainly related to formatting from a quick glance. So we will have to at least find a consistent configuration for both.

petricm: yes, you need also the `setup.cfg` e.g. https://github.com/AIDASoft/DD4hep/blob/master/setup.cfg

tmadlener: With the right configuration (taken from ilcdirac) and fixing the things pylint complains about everything seems to fine.

include/podio/ASCIIWriter.h
```diff
@@ -1,81 +1,85 @@
-#ifndef ASCIIWRITER_H
-#define ASCIIWRITER_H
+#ifndef PODIO_ASCIIWRITER_H
+#define PODIO_ASCIIWRITER_H
```
vvolkl, line 2: As mentioned by Frank:  Spelling `ASCIIWRITER` like the classname here (`PODIO_ASCIIWriter_H`) would make search and replace operations a bit more convenient

tmadlener: Unfortunately the [llvm-header-guard check](https://clang.llvm.org/extra/clang-tidy/checks/llvm-header-guard.html) is not configurable and simply enforces the [LLVM header guard style](https://llvm.org/docs/CodingStandards.html#header-guard), and there doesn't seem to be anything similar with more freedom. So our choices are:
- Drop the check, resulting in no check for the header guards.
- Keep the check, and have more work in search and replace.

I, personally, would go for the first option here, since this is one of the checks that can be automatically fixed by `clang-tidy` (using the `--fix` option when calling `clang-tidy`). However, if more people prefer the second option removing this check is trivial.

andresailer: I agree with Thomas. This automatic check and fix means one doesn't have to replace them themselves. And automatic unique header guards prevent real issues.

Also: emacs by default does a case aware replacement, so it would replace an all upper case word by an all upper case word. (CamelCasing is more an issue though)
```
ASCIIWRITER
ASCIIWriter

replace asciiwriter -> UTF8Reader gives

UTF8READER
UTF8Reader
```

andresailer: Real world issue: https://github.com/AIDASoft/DD4hep/pull/672

gaede: Fine with me. Just had gotten to use the same camel case for the header guards as in the class name for search and replace. We should than activate the `--fix` for the header guards...

tmadlener: The `--fix` option is "global" for `clang-tidy`, so it will try to fix all the things that it can fix. In general, during CI we are running all checkers with the automatic fixes turned off. But the main point is that it is possible to run `clang-tidy` and `clang-format` locally in a way that fixes are applied automatically.

tmadlener:
Superseeded by
- #249 
- #253
- #254

Closing this since all of them have been merged now.

PR #194, opened 2021-06-01
Add indexed access to the RelationRange

tmadlener:
BEGINRELEASENOTES
- Make it possible to do indexed access on a `RelationRange`, making the interface more akin to a `const std::vector`

ENDRELEASENOTES

Inspired by key4hep/EDM4hep#111

With these changes it becomes possible to do
```cpp
const auto& mcs = store.get<edm4hep::MCParticleCollection>("mcparticles");
auto particle = mcs[0];

auto parents = particle.getParents();
auto firstParent = parents[0]; // this was not possible before

for (auto par : parents) { /* do something with par. This was possible before already */ }
```

### DIFF
--- a/include/podio/RelationRange.h
+++ b/include/podio/RelationRange.h
@@ -16,19 +16,36 @@ namespace podio {
 
     RelationRange() = delete;
 
-    RelationRange(ConstIteratorType begin, ConstIteratorType end) : m_begin(begin), m_end(end) {}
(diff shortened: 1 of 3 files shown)


gaede (approved):
Nice.

hegner:
Indeed very nice. In particular having both operator[] and at :-)

PR #197, opened 2021-06-14
Disentangle storage concerns from collection and implement subset collections

### LINKED ISSUE #146: Idea: True reference collections

### LINKED ISSUE #166: Variadic create function for Collection breaks with Objects that have relations
The variadic `create` function of the Collections:
https://github.com/AIDASoft/podio/blob/cf3d239cbc4932e2b1d7ee42bcbef55ee869b3c1/python/templates/Collection.h.jinja2#L164-L170

is broken for all collections that store objects which have relations. E.g. the following will break
```cpp
auto store = podio::EventStore();
auto& clusters = store.create<ExampleClusterCollection>("clusters");
auto variadic_cluster = clusters.create(3.14f);
auto normal_cluster = clusters.create();
normal_cluster.energy(42);

// this will break and crash the program
variadic_cluster.addClusters(normal_cluster);
```

The reason is that the constructor for `Obj` that starts from a `Data` struct does not initialize the relation pointers. Especially for OneToManyRelations this is a problem, because it expects a valid `std::vector` pointer to be present in the `Obj`.

The fix would not be too hard, but the other question is whether we want to support this variadic template `create` method in the first place, since it relies on the order of the members `Data` and together with implicit conversions, this is a pretty dangerous constructor that could lead to very subtle bugs.

tmadlener:
BEGINRELEASENOTES
- Introduce a `podio::CollectionBuffers` class that contains everything that is necessary for I/O of a given collection. **This is a breaking change in the collection interface** 
- Introduce  and generate a `CollectionData` class for each datatype that only manages the storage of a given collection.
  - Exposes only the `Obj` entries of each collection as well as the necessary functionality to add a new object (and its relations) to the collection.
- Implement "subset" collections that behave exactly the same as normal collections apart from an additional function call when creating them.

ENDRELEASENOTES

The `CollectionBuffers` helpers class makes the interaction between the readers/writers and the collections a bit clearer as it now happens in one function call instead of the three it took previously to connect all the buffers to the I/O system. This is a breaking change in the collection interface, but it is quite easy to fix on the calling side, where we now have
```cpp
const auto collBuffers = coll->getBuffers();

// the data buffer (address)
void* rawbuffer = collBuffers.data; // previously was coll->getBufferAddress();
const auto* buffer = collBuffers.dataAsVector<Data>(); // get a vector<Data>* if we know the datatype
// relations and vector members
const auto* refCollections = collBuffers.references; // previously was coll->referenceCollections();
const auto* vecMembers = collBuffers.vectorMembers; // previously was coll->vectorMembers();
```

The `CollectionData` classes manage everything that concerns the storage of a collection. They only expose the `entries` member to give a collection access to the entries of a collection. The `prepareAfterRead`, `prepareBeforeWrite`, `setReferences`  and `getBuffers` functions of the `Collection` are now essentially only passing through to the `CollectionData` which does all the heavy lifting. This now make the storage concerns separated enough from the collection interface that subset (aka "true reference") collections can be implemented without having to introduce a separate `RefCollection` class. The interface is the following:

```cpp
// FILLING
auto& recos = store.create<edm4hep::ReconstructedParticleCollection>("recos");
// fill

auto& muons = store.create<edm4hep::ReconstructedParticleCollection>("muons");
muons.setSubsetCollection(); // declare this as a subset collection
// Here I can only store objects already tracked by another collection
muons.push_back(recos[0]);

// READING
auto& muons = store.get<edm4hep::ReconstructedParticleCollection>("muons");
for (auto muon : muons) { /* use as ConstReconstructedParticle */ }
```

So when creating a subset collection one has to declare it as such with a separate function call, but when reading they behave as if they were a normal collection. The function call to declare a collection to be a subset collection could potentially be put into the `EventStore::create` call somehow if it is deemed too cumbersome.

Supersedes #177
Fixes #146, Fixes #166

- [x] Subset collections
- [x] Test with edm4hep
- [x] Clean up templates (and generated code)
- [x] Clean up commits

### DIFF
--- a/include/podio/CollectionBase.h
+++ b/include/podio/CollectionBase.h
@@ -1,29 +1,24 @@
 #ifndef COLLECTIONBASE_H
 #define COLLECTIONBASE_H
 
+#include "podio/ObjectID.h"
+#include "podio/CollectionBuffers.h"
+
 #include <string>
 #include <utility>
 #include <vector>
 
-#include "podio/ObjectID.h"
+
 
 namespace podio {
   // forward declarations
-  class ObjectID;
   class ICollectionProvider;
-  class CollectionBase;
-
-  using CollRefCollection = std::vector<std::vector<podio::ObjectID>*>;
-  using VectorMembersInfo = std::vector<std::pair<std::string, void*>>;
 
   class CollectionBase {
   public:
     /// prepare buffers for serialization
-    virtual void  prepareForWrite() = 0;
+    virtual void prepareForWrite() = 0;
 
-    //virtual void  write(CollectionBuffer& buffer) = 0;
-    //virtual void  read(CollectionBuffer& buffer) = 0;
-    
     /// re-create collection from buffers after read
     virtual void  prepareAfterRead() = 0;
 
@@ -36,11 +31,10 @@ namespace podio {
     /// get collection ID
     virtual unsigned getID() const = 0;
 
-    /// set I/O buffer
-    virtual void  setBuffer(void*) = 0;
+    /// Get the collection buffers for this collection
+    virtual podio::CollectionBuffers getBuffers() = 0;
 
-    /// get address of the pointer to the I/O buffer
-    virtual void* getBufferAddress() = 0;
+    // virtual void setBuffers(const podio::CollectionBuffers& buffers) = 0;
 
     /// check for validity of the container after read
     virtual bool isValid() const = 0;
@@ -56,12 +50,6 @@ namespace podio {
 
     /// clear the collection and all internal states
     virtual void clear() = 0 ;
-
-    /// return the buffers containing the object-relation information
-    virtual CollRefCollection* referenceCollections() = 0;
-
-    /// return pointers to vector members
-    virtual VectorMembersInfo* vectorMembers() = 0;
   };
 
 } // namespace
--- a/include/podio/CollectionBase.h
+++ b/include/podio/CollectionBase.h
@@ -34,8 +34,6 @@ namespace podio {
     /// Get the collection buffers for this collection
     virtual podio::CollectionBuffers getBuffers() = 0;
 
-    // virtual void setBuffers(const podio::CollectionBuffers& buffers) = 0;
-
     /// check for validity of the container after read
     virtual bool isValid() const = 0;
 
--- a/include/podio/CollectionBase.h
+++ b/include/podio/CollectionBase.h
@@ -50,6 +50,12 @@ namespace podio {
 
     /// clear the collection and all internal states
     virtual void clear() = 0 ;
+
+    /// check if this collection is a reference collection
+    virtual bool isReferenceCollection() const = 0;
+
+    /// declare this collection to be a reference collection
+    virtual void setReferenceCollection(bool setToRef=true) = 0;
   };
 
 } // namespace
--- a/include/podio/CollectionBase.h
+++ b/include/podio/CollectionBase.h
@@ -51,11 +51,11 @@ namespace podio {
     /// clear the collection and all internal states
     virtual void clear() = 0 ;
 
-    /// check if this collection is a reference collection
-    virtual bool isReferenceCollection() const = 0;
+    /// check if this collection is a subset collection
+    virtual bool isSubsetCollection() const = 0;
 
-    /// declare this collection to be a reference collection
-    virtual void setReferenceCollection(bool setToRef=true) = 0;
+    /// declare this collection to be a subset collection
+    virtual void setSubsetCollection(bool setSubset=true) = 0;
   };
 
 } // namespace
--- a/include/podio/ROOTReader.h
+++ b/include/podio/ROOTReader.h
@@ -80,6 +80,8 @@ class ROOTReader : public IReader {
 
 
   private:
+    void createCollectionBranches(const std::vector<std::tuple<int, std::string, bool>>& collInfo);
+
     std::pair<TTree*, unsigned> getLocalTreeAndEntry(const std::string& treename);
     // Information about the data vector as wall as the collection class type
     // and the index in the collection branches cache vector
--- a/python/templates/CollectionData.cc.jinja2
+++ b/python/templates/CollectionData.cc.jinja2
@@ -1,0 +1,157 @@
+{% import "macros/utils.jinja2" as utils %}
+{% import "macros/collections.jinja2" as macros %}
+// AUTOMATICALLY GENERATED FILE - DO NOT EDIT
+
+#include "{{ incfolder }}{{ class.bare_type }}CollectionData.h"
+
+{% for include in includes_coll_cc %}
+{{ include }}
+{% endfor %}
+
+{{ utils.namespace_open(class.namespace) }}
+{% with class_type = class.bare_type + 'CollectionData' %}
+
+{{ class_type }}::{{ class_type }}() :
+{%- for relation in OneToManyRelations + OneToOneRelations %}
+  m_rel_{{ relation.name }}(new std::vector<{{ relation.namespace }}::Const{{ relation.bare_type }}>()),
+{%- endfor %}
+{%- for member in VectorMembers %}
+  m_vec_{{ member.name }}(new std::vector<{{ member.full_type }}>()),
+{%- endfor -%}
+  m_data(new {{ class.bare_type }}DataContainer()) {
+{% for relation in OneToManyRelations + OneToOneRelations %}
+  m_refCollections.push_back(new std::vector<podio::ObjectID>());
+{% endfor %}
+{% for member in VectorMembers %}
+  m_vecmem_info.emplace_back("{{ member.full_type }}", &m_vec_{{ member.name }});
+{% endfor %}
+}
+
+{{ class_type }}::~{{ class_type }}() {
+  delete m_data;
+{% if OneToManyRelations or OneToOneRelations %}
+  for (auto& pointer : m_refCollections) delete pointer;
+{% endif %}
+{% for relation in OneToManyRelations + OneToOneRelations %}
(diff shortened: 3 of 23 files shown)


tmadlener:
This is ready for review now. It has become quite big, because it also incorporates reference collections already. That could be moved to a separate PR if necessary. However, since this splitting of concerns is also at least partially driven by the reference collections, I thought it is good to have these two together.

andresailer:
There is already a function called `referenceCollections`
https://github.com/AIDASoft/podio/blob/02d613f69dde4f8c571e0084e6036ea89ac048db/python/templates/Collection.h.jinja2#L87
Which is not related to the reference collection discussed here, is it?

tmadlener:
> Which is not related to the reference collection discussed here, is it?

That is correct. I think the naming is a bit confusing here as *reference* is used slightly differently in different contexts.

`referenceCollections` returns the `ObjectID`s for all `OneToManyRelations` and `OneToOneRelations` that are persisted for a given collection. `podio::CollRefCollection` is just a typedef:
https://github.com/AIDASoft/podio/blob/eb298b1c0e935c8ba18a8c5726328846ccfbaa89/include/podio/CollectionBase.h#L16

On the other hand the *reference collections* this PR is about, are collections that hold references to objects that are owned by a different collection.

Maybe this can be addressed with a more consistent naming? E.g. by renaming `CollectionBuffers::references` to `CollectionBuffers::relations` (`referenceCollections` will be gone in any case after this PR).

tmadlener:
As discussed during the last meeting, I have renamed the reference collections to subset collections, following the same naming as in LCIO to try and reduce the confusion.

hegner:
Hi @tmadlener - what is the current status of this? I see the item of code cleanup open.

tmadlener:
From a quick look, there still are some comments in the templates that can be removed. I will have another look at this to see if there is more. Other than that, basically done, unless there is additional feedback.

tmadlener:
I have replaced the last spurious occurrences of "reference collections" with "subset collections" and also removed commented code from the templates.

include/podio/CollectionBase.h
```diff
@@ -57,11 +51,11 @@ namespace podio {
     /// clear the collection and all internal states
     virtual void clear() = 0 ;
 
-    /// return the buffers containing the object-relation information
-    virtual CollRefCollection* referenceCollections() = 0;
+    /// check if this collection is a reference collection
```
hegner, line 54: "reference collection"

include/podio/CollectionBase.h
```diff
@@ -57,11 +51,11 @@ namespace podio {
...
+    virtual bool isSubsetCollection() const = 0;
 
-    /// return pointers to vector members
-    virtual VectorMembersInfo* vectorMembers() = 0;
+    /// declare this collection to be a reference collection
```
hegner, line 57: "reference collection"

include/podio/ROOTReader.h
```diff
@@ -80,6 +80,8 @@ class ROOTReader : public IReader {
 
 
   private:
+    void createCollectionBranches(const std::vector<std::tuple<int, std::string, bool>>& collInfo);
+
     std::pair<TTree*, unsigned> getLocalTreeAndEntry(const std::string& treename);
     // Information about the data vector as wall as the collection class type
```
hegner, line 86: ...as well...

python/templates/CollectionData.cc.jinja2
```diff
@@ -0,0 +1,211 @@
...
+{{ class_type }}::{{ class_type }}() :
+{%- for relation in OneToManyRelations + OneToOneRelations %}
+  m_rel_{{ relation.name }}(new std::vector<{{ relation.namespace }}::Const{{ relation.bare_type }}>()),
+{%- endfor %}
+{%- for member in VectorMembers %}
+  m_vec_{{ member.name }}(new std::vector<{{ member.full_type }}>()),
+{%- endfor -%}
+  m_data(new {{ class.bare_type }}DataContainer()) {
+{% for relation in OneToManyRelations + OneToOneRelations %}
+  m_refCollections.push_back(new std::vector<podio::ObjectID>());
```
hegner, line 24: why not emplace_back?

python/templates/CollectionData.cc.jinja2
```diff
@@ -0,0 +1,211 @@
...
+    isRefColl ? nullptr : (void*)&m_data,
+    &m_refCollections, // only need to store the ObjectIDs of the referenced objects
+    &m_vecmem_info
+  };
+}
+
+void {{ class_type }}::prepareForWrite(bool isRefColl) {
+  for (auto& pointer : m_refCollections) { pointer->clear(); }
+
+  // If this is a reference collection use the relation storing mechanism to
```
hegner, line 93: reference collection...

hegner (changes requested):
Thanks. Nice work and cleanup done alongside.

There are two things that we should add to the tests 
(unless I overlooked it):
1) having a subset collection in a file with the original object missing.
2) check the subset collection for adding objects from two different collections.

python/templates/CollectionData.cc.jinja2
hegner: well. I think I can agree with you. thanks.

hegner:
> > 1. having a subset collection in a file with the original object missing.
> 
> This will currently result in a `nullptr` access and a pretty ugly crash, I think. How should we handle this, ideally? Resp. what is the latest point that we can get away with?

It depends a bit on whether we want to really only have subsets of collections, Or whether we dare to have references into multiple collections (better not!). If we are only talking about subsets, If we have real subsets, a consistency check on whether the previous collection is there would suffice. We can play with the reader and some metadata about collection dependencies there.

In general, I have my issues with the subset collections, when it comes to later dropping of collections... but that would be a different discussion.

tmadlener:
From a technical point of view "subset" collections can store elements from many different collections. The only requirement that is really there is that all objects that are put into a subset collection are already stored in another collection, as they need to have a valid ObjectID. (This is also the reason why I called them reference collections in the beginning). From there storage basically works equivalent to `OneToOneRelations`.

Looking at the handling of relations in general, the current implementation seems to break quite easily if collections where related objects live are are no longer present. One possibility for subset collections could be to have a way of storing a copy of the data of the original objects on demand. However, that would break some of the assumptions when a collection stored this way is read in again (most importantly the ObjectIDs of the stored elements would have changed).

tests/write_test.h
```diff
@@ -166,6 +171,31 @@ void write(podio::EventStore& store, WriterT& writer) {
...
+      if (p.id() % 2) {
+        mcpsRefs.push_back(p);
+      }
+    }
+    // ----------------- add the "even" counterparts from a different collection
+    for (auto p : moreMCs) {
+      if (p.id() % 2 == 0) {
+        mcpsRefs.push_back(p);
+      }
+    }
```
tmadlener, line 192: > 2. check the subset collection for adding objects from two different collections.

I added this test here that uses original objects from two different collections, and so also shows that the subset collections do not technically have to be subsets of one collection.

hegner: thanks

hegner:
There was a notion of marking ObjectIDs as invalid. we should come back to that.

Thanks for clarifying the subset part. In that case, we have to couple the data with some metadata to allow a check-on-write of completeness (are all required collections there?) Otherwise the reading will have to do it, which is much slower.

hegner (approved):

Issue #204, opened 2021-08-10
Proposal to make the Const classes default and explicitly mark the mutable ones as mutable


tmadlener:
This issue is mainly intended to gather feedback from users we are potentially not yet aware of. During the discussion at the [EDM4hep meeting on Aug 10](https://indico.cern.ch/event/1065461/) there was general agreement that the following proposal should be adopted. However, we also agreed that we would still like to collect more feedback as it constitutes a **breaking change**. Additionally, it should also help us with possible migration strategies.

**Please let us know what you think**

## TL;DR (reasoning below)
The proposal is to make the `Const` user-facing classes that are generated by podio the "default" classes and rename the current "default" classes to something that explicitly marks them as mutable. Either via
- putting these classes into a separate namespace (e.g. `mut` or `mod`)
- or by sticking a `Mutable` (or `Mut`) to the class name as pre- or post-fix

The main reason for this proposal is that it makes the default use case of simply reading data easier to write, and additionally making it very clear when something needs to be mutated. The details of the naming scheme are still part of the discussion and input is welcome.

## Current situation in podio
For the example datatype
```yaml
MCParticle:
  Members:
    - int pdg // PDG
```
podio generates the following user classes at the moment
```cpp
struct ConstMCParticle {
  const int& getPdg() const;
};

struct MCParticle {
  void setPdg(int);
  const int& getPdg() const;
  operator ConstMCParticle() const; // conversion to Const
};
```
The only difference between the two classes is that the `Const` version offers only getter functions, but no setter functions. The collection interface looks something like this (just showing enough to make the point)

```cpp
struct MCParticleConstCollectionIterator {
  ConstMCParticle operator*();
};

struct MCParticleCollectionIterator {
  MCParticle operator*();
};

struct MCParticleCollection {
  MCParticle operator[](unsigned);
  ConstMCParticle operator[](unsigned) const;
  
  MCParticleCollectionIterator begin();
  MCParticleConstCollectionIterator begin() const;
};
```
Again a duplication of the classes is necessary here to make the interface const-correct; Depending on whether the collection is `const` (in the c++ sense) access to the `Const` (in the podio sense) or the "normal" classes is granted.

### Problems / Issues with the current approach
- The presence of the `Const` classes is not documented (#198) **even though they should be the default use case for reading**
- The default use case should be the "easier one to use"
- There is no implicit conversion from the "default" class to the `Const` class. This is by design as these conversions should only make the access more strict.

[...]

PR #205, opened 2021-08-11
Make the default classes immutable and mark mutable classes explictly

tmadlener:
BEGINRELEASENOTES
- Make the default classes immutable and mark mutable classes explictly via their class name (e.g. `Hit` and `MutableHit`). See a brief discussion in #204 for more details on the reasons for this **breaking change**.
- After these changes collections return mutable objects via their `create` functionality, and will only give access to the default (immutable) objects when they are `const` (e.g. when they are read from file).
- In general these changes should make it easier for users to write interface that behave as expected, and also make it very obvious where objects are actually mutated already from looking at an interface definition.

ENDRELEASENOTES

This is a first shot to implement the proposal #204. As a first draft implementation it prepends `Mutable` to all class names that are mutable and removes the `Const` prefix from all immutable ones.

Built on top of #143!

- [x] Documentation
- [x] More detailed release notes
- [x] Check how much this breaks EDM4hep and other Key4hep dependencies
- [x] Also rename `ConstExtraCode` and `ExtraCode`?


tmadlener:
A more realistic diff for usage where also mutable classes are actually used from k4SimDelphes:

https://github.com/tmadlener/k4SimDelphes/commit/3d8f0f167a5082341d3584385eb8db1dac66d4db

PR #206, opened 2021-08-13
Switch to Catch2 v3

tmadlener:
BEGINRELEASENOTES
- Switch to Catch2 v3 test library and by default assume that it is available. Use the 'USE_EXTERNAL_CATCH2` cmake option to control whether podio should use an external installation or if it should fetch and build it internally instead.
- Remove `catch.hpp` header that was previously shipped, since it is no longer needed.

ENDRELEASENOTES


vvolkl (approved):

hegner:
Looks great. Thanks! 
Could you as well update the top level howto? As to list the new potential dependency.

PR #209, opened 2021-08-13
Remove Python 2 from docs

hegner:
BEGINRELEASENOTES
- Remove mention of Python 2 compatibility
ENDRELEASENOTES


tmadlener:
Should we enforce a minimum python version also in our cmake config? Currently we require that it is the same as the one that has been used for building root, but that doesn't strictly guarantee > 3.6:

https://github.com/AIDASoft/podio/blob/2d36032cfc0d5f4e42fa58268dadb0a0cf6f1269/CMakeLists.txt#L82

Additionally, we could us the `Python::Interpreter` target ([link](https://cmake.org/cmake/help/latest/module/FindPython.html#imported-targets)) in the generator macro instead of `python` to avoid running into problems with systems that still have a `python` that is `python2`.

PR #212, opened 2021-08-25
Add simple tools to look at podio edm files

jummorenope:
BEGINRELEASENOTES
- Added tools PodioReader and PodioEventOverview to get information on the contents on files from the command line
- Added readEvent and goToEvent function to IReader interface and implemented goToEvent on SIOReader
- Had to adapt collection templates to make the dumping to ascii files work
ENDRELEASENOTES

### DIFF
--- a/include/podio/CollectionBase.h
+++ b/include/podio/CollectionBase.h
@@ -8,6 +8,8 @@
 #include <utility>
 #include <vector>
 
+
+
 namespace podio {
   // forward declarations
   class ICollectionProvider;


tools/PodioEventOverview.cpp
```diff
@@ -0,0 +1,59 @@
...
+    std::cout<<"ERROR FileName not received"<<std::endl;
+    return 1;
+  }
+  std::string FileName{argv[1]};
+  
+  //Declaring the reader object and openning the file
+  auto reader = getReader(FileName);
+  reader->openFile(FileName);
+  int eventNumber=reader->getEntries();
+  int readEvent=1;
```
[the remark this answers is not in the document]
hegner: I see the feature has been added now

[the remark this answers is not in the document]
ljcaviedesb: Yes, it has

tools/PodioEventOverview.cpp
```diff
@@ -0,0 +1,87 @@
...
+  std::cout<<std::endl;
+  
+  //Getting Table containing the info about the collections
+  const auto collIDTable = reader->getCollectionIDTable();
+  const auto collNames = collIDTable->names();
+  //Iterating over all events to get size for each event
+  for(int i=startEvent; i<readEvent; i++){
+
+    std::cout<<"Event Number "<<i<<std::endl;
+    std::cout<<std::left<<std::setw(30)<<"Name"<<std::left<<std::setw(40)<<"Type"<<"Colection Size"<<std::endl;
```
tmadlener, line 69:
```suggestion
    std::cout<<std::left<<std::setw(30)<<"Name"<<std::left<<std::setw(40)<<"Type"<<"Collection Size"<<std::endl;
```

A minor typo that I just spotted during your presentation today :)

include/podio/CollectionBase.h
```diff
@@ -8,7 +8,8 @@
 #include <utility>
 #include <vector>
 
-
+#include "podio/ObjectID.h"
+#include "podio/ASCIIWriter.h"
```
tmadlener, line 12: I think these can be safely removed, right?

include/podio/tools.h
```diff
@@ -0,0 +1,15 @@
+#include <iostream>
+#include <iomanip>
```
tmadlener, line 2:
```suggestion
#include <memory>
```

That should be the only header from the STL that is necessary here, since the others are related to output and should instead be included where they are really needed (if not already the case)

tools/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+ADD_EXECUTABLE( PodioEventOverview PodioEventOverview.cpp )
+TARGET_LINK_LIBRARIES( PodioEventOverview podioRootIO podioSioIO podio )
+
+
```
tmadlener, line 4:
```suggestion
INSTALL(TARGETS PodioEventOverview DESTINATION ${CMAKE_INSTALL_BINDIR})
```
We also need to install the binary to `bin` to make it usable after it has been built.

tmadlener (changes requested):
I have just a few minor change requests that should be addressed before this can be merged.

include/podio/CollectionBase.h
jummorenope: Yes, they can

tmadlener:
Superseded by #323

PR #213, opened 2021-09-01
Add user data collections w/ basic types

gaede:
BEGINRELEASENOTES
- add possibility to store additional user data as collections of fundamental types in PODIO files
     -  uses `std::vector<basic_type>`
     - stored in simple branch in root (and simple block in SIO)
     - all fundamental types supported in PODIO (except bool) can be written
- example code:
```c++
  auto& usrInts = store.create<podio::UserDataCollection<uint64_t> >("userInts");
  auto& usrDoubles = store.create<podio::UserDataCollection<double> >("userDoubles");
  // ...

  // add some unsigned ints
  usrInts.resize( i + 1 ) ;
  int myInt = 0 ;
  for( auto& iu : usrInts ){
    iu = myInt++  ;
  }
  // and some user double values
  unsigned nd = 100 ;
  usrDoubles.resize( nd ) ;
  for(unsigned id=0 ; id<nd ; ++id){
    usrDoubles[id] = 42. ;
  }
```

- should replace https://github.com/key4hep/EDM4hep/pull/114 in a more efficient way

ENDRELEASENOTES

This should address most needs for storing user defined data in PODIO in an efficient way without any overhead.  In ROOT every user data collection is stored directly in a branch of the given type and can be accessed directly.
No relations to or from the user defined collections are foreseen. Users need to keep track if indices, e.g. by running collections parallel to existing PODIO (EDM4hep) collections.

This PR should work now as an example for possible treatment of additional user data. Open questions:

- [x] should the set of basic_types be restricted ?
- [x] if so, to which types ?
- [x] some code cleanup/improvement needed (see suggestions below)
- [x] ...

### DIFF
--- a/include/podio/UserDataCollection.h
+++ b/include/podio/UserDataCollection.h
@@ -23,10 +23,10 @@ namespace podio {
 
     const std::map< const std::type_index, const std::string >_typeMap =
     {
-      { std::type_index( typeid(int) ),   "podio::int"   },
-      { std::type_index( typeid(long) ),  "podio::long"  },
-      { std::type_index( typeid(float) ), "podio::float" },
-      { std::type_index( typeid(double) ),"podio::double"}
+      { std::type_index( typeid(int) ),   "podio::User_int"   },
+      { std::type_index( typeid(long) ),  "podio::User_long"  },
+      { std::type_index( typeid(float) ), "podio::User_float" },
+      { std::type_index( typeid(double) ),"podio::User_double"}
     } ;
 
   public:
@@ -154,16 +154,10 @@ namespace podio {
 
   /// some typedefs to make ROOT and the ROOTWriter happy
 
-  typedef UserDataCollection<int>    intCollection ;
-  typedef UserDataCollection<long>   longCollection ;
-  typedef UserDataCollection<float>  floatCollection ;
-  typedef UserDataCollection<double> doubleCollection ;
-
-  // are these needed or not ? the ROOT branch is created w/ "vector<xxxData>" where xxx is int, float, ...
-   // typedef int    intData ;
-   // typedef long   longData ;
-   // typedef float  floatData ;
-   // typedef double doubleData ;
+  typedef UserDataCollection<int>    User_intCollection ;
+  typedef UserDataCollection<long>   User_longCollection ;
+  typedef UserDataCollection<float>  User_floatCollection ;
+  typedef UserDataCollection<double> User_doubleCollection ;
 
 } // namespace
 
--- a/include/podio/UserDataCollection.h
+++ b/include/podio/UserDataCollection.h
@@ -49,6 +49,17 @@ namespace podio {
     constexpr bool isSupported() {
       return inTuple<T>(SupportedUserDataTypes{});
     }
+
+
+    /// some static type checks to ensure our assumptions on the fundamental types are correct
+    // note: eventually we might need some more generic tests and checks on type equivalence and
+    //       size_of supported types in PODIO as this also can affect the generated POD types...
+    static_assert( std::is_same_v<int16_t,short>);
+    static_assert( std::is_same_v<int32_t,int>);
+    static_assert( std::is_same_v<int64_t,long long>);
+    static_assert( std::is_same_v<uint16_t,unsigned short>);
+    static_assert( std::is_same_v<uint32_t,unsigned>);
+    static_assert( std::is_same_v<uint64_t,unsigned long long>);
   }
 
   /**
@@ -64,26 +75,26 @@ namespace podio {
   template <typename BasicType, typename = EnableIfSupportedUserType<BasicType> >
   constexpr const char*  userDataTypeName() ;
 
-  PODIO_ADD_USER_TYPE(int)
+  // PODIO_ADD_USER_TYPE(int)
   PODIO_ADD_USER_TYPE(long)
   PODIO_ADD_USER_TYPE(float)
   PODIO_ADD_USER_TYPE(double)
-  PODIO_ADD_USER_TYPE(unsigned)
-//  PODIO_ADD_USER_TYPE(unsigned int)
+  // PODIO_ADD_USER_TYPE(unsigned)
+  // PODIO_ADD_USER_TYPE(unsigned int)
   PODIO_ADD_USER_TYPE(unsigned long)
   PODIO_ADD_USER_TYPE(char)
-  PODIO_ADD_USER_TYPE(short)
-  PODIO_ADD_USER_TYPE(long long)
-  PODIO_ADD_USER_TYPE(unsigned long long)
-//  PODIO_ADD_USER_TYPE(int16_t)
-//  PODIO_ADD_USER_TYPE(int32_t)
-//  PODIO_ADD_USER_TYPE(int64_t)
+  // PODIO_ADD_USER_TYPE(short)
+  // PODIO_ADD_USER_TYPE(long long)
+  // PODIO_ADD_USER_TYPE(unsigned long long)
+  PODIO_ADD_USER_TYPE(int16_t)
+  PODIO_ADD_USER_TYPE(int32_t)
+  PODIO_ADD_USER_TYPE(int64_t)
   PODIO_ADD_USER_TYPE(uint16_t)
-//  PODIO_ADD_USER_TYPE(uint32_t)
-//  PODIO_ADD_USER_TYPE(uint64_t)
+  PODIO_ADD_USER_TYPE(uint32_t)
+  PODIO_ADD_USER_TYPE(uint64_t)
 
 // note: duplicate types on 'standard' hardware are commented out here
-
+//       but left in for a potential updata in the future...
 
   /** Collection of basic types for additional user data not defined in the EDM.
    *  The data is stored in an std::vector<basic_type>. Supported are all basic types supported in
@@ -197,6 +208,10 @@ namespace podio {
 
 
   };
+
+
+// don't make this macro public as it should only be used internally here...
+#undef PODIO_ADD_USER_TYPE
 
 } // namespace
 
--- a/include/podio/UserDataCollection.h
+++ b/include/podio/UserDataCollection.h
@@ -17,6 +17,9 @@ namespace podio {
   class  UserDataTypes{
   private:
     UserDataTypes() = default ;
+    UserDataTypes(const UserDataTypes& ) = delete;
+    UserDataTypes& operator=(const UserDataTypes& ) = delete;
+    ~UserDataTypes() = default;
 
     const std::map< const std::type_index, const std::string >_typeMap =
     {
@@ -31,6 +34,11 @@ namespace podio {
     std::string name( std::type_index idx ){
       auto it = _typeMap.find( idx ) ;
       return it != _typeMap.end() ? it->second  :  "UNKNOWN"  ;
+    }
+
+    // typename w/ 'podio::' replaced by 'podio_'
+    std::string sio_name( std::type_index idx ){
+      return std::string("podio_") + name( idx ).substr( 7 , 1024)  ;
     }
 
     static UserDataTypes& instance(){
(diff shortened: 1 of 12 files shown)


include/podio/UserDataCollection.h
```diff
@@ -0,0 +1,169 @@
...
+
+    static UserDataTypes& instance(){
+      static UserDataTypes me ;
+      return me ;
+    }
+  } ;
+
+
+  /// Templated base class for storing std::vectors of basic types as user data in PODIO
+  template <typename BasicType>
```
tmadlener, line 52: This template should be constrained to only accept those types that we actually support via the mapping above. Currently it is possible to do something like
```cpp
auto& coll = store.create<podio::UserDataCollection<unsigned>>("unsignedColl");
writer.registerForWrite("unsignedColl");
```
which compiles fine but crashes at some point at run-time. It should be possible to catch this at compile time already either via some `static_assert` or `enable_if` trickery.

include/podio/UserDataSIOBlock.h
```diff
@@ -0,0 +1,63 @@
...
+      podio::UserDataTypes::instance().name( typeid(BasicType) ) , this);
+  }
+
+  UserDataSIOBlock(const std::string& name) :
+    SIOBlock(name, sio::version::encode_version(0, 1)) {}
+
+
+  virtual void read(sio::read_device& device, sio::version_type /*version*/) override {
+    auto collBuffers = _col->getBuffers();
+    if (not _col->isSubsetCollection()) {
```
tmadlener, line 32: This check could be removed for the `UserDataCollection`, since it will always return true.

tests/read_test.h
```diff
@@ -323,6 +326,32 @@ void processEvent(podio::EventStore& store, int eventNum) {
   check_fixed_width_value(arbComps.fixedInteger64, std::int64_t{-1234567890123456789}, "int64");
   check_fixed_width_value(arbComps.fixedInteger32, std::int32_t{-1234567890}, "int32");
   check_fixed_width_value(arbComps.fixedUnsigned16, std::uint16_t{12345}, "uint16");
+
+
+
+  auto& usrInts = store.get<podio::UserDataCollection<int> >("userInts");
+
+#if 0 //access via vector
```
tmadlener, line 334: Are these left over from testing, or are they actually required?

include/podio/UserDataCollection.h
```diff
@@ -0,0 +1,169 @@
...
+
+  typedef UserDataCollection<int>    intCollection ;
+  typedef UserDataCollection<long>   longCollection ;
+  typedef UserDataCollection<float>  floatCollection ;
+  typedef UserDataCollection<double> doubleCollection ;
+
+  typedef int    intData ;
+  typedef long   longData ;
+  typedef float  floatData ;
+  typedef double doubleData ;
```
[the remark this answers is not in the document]
gaede: Thanks, Thomas. This seems to do the trick (at least on my centos7 test system) However I do not understand why and more importantly how this can even work, as the branch for the user collection is defined (by the ROOTReader) as `vector<doubleData>` and `doubleData` is a non existing type w/o these typedefs !???

include/podio/UserDataCollection.h
```diff
@@ -0,0 +1,169 @@
...
+    /// get collection ID
+    unsigned getID() const override final {
+      return m_collectionID ;
+    }
+
+    /// Get the collection buffers for this collection
+    podio::CollectionBuffers getBuffers() override final {
+      return { &_vecPtr,
+	&m_refCollections, // only need to store the ObjectIDs of the referenced objects
+	&m_vecmem_info } ;
```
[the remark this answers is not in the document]
gaede: This is what I had in initially - but it crashes in the SIOWriter as the SIOBlocks have:
```
  //---- read ref collections -----
  auto* refCols = collBuffers.references;
  for( auto* refC : *refCols ){
    unsigned size{0};
    device.data( size ) ;
    refC->resize(size) ;
    podio::handlePODDataSIO( device ,  &((*refC)[0]), size ) ;
  }
```
Should add a check for null pointer here  ...

include/podio/UserDataCollection.h
```diff
@@ -0,0 +1,169 @@
...
+#include <vector>
+#include <typeindex>
+
+
+namespace podio {
+
+
+  ///helper class providing names for basic types used in UserDataCollection<T>
+  class  UserDataTypes{
+  private:
```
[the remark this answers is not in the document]
gaede: Probably. As this is still WIP I wanted to see how far we can take this approach and wanted to have a potentially extendible map (w/ adding a `registerTypeName()` method or similar...

include/podio/UserDataCollection.h
```diff
@@ -0,0 +1,169 @@
...
+    UserDataTypes& operator=(const UserDataTypes& ) = delete;
+    ~UserDataTypes() = default;
+
+    const std::map< const std::type_index, const std::string >_typeMap =
+    {
+      { std::type_index( typeid(int) ),   "podio::int"   },
+      { std::type_index( typeid(long) ),  "podio::long"  },
+      { std::type_index( typeid(float) ), "podio::float" },
+      { std::type_index( typeid(double) ),"podio::double"}
+    } ;
```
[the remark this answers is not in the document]
gaede: Open for discussion - this is just an initial set that probably would have to be extended by at least some `short` and `unsigned`, maybe `fixed_size` types. This depends if we can make the mechanism extendible, i.e. such that users can add additional types if needed or not. For SIO this should work easily - with ROOT I am not so sure due to the discionary, etc...

include/podio/UserDataCollection.h
gaede: Fully agreed, if we decide to only provide a fixed set of types - if not, then of course it cannot be done and we leave failure to runtime...

include/podio/UserDataSIOBlock.h
gaede: Indeed.

tests/read_test.h
gaede: Sort of: testing - and as example how the underlying vector could be accessed, if needed. Can also remove eventually.
All this is still very much WIP ...

src/ROOTWriter.cc
```diff
@@ -55,7 +55,18 @@ void ROOTWriter::createBranches(const std::vector<StoreCollection>& collections)
...
+
+      std::string valTypeName = coll->getValueTypeName() ;
+
+      auto collClassName = "vector<" + valTypeName + "Data>";
+
+      auto it = valTypeName.find("podio::User_") ;
+      if(it == 0 ){
+	valTypeName = valTypeName.substr( 12 , 1024 ) ; // rm 'podio_user_' prefix
+	 collClassName = "vector<" + valTypeName +">";
+      }
```
tmadlener, line 67: Should we consider adding a `getDataTypeName` function (or similar) to the `podio::CollectionBase`? It would make some of the string juggling easier to do for general use, and give us a customization point where these user classes could be handled more nicely. As far as I can tell the major use cases for `getValueTypeName` currently are:
- Keys for the map inside the `SIOBlockFactory`
- Creating the `vector<XYZData>` for the root writer and reader

So maybe it could even be a `getDataVectorTypeName` alongside the `getValueTypeName` (?)

vvolkl:
Hi Frank,

I think the access to the underlying vector here is very nice! In my opinion the available data types are fine; they can always be expanded if necessary. Thomas' point for a compile time check is still a good one though.

python/templates/Collection.h.jinja2
```diff
@@ -61,8 +61,12 @@ public:
   /// number of elements in the collection
   size_t size() const override final;
 
+  /// fully qualified type name
+  std::string getTypeName() const override { return std::string("{{ (class | string ).strip(':')+"Collection" }}"); }
   /// fully qualified type name of elements - with namespace
   std::string getValueTypeName() const override { return std::string("{{ (class | string ).strip(':') }}"); }
+  /// fully qualified type name of stored POD elements - with namespace
+  std::string getDataTypeName() const override { return std::string("{{ (class | string ).strip(':')+"Data" }}"); }
```
tmadlener, line 69: It looks like this will always be used like this
```cpp
const auto bufferClassName = "std::vector<" + coll->getDataTypeName() + ">";
```
Should we simply make this return the full thing (and potentially give it a name that indicates that)? Or do we foresee any potential use cases where the `Data` name alone would be useful?

tmadlener (commented):
As already discussed during the meeting yesterday, I think this is a very nice addition and I like the implementation. I have left one comment that is not only about implementation details regarding the return value of `getDataTypeName`. I am not sure what others think there.

From a technical point of view I would then simply try to push as many failures to compile time rather than runtime, since we have a closed set of types that we want to support with this (the [`BUILTIN_TYPES`](https://github.com/AIDASoft/podio/blob/cd9b3561a088b84f4fadc6e678cc24cbf58f45a5/python/generator_utils.py#L43-L46) sans `std::string` and the [`ALLOWED_FIXED_WIDTH_TYPES`](https://github.com/AIDASoft/podio/blob/cd9b3561a088b84f4fadc6e678cc24cbf58f45a5/python/generator_utils.py#L54-L55)).

python/templates/Collection.h.jinja2
gaede: In a sense the fact that a vector<> is used here is an implementation detail of the ROOTReader/Writer and other future implementations (HDF5 ?) might make a different use of the underlying POD data type. 
This is why I thought it would be more generic this way...

include/podio/UserDataCollection.h
```diff
@@ -0,0 +1,135 @@
...
+namespace podio {
+
+  template <typename BasicType>
+  std::string userDataTypeName() { throw std::runtime_error( std::string(" unsupported type for UserDataCollection: ")
+							     + typeid(BasicType).name() ) ; }
+
+  template <>  std::string userDataTypeName<int>()   {return "int" ; }
+  template <>  std::string userDataTypeName<float>() {return "float" ; }
+  template <>  std::string userDataTypeName<long>()  {return "long" ; }
+  template <>  std::string userDataTypeName<double>(){return "double" ; }
```
tmadlener, line 22: These could be made into a
```cpp
template<typename BasicType, 
         typename = EnableIfSupportedUserType<BasicType>> 
constexpr const char* userDataTypeName();

template<> constexpr const char* userDataTypeName<int>() { return "int"; }
// ... etc.
```
without any changes necessary to the rest of the code. Removing the implementation for the default case would also lead to a linking error if anyone tries to call this with an unsupported type. To ease the adding of additional names one could introduce a small macro, but still with that we would have to spell out the list of supported times a second time (after the initial `std::tuple`).
```cpp
#define ADD_USER_TYPE(type) template<> constexpr const char* userDataTypeName<type>( return #type; }
```

tmadlener (commented):
To implement a compile-time constraint on the usable types it would be possible to do something along the lines of the following: Define a `std::tuple` containing all supported types.
```cpp
using SupportedUserDataTypes = std::tuple<
    int,
    float,
    long,
    double
    >;
```
Around that it is then rather straight forward to implement the necessary `enable_if` machinery to make compilation fail in case of unsupported types
```cpp
namespace detail {
  /**
   * Helper function to check whether a type T is in a std::tuple<Ts...>
   */
  template<typename T, typename ...Ts>
  constexpr bool inTuple(std::tuple<Ts...>) {
    return (std::is_same_v<T, Ts> || ...);
  }

  /**
   * Compile time helper function to check whether the given type is in the list
   * of supported types
   */
  template<typename T>
  constexpr bool isSupported() {
    return inTuple<T>(SupportedUserDataTypes{});
  }
}

/**
 * Alias template to be used to enable template specializations only for the types listed in the
 * SupportedUserDataTypes list
 */
template<typename T>
using EnableIfSupportedUserType = std::enable_if_t<detail::isSupported<T>()>;
```

Usage:
```cpp
template<typename T
         typename = EnableIfSupportedUserType<T>>
// class or function that should only be available if T is in SupportedUserDataTypes
```

tmadlener (commented):
This looks good to me, I just have a few comments that are all more or less about the same thing: Trying to cut down the list of supported types to the minimum while keeping all the currently present functionality.

However, they can also be addressed later IMHO.

gaede:
Thanks @tmadlener - added your suggestions. As discussed privately, we might need a more generic treatment of ensuring consistency and size of fundamental types in PODIO eventually ...

include/podio/UserDataCollection.h
```diff
@@ -0,0 +1,229 @@
...
+
+    // some templates to ensure valid and supported user types
+    // as suggested by T.Madlener, DESY
+
+    /** tuple of basic types supported in user vector
+     */
+    using SupportedUserDataTypes = std::tuple<
+      int, long, float, double,
+      unsigned int, unsigned, unsigned long,
+      char, short, long long,
```
hegner, line 30: can you quickly clean up the non-fixed width types?

gaede: I actually intentionally left them in, as this is what we currently support in PODIO. I think we should address the issue of 'set of supported types' in PODIO consistently in a dedicate PR (with sufficient documentation). What do you think?

hegner: I think for a new feature we should already start restricted. We can open up easily later in case. The other way round is more complicated.

PR #214, opened 2021-09-01
Make the generation macro use the python found by CMake

tmadlener:
BEGINRELEASENOTES
- Make the CMake datamodel generation macro use the python interpreter that is also found by CMake to avoid accidentally picking up an unsuitable system provided version that might be on `PATH`.

ENDRELEASENOTES


andresailer:
I think you need to add
`find_dependency(Python Interpreter)` to `podioConfig.cmake.in` or something like that, because, for example in EDM4hep we are not looking for Python and then Python_EXECUTABLE would be empty?
https://github.com/key4hep/EDM4hep/search?q=find_package&type=

tmadlener:
You are right. I didn't think about that.

andresailer (approved):

hegner:
Do we merge before or after the release?

tmadlener:
I think this could be included in the release as it should only make things more stable than they are currently.

PR #217, opened 2021-09-10
Fix a few smaller generation issues (once more)

### LINKED ISSUE #203: Destructors for generated Obj classes should be defaulted if there are no relations
Currently the `~Obj` destructor is generated unconditionally and is defined as (sketched)
```cpp
~Obj() {
  // deal with OneToManyRelations and VectorMembers
  // deal with OneToManyRelations
}
```
Hence, if there are no relations defined for a datatype it could simply be declared as
```cpp
virtual ~Obj() = default;
```
instead of implementing an empty destructor.

### LINKED ISSUE #216: OneToOneRelations can only be used if they are in the same namespace
As of cd9b3561a088b84f4fadc6e678cc24cbf58f45a5 the following datamodel definition does not compile
```yaml
datatypes:
  example::NamespacedType:
    # necessary definition

  DataTypeWithRelation:
    # necessary definition
    OneToOneRelation:
      - example::NamespacedType relation // a relation in a different namespace
```

The corresponding compiler error:
```console
../tests/src/DataTypeWithRelationCollectionData.cc: In member function ‘void DataTypeWithRelationCollectionData::setReferences(const podio::ICollectionProvider*, bool)’:
../tests/src/DataTypeWithRelationCollectionData.cc:107:36: error: expected type-specifier before ‘ConstNamespacedType’
  107 |       entries[i]->m_relation = new ConstNamespacedType((*tmp_coll)[id.index]);
      |                                    ^~~~~~~~~~~~~~~~~~~
../tests/src/DataTypeWithRelationCollectionData.cc:106:42: warning: unused variable ‘tmp_coll’ [-Wunused-variable]
  106 |       example::NamespacedTypeCollection* tmp_coll = static_cast<example::NamespacedTypeCollection*>(coll);
      |                                          ^~~~~~~~

```

tmadlener:
BEGINRELEASENOTES
- Make the `Obj` destructors `= default` where possible, i.e. if a datatype has no relations to handle
- Make the assignment operators of the user facing classes use the "copy-and-swap" idiom
- Fix the problem where `OneToOneRelations` needed to be from the same namespace as the datatype they are used in (#216)

ENDRELEASENOTES

Fixes #216, fixes #203

### DIFF
--- a/python/podio_class_generator.py
+++ b/python/podio_class_generator.py
@@ -250,6 +250,8 @@ class ClassGenerator(object):
     datatype['forward_declarations_obj'] = fwd_declarations
     datatype['includes_obj'] = self._sort_includes(includes)
     datatype['includes_cc_obj'] = self._sort_includes(includes_cc)
+    needs_destructor = datatype['VectorMembers'] or datatype['OneToManyRelations'] or datatype['OneToOneRelations']
+    datatype['obj_needs_destructor'] = needs_destructor
 
   def _preprocess_for_class(self, datatype):
     """Do the preprocessing that is necessary for the classes and Const classes"""
--- a/python/templates/Obj.cc.jinja2
+++ b/python/templates/Obj.cc.jinja2
@@ -43,6 +43,7 @@
 {% endfor %}
 }
 
+{% if obj_needs_destructor -%}
 {{ obj_type }}::~{{ obj_type }}() {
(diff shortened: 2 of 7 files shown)


hegner (approved):
I really like the move to swap. if the tests run fine, I will merge :-)

PR #221, opened 2021-09-28
[WIP] Improve user facing API: make Object ctor from Obj* private

vvolkl:
This PR makes [this and similar constructors](https://edm4hep.web.cern.ch/classedm4hep_1_1_m_c_particle.html#a4cc4912e06c4ee5df29edaa97a18317d) 
```


edm4hep::MCParticle::MCParticle  (  MCParticleObj *  obj  ) 
```
private. As far as I can see there is no way for users to access a `MCParticleObj*` and they are not intended to do so, so this should not be part of the public API.

The downside implementationwise is that the constructors templating macro does not cover private, so I had to move it out of the macro.

BEGINRELEASENOTES
- Improve user facing API: make Object ctor from Obj* private

ENDRELEASENOTES

### DIFF
--- a/python/templates/ConstObject.h.jinja2
+++ b/python/templates/ConstObject.h.jinja2
@@ -35,6 +35,11 @@ public:
 {{ macros.common_object_funcs(class.bare_type, prefix='Const') }}
 
 private:
+
+  /// constructor from existing {{ class.bare_type }}Obj
+  Const{{  class.bare_type }}({{ class.bare_type }}Obj* obj);
+
+
   {{ class.bare_type }}Obj* m_obj;
 };
 
--- a/python/templates/Object.h.jinja2
+++ b/python/templates/Object.h.jinja2
@@ -43,6 +43,9 @@ public:
 {{ macros.common_object_funcs(class.bare_type) }}
 
 private:
+  /// constructor from existing {{ class.bare_type }}Obj
+  {{  class.bare_type }}({{ class.bare_type }}Obj* obj);
+
   {{ class.bare_type }}Obj* m_obj;
 };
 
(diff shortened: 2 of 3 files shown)


tmadlener:
I agree that the constructor from an `Obj*` should be private. However, it is not entirely trivial, because it is also used for `OneToOneRelations` that are missing or not (yet) set, i.e. the constructor from a `nullptr` below.

https://github.com/AIDASoft/podio/blob/ed207e8be419c698124a5570d90daee6c31161e5/python/templates/macros/implementations.jinja2#L79-L84

One way out of this would be to declare all of the related classes as `friend`, but that can become quite a list, and I am not sure we want to go into that direction.

tmadlener:
I agree that the constructor from an `Obj*` should be private. However, it is not entirely trivial, because it is also used for `OneToOneRelations` that are missing or not (yet) set, i.e. the constructor from a `nullptr` below.

https://github.com/AIDASoft/podio/blob/ed207e8be419c698124a5570d90daee6c31161e5/python/templates/macros/implementations.jinja2#L79-L84

One way out of this would be to declare all of the related classes as `friend`, but that can become quite a list, and I am not sure we want to go into that direction.

tmadlener:
I just realized this constructor is also used in `setReferences`, e.g. here:

https://github.com/AIDASoft/podio/blob/master/python/templates/macros/collections.jinja2#L63-L69

tmadlener:
I just realized this constructor is also used in `setReferences`, e.g. here:

https://github.com/AIDASoft/podio/blob/master/python/templates/macros/collections.jinja2#L63-L69

vvolkl:
Ok, first step is then probably to update to doc string to clearly state that this is an internal method, and then fix this in the long run.

tmadlener:
Superseded by #514

PR #230, opened 2021-10-11
Add test to read a legacy (v00-13) root file

vvolkl:
I created a root file using the `write` test exe of podio v00-13 and am reading this from a server. This means that the testsuite now needs network access, but nevertheless an important test to have.

BEGINRELEASENOTES
- Add test to read a legacy (v00-13) root file
ENDRELEASENOTES

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -25,7 +25,7 @@ function(CREATE_PODIO_TEST sourcefile additional_libs)
     )
 endfunction()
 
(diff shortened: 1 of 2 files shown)


tmadlener:
One problem I just realized with this is that the read tests also evolve and now probably expect different (i.e. additional) content. E.g. from #197 (subset collections were added), or from #213 (user data).

#232 should fix the metadata issues, but the content problems are still there with that.

PR #232, opened 2021-10-11
Make it possible read "old" podio files

tmadlener:
BEGINRELEASENOTES
- Make it possible to read "old" podio data files that have been written with podio < 0.13.1 (i.e. before #197) was merged.
  - Reconstruct the `"CollectionInfoType"` branch that as introduced there via other means and simply assume that all collections are proper collections (since subset collections didn't exist prior).

ENDRELEASENOTES

- [ ] Check if something similar can be done for sio


src/ROOTReader.cc
```diff
@@ -144,12 +144,20 @@ namespace podio {
...
+    // Check if the CollectionTypeInfo branch is there and assume that the file
+    // has been written with with podio pre #197 (<0.13.1) if that is not the case
+    if (auto* collInfoBranch = root_utils::getBranch(metadatatree, "CollectionTypeInfo")) {
+      auto collectionInfo = new std::vector<root_utils::CollectionInfoT>;
+      collInfoBranch->SetAddress(&collectionInfo);
+      metadatatree->GetEntry(0);
+      createCollectionBranches(*collectionInfo);
+      delete collectionInfo;
+    } else {
+      std::cout << "Reading old-style file" << std::endl;
```
hegner, line 156: if you integrate that into any framework of decent size you will have no idea where this message comes from. maybe adding "PODIO" and the name of the file will help

tmadlener: Ah yes, of course. Left over from testing.

PR #234, opened 2021-10-12
Make sure to include <cstdint> when using fixed width ints

tmadlener:
BEGINRELEASENOTES
- Make sure that `#include <cstdint>` is present when using fixed with integers in datatypes

ENDRELEASENOTES


hegner:
@tmadlener - feel free to merge the corresponding patch in EDM4hep

PR #235, opened 2021-10-13
Test and fix for persisting a collection without its associated collections

fdplacido:
BEGINRELEASENOTES
- Fix crashes that happen when reading collections that have related objects in collections that have not been persisted.
- Fix similar crashes for subset collections where the original collection has not been persisted.
  - The expected behavior in both cases is that podio does not crash when reading such collections, but only once the user tries to actually access such a missing object. Each object has an `isAvailable` function to guard against such crashes if need be.
- Add a test that makes sure that the expected behavior is the one that is observed.
- Fix a somewhat related bug in `setReferences` which was mistakenly a no-op for collections of a type without relations. Since this is the mechanism we use for restoring subset collections it obviously has to be present for all types.

ENDRELEASENOTES

Original problem:
- Create some collection (cluster), then add an associated collection to it (addHit). Finally persist only the original collection but not its associated collections.
- When reading these collections this will fail (segfault)
- Expected behaviour: to be able to read the clusters without a crash, marking the associated collections as invalid.

I created a test that writes and reads some cluster collection, which will fails with a segfault.
Only the clusters are registered for write: the hits are not written. If hits are registered for write the test works without issue.
I pushed a fix which checks the result of `collectionProvider->get()`, and if this fails will just continue to the next referenced collection. 

As an example, this would generated something like this:

```cpp
void ExampleClusterCollectionData::setReferences(const podio::ICollectionProvider* collectionProvider, bool isSubsetColl) {
// ...
  // Normal collections have to resolve all relations
  for (unsigned int i = 0, size = m_refCollections[0]->size(); i != size; ++i) {
    const auto id = (*m_refCollections[0])[i];
    if (id.index != podio::ObjectID::invalid) {
      podio::CollectionBase* coll = nullptr;
      auto collProvResult = collectionProvider->get(id.collectionID, coll);
      if (collProvResult == false) { continue; }
      ExampleHitCollection* tmp_coll = static_cast<ExampleHitCollection*>(coll);
      const auto tmp = (*tmp_coll)[id.index];
      m_rel_Hits->emplace_back(tmp);
    } else {
      m_rel_Hits->emplace_back(nullptr);
    }
  }
// ...
}
```

### DIFF
--- a/python/templates/Collection.cc.jinja2
+++ b/python/templates/Collection.cc.jinja2
@@ -119,7 +119,7 @@ void {{ collection_type }}::push_back({{ class.bare_type }} object) {
   } else {
     const auto obj = object.m_obj;
     if (obj->id.index < 0) {
-      throw std::invalid_argument("Object needs to be tracked by another collection in order for it to be storable in a subset collection");
+      throw std::invalid_argument("Objects can only be stored in a subset collection if they are already elements of a collection");
     }
     m_storage.entries.push_back(obj);
     // No need to handle any relations here, since this is already done by the
--- a/python/templates/Collection.cc.jinja2
+++ b/python/templates/Collection.cc.jinja2
@@ -94,9 +94,7 @@ void {{ collection_type }}::prepareAfterRead() {
 }
 
 bool {{ collection_type }}::setReferences(const podio::ICollectionProvider* collectionProvider) {
-  m_storage.setReferences(collectionProvider, m_isSubsetColl);
-
-  return true; //TODO: check success
+  return m_storage.setReferences(collectionProvider, m_isSubsetColl);
 }
 
 void {{ collection_type }}::push_back({{ class.bare_type }} object) {
--- a/python/templates/CollectionData.cc.jinja2
+++ b/python/templates/CollectionData.cc.jinja2
@@ -158,13 +158,13 @@ void {{ class_type }}::createRelations({{ class.bare_type }}Obj* obj) {
 }
 {% endif %}
 
-void {{ class_type }}::setReferences(const podio::ICollectionProvider* collectionProvider, bool isSubsetColl) {
+bool {{ class_type }}::setReferences(const podio::ICollectionProvider* collectionProvider, bool isSubsetColl) {
   if (isSubsetColl) {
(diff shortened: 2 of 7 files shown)


tmadlener (commented):
I had the suspicion that there was a problem with this.

Thanks for confirming that there indeed is one and for fixing it.

andresailer:
What happens now if there are two Hit Collections and only one of them is written, but a cluster in principle has links to hits in both collections?

tmadlener:
> What happens now if there are two Hit Collections and only one of them is written, but a cluster in principle has links to hits in both collections?

Relations are resolved one by one, so that in that case the cluster will still have all hits that have been present before writing. However, only those that have actually been written will also be available. I.e. in principle you would have to query each hit, whether it is available or not, before trying to access it.

fdplacido:
I created an issue to continue and track the inconsistent use of collections and subset collections here: https://github.com/AIDASoft/podio/issues/244

So this PR can be merged.

tmadlener:
Fixed the usage of `Mutable` classes now that #205 has been merged, and rebased everything onto the latest master.

Additionally, subset collections now behave the same as normal relations: I.e. if the referred to objects cannot be found they are simply not available, but the subset collection will still have the same size as before.

There was another problem that I had to fix, which seemed to have slipped through our tests until now: the `setReferences` functionality did only do any actual work for collections of types with at least one relation. The problem is that `setReferences` is also responsible for recreating the subset collections when reading. Hence, it should always do something.

tmadlener:
Fixed the usage of `Mutable` classes now that #205 has been merged, and rebased everything onto the latest master.

Additionally, subset collections now behave the same as normal relations: I.e. if the referred to objects cannot be found they are simply not available, but the subset collection will still have the same size as before.

There was another problem that I had to fix, which seemed to have slipped through our tests until now: the `setReferences` functionality did only do any actual work for collections of types with at least one relation. The problem is that `setReferences` is also responsible for recreating the subset collections when reading. Hence, it should always do something.

tmadlener:
@hegner also this would be nice to have before the newer developments start.

tmadlener (approved):

hegner:
Thanks @fdplacido and @tmadlener !

PR #238, opened 2021-10-20
Add build version and utilities to work with it

tmadlener:
BEGINRELEASENOTES
- Extend the `podioVersion.h` header that is configured by cmake to hold some version utilities.
  - `podio::version::Version` class holding three `uint16_t`s for major, minor and patch version, plus `constexpr` comparison operators.
  - static const(expr) `podio::version::build_version` that holds the current (i.e. last tag) version of podio
  - Add preprocessor macros with similar functionality
    - `PODIO_VERSION` takes a major, minor and a patch version number and encodes it into a 64 bit version constant. 
    - `PODIO_[MAJOR|MINOR|PATCH]_VERSION` macros can extracts these values again from a 64 bit encoded version. 
    - `PODIO_BUILD_VERSION` holds the 64 bit encoded current (i.e. last tag) version of podio
- Reorder the read tests slightly and make some sections version dependent
- Add legacy file read test from #230 

ENDRELEASENOTES

This is a first shot at this. Some of the open questions that might warrant discussion:
- [x] Do we need 3 16 bit unsigned ints or should we try to have all three parts in one 32 bit integer?
- [x] ~Double overload of the `compatible` function: Should we consider a different name for the one that checks if a version is compatible with the current `build_version`?~ Removed for now as meaning is not well defined (yet).
- [x] Do we need pre-processor capabilities for versions as well, or are we content with compile time functionality?
- [x] Is configuring the `podioVersion.h` header into the `include/podio` source directory what we should be doing?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,6 @@ spack*
 
 # Generated files
 *podio_generated_files.cmake
+
+# Populated by cmake before build
+/include/podio/podioVersion.h
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,7 +36,7 @@ endif()
 # ``-DCMAKE_CXX_STANDARD=<standard>`` when invoking CMake
 set(CMAKE_CXX_STANDARD 17 CACHE STRING "")
 
-if(NOT CMAKE_CXX_STANDARD MATCHES "17")
+if(NOT CMAKE_CXX_STANDARD MATCHES "17|20")
   message(FATAL_ERROR "Unsupported C++ standard: ${CMAKE_CXX_STANDARD}")
 endif()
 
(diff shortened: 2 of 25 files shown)


tmadlener:
Looks like getting the file via http is not working at the moment(?) on mac:
https://github.com/AIDASoft/podio/pull/238/checks?check_run_id=3948979989#step:4:512

tmadlener:
Switched to downloading the legacy input file up front, instead of going through roots http support. That makes the tests work in the macOS workflows as well.

tmadlener:
As discussed, I removed the `compatible` functionality, so that the PR now only holds the basic building blocks for versioning.

hegner (approved):

PR #240, opened 2021-10-27
Add PodioEventOverview tool to get a quick overview of file contents

tmadlener:
BEGINRELEASENOTES
- Add `PodioEventOverview` executable that allows to get a quick overview over the contents of an event.
- Make `goToEvent` and `readEvent` part of the `podio::IReader` interface as necessary prerequisites for the tool.
- Add a `createReader` utility function that creates the corresponding reader from the file ending of a passed file.

ENDRELEASENOTES

builds on and supersedes #212

- [ ] Should I squash some of the earlier commits before merging?

PR #242, opened 2021-10-27
[WIP] Add PodioDumpEvent tool

tmadlener:
BEGINRELEASENOTES
- Add a first version of a python `PodioDumpEvent` tool that prints all the contents of a given event in detail.
- Make it possible to use the python `EventStore` to open sio files.
  - Split the python tests into a base case and dedicated specializations for root and sio input files for easier coverage of both cases without having to duplicate code.
  - Split the python event store functionality off from the `podioRootIO` library into a dedicated separate shared library.

ENDRELEASENOTES

- Builds on top of #240. Marking this as WIP until that is merged and the diff of this one becomes a bit less noisy


andresailer:
-> `podioFrameInspect` ?

tmadlener:
Superseded by #323

PR #247, opened 2021-11-15
[WIP] Add ROOTNTuple{Reader|Writer} I/O Backends

vvolkl:
BEGINRELEASENOTES
-  Add RNTuple{Reader|Writer} I/O Backends

ENDRELEASENOTES

This adds an (experimental) I/O backend for RNTuple. ROOT should present users with the following warning when using the writer:

```
Warning in <[ROOT.NTuple] Warning /cvmfs/tmp/spack-stage/gitlab-runner/spack-stage-root-6.24.00-cbuvics7rcfotr66dayarc76fp77s42e/spack-src/tree/ntuple/v7/src/RPageStorageFile.cxx:65 in ROOT::Experimental::Detail::RPageSinkFile::RPageSinkFile(std::string_view, TFile&, const ROOT::Experimental::RNTupleWriteOptions&)>: The RNTuple file format will change. Do not store real data with this version of RNTuple!
```
so it should be clear that this is not to be used in production.

I chose `ROOTNTuple{Reader|Writer}` as class names as the CMake configuration already links files prefixed with `ROOT` differently - `RNTupleWriter` is shorter, but already exists  in ROOT itself (albeit in the `ROOT::Experimental` namespace).

The overall structure is very similar to the ROOTWriter. The code that allows to write data the "SetBranchAddress" way, via void pointers and a class name string, has  a few fairly opaque lines that have been discussed with the rntuple authors and may still be replaced with a simpler interface.  

- [x] RNTuple does not want to write `int16_t` members - follow this up with the  rntuple authors
- [x] Add the ROOTNTupleWriter
- [ ] write Metadata, references and vectormembers
- [ ] Comments and Docs


tmadlener:
Looks like the dev3/dev4 CI builds are broken because of some warnings in the build: https://github.com/AIDASoft/podio/runs/4217295201?check_suite_focus=true#step:4:219

If newer root versions are able to handle `int16_t` we should see these workflows succeed once the warnings are fixed.

src/ROOTNTupleWriter.cc
```diff
@@ -0,0 +1,143 @@
...
+void ROOTNTupleWriter::createModels(const std::vector<StoreCollection>& collections) {
+  for (auto& [name, coll] : collections) {
+    const auto collBuffers = coll->getBuffers();
+    if (collBuffers.data) {
+      // note: these interfaces may still be simplified on the rntuple side
+      auto collClassName = "vector<" + coll->getDataTypeName() +">";
+      std::cout << collClassName << std::endl;
+      auto field = rnt::Detail::RFieldBase::Create(name,  collClassName).Unwrap();
+      // what I cast buffers.data to  does not seem to matter, but it seems I have to cast it to something?
+      void* vv =  (void*) *static_cast<std::vector<float>**>(collBuffers.data);
```
vvolkl, line 85: @jblomer this is the cast I am struggling to understand.

The stack trace when this is not casted:
```
1: vector<ExampleWithFixedWidthIntegersData>
41: Warning in <[ROOT.NTuple] Warning /tmp/ubuntu/spack-stage/spack-stage-root-6.26.02-4uthei6jyfh7oowt6ogsl6pakm57qlq4/spack-src/tree/ntuple/v7/src/RPageStorageFile.cxx:51 in ROOT::Experimental::Detail::RPageSinkFile::RPageSinkFile(std::string_view, const ROOT::Experimental::RNTupleWriteOptions&)>: The RNTuple file format will change. Do not store real data with this version of RNTuple!
41: Fatal: (typedValue->size() % fItemSize) == 0 violated at line 1021 of `/tmp/ubuntu/spack-stage/spack-stage-root-6.26.02-4uthei6jyfh7oowt6ogsl6pakm57qlq4/spack-src/tree/ntuple/v7/src/RField.cxx'
41: aborting
41: #0  0x00007f9370c3cc3a in wait4 () from /lib/x86_64-linux-gnu/libc.so.6
41: #1  0x00007f9370babf67 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
41: #2  0x00007f937139fc9a in TUnixSystem::Exec (shellcmd=<optimized out>, this=0x156d800) at /tmp/ubuntu/spack-stage/spack-stage-root-6.26.02-4uthei6jyfh7oowt6ogsl6pakm57qlq4/spack-src/core/unix/src/TUnixSystem.cxx:2108
41: #3  TUnixSystem::StackTrace (this=0x156d800) at /tmp/ubuntu/spack-stage/spack-stage-root-6.26.02-4uthei6jyfh7oowt6ogsl6pakm57qlq4/spack-src/core/unix/src/TUnixSystem.cxx:2399
41: #4  0x00007f937127d0f2 in DefaultErrorHandler (level=<optimized out>, abort_bool=<optimized out>, location=<optimized out>, msg=0x67ba800 "(typedValue->size() % fItemSize) == 0 violated at line 1021 of `/tmp/ubuntu/spack-stage/spack-stage-root-6.26.02-4uthei6jyfh7oowt6ogsl6pakm57qlq4/spack-src/tree/ntuple/v7/src/RField.cxx'") at /tmp/ubuntu/spack-stage/spack-stage-root-6.26.02-4uthei6jyfh7oowt6ogsl6pakm57qlq4/spack-src/core/base/src/TErrorDefaultHandler.cxx:174
41: #5  0x00007f937132f878 in ErrorHandler(Int_t, const char *, const char *, typedef __va_list_tag __va_list_tag *) (level=6000, location=0x7f9371c2df4c "", fmt=<optimized out>, ap=<optimized out>) at /tmp/ubuntu/spack-stage/spack-stage-root-6.26.02-4uthei6jyfh7oowt6ogsl6pakm57qlq4/spack-src/core/foundation/src/TError.cxx:152
41: #6  0x00007f9371330129 in Fatal (location=location
41: entry=0x7f9371c2df4c "", fmt=<optimized out>) at /tmp/ubuntu/spack-stage/spack-stage-root-6.26.02-4uthei6jyfh7oowt6ogsl6pakm57qlq4/spack-src/core/foundation/src/TError.cxx:249
41: #7  0x00007f9371bce68b in ROOT::Experimental::RVectorField::AppendImpl (this=0x2447860, value=...) at /tmp/ubuntu/spack-stage/spack-stage-root-6.26.02-4uthei6jyfh7oowt6ogsl6pakm57qlq4/spack-src/tree/ntuple/v7/src/RField.cxx:1021
41: #8  0x00007f93720b7e39 in ROOT::Experimental::Detail::RFieldBase::Append (this=0x7391, value=...) at /cvmfs/sw.hsf.org/spackages5/root/6.26.02/x86_64-ubuntu20.04-gcc9.4.0-opt/4uthe/include/ROOT/RField.hxx:203
41: #9  ROOT::Experimental::RNTupleWriter::Fill (this=0x24ae5c0, entry=...) at /cvmfs/sw.hsf.org/spackages5/root/6.26.02/x86_64-ubuntu20.04-gcc9.4.0-opt/4uthe/include/ROOT/RNTuple.hxx:397
41: #10 ROOT::Experimental::RNTupleWriter::Fill (this=0x24ae5c0) at /cvmfs/sw.hsf.org/spackages5/root/6.26.02/x86_64-ubuntu20.04-gcc9.4.0-opt/4uthe/include/ROOT/RNTuple.hxx:392
41: #11 podio::ROOTNTupleWriter::writeEvent (this=<optimized out>) at /home/vavolkl/r/podio/src/ROOTNTupleWriter.cc:73
41: #12 0x0000000000409555 in write<podio::ROOTNTupleWriter> (store=..., writer=...) at /home/vavolkl/r/podio/tests/write_test.h:351
41: #13 0x0000000000405bbd in main () at /home/vavolkl/r/podio/tests/write_rntuple.cpp:24
```

tmadlener: This is how the `collBuffers.data` is populated:
https://github.com/AIDASoft/podio/blob/9d078f82d366e9363cbf2e941e3647badbcba690/python/templates/CollectionData.cc.jinja2#L76-L80

The `m_data` member in this case is a `std::unique_ptr<std::vector<XYZData>>` where `XYZ` is a datatype defined in the yaml file and the `Data` class is the corresponding POD that is generated for it.

For reference:
https://github.com/AIDASoft/podio/blob/9d078f82d366e9363cbf2e941e3647badbcba690/python/templates/CollectionData.h.jinja2#L92
https://github.com/AIDASoft/podio/blob/9d078f82d366e9363cbf2e941e3647badbcba690/python/templates/CollectionData.h.jinja2#L24

jblomer:
> @jblomer Here is the test datamodel: https://github.com/AIDASoft/podio/blob/master/tests/datalayout.yaml
> 
> A dump of the file produced by the [write test](https://github.com/AIDASoft/podio/blob/master/tests/write_test.h) is attached [write_test.log](https://github.com/AIDASoft/podio/files/8977327/write_test.log)

Many thanks!  That looks good to me!  At some point we may support intra-event references natively in RNTuple.  At that point, we can discuss again if/how podio object links should be represented by native links.

src/ROOTNTupleWriter.cc
jblomer: I think this may be a bug of an older RNTuple version.  If I'm not mistaken, with the latest master branch the code should not compile anymore.

The sequence of calls with the latest master has to be like this:
- `model->AddField(std::move(field)) // (for all fields)`
- `model->Freeze(); // no more fields can be added, the default entry becomes available`
- `model->GetDefaultEntry()->CaptureValueUnsafe("fieldName", ptrToObject); // for all fields`

jblomer: To give a bit more context: the idea behind the API change is to clearly separate the two phases of the model life time: phase 1 where the model is built; once built, phase 2 where it can be used to generate entries. We now forbid generating entries in phase 1 to avoid "half-baked entries".

PR #252, opened 2021-12-08
Make collections movable

tmadlener:
BEGINRELEASENOTES
- Make the `CollectionData` classes use `unique_ptr` instead of raw pointers, wherever they actually own the pointer.
- Implement move constructors and move assignment operators for collections. Thanks to the usage of `unique_ptr` for ownership management in the `CollectionData`, these can be `default`ed in `Collection` and `CollectionData`.
- Add a few tests to check that moving collections actually works.

ENDRELEASENOTES

- [x] Builds on #251

### DIFF
--- a/include/podio/CollectionBuffers.h
+++ b/include/podio/CollectionBuffers.h
@@ -6,11 +6,15 @@
 #include <string>
 #include <utility>
 #include <vector>
+#include <memory>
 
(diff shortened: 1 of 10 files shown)


hegner:
Thanks for updating after our discussion. :-)

PR #254, opened 2021-12-09
Add clang-format and clang-tidy configurations

tmadlener:
BEGINRELEASENOTES
- Add a `.clang-format` and `.clang-tidy` config file for consistent formatting and following a few coding guidelines.
- Add pre-commit hooks that run `clang-tidy` and `clang-format` 
- Make all currently present files follow the formatting and guidelines of the present configuration.
- Make the `PODIO_GENERATE_DATAMODEL` macro look for a `.clang-format` file and the presence of `clang-format`and automatically format all the generated files if both are there.

ENDRELEASENOTES

- [x] Needs #253 (starts from where that ends).
- Cheery picked from #188 (see discussion there for the resulting `.clang-format` and `.clang-tidy` configuration.

PR #257, opened 2022-01-31
Add templated links between arbitrary datatypes

### LINKED ISSUE #255: Allow for relations between two arbitrary types
Currently to relate two collections or objects of two collections a dedicated type is necessary. This datatype usually has a `float weight` as single datamember and then two `OneToOneRelation`s to connect two objects.

E.g. from [EDM4hep](https://github.com/key4hep/EDM4hep/blob/b1b0b2bb68009c253079f652423ede4b4dec6b5e/edm4hep.yaml#L413-L420)
```yaml
  edm4hep::MCRecoParticleAssociation:
    Description: "Used to keep track of the correspondence between MC and reconstructed particles"
    Author: "C. Bernet, B. Hegner"
    Members:
      - float weight                        // weight of this association
    OneToOneRelations :
     - edm4hep::ReconstructedParticle  rec  // reference to the reconstructed particle
     - edm4hep::MCParticle sim              // reference to the Monte-Carlo particle
```

LCIO allows for arbitrary relations between two objects, via the [`LCRelation`](https://ilcsoft.desy.de/LCIO/current/doc/doxygen_api/html/classEVENT_1_1LCRelation.html). In order to handle all of the relations that are used in the wild many new `Association` types have to be introduced to EDM4hep, e.g.
- key4hep/EDM4hep#122
- key4hep/EDM4hep#133

It might be simpler to implement a generic (templated) association or relation type in podio that then allows to use (and persist) arbitrary relations between objects.

tmadlener:
BEGINRELEASENOTES
- Add a templated `Link` (and corresponding `LinkCollection`) class to allow for arbitrary links between two podio generated datatypes.
  - The `Link`s work similar to a podio generated class of the form
```yaml
Link:
  Members: [float weight]
  OneToOneRelations:
    - FromT from
    - ToT to
```
- They offer templated access to the objects they link via `get<T>` on top of the usual "generated" access functions. The access functions are using a `get` prefix, e.g. `getFrom`.
- They can be used in memory simply by declaring a collection via `podio::LinkCollection<F, T>`. **In order to enable I/O for link collections it is necessary to register them via the `PODIO_DECLARE_LINK` macro**

ENDRELEASENOTES

Fixes #255 

-------
**NOTE (Jul 29 2024): The following discussion and descriptions span quite a bit of time and the original proposal used `Association` instead of `Link` as it's name (see https://github.com/key4hep/EDM4hep/issues/344 for some reasoning of that). Refer to the included `links.md` file for the latest version of the documentation with usage examples**. We leave the discussions unchanged below.

-------

### Preliminary (user facing) documentation

The main template that is introduced is the `AssociationT` and two type-aliases to have mutable and immutable `Association` types:
```cpp
template<typename FromT, typename ToT, bool Mutable>
class AssociationT;

template<typename FromT, typename ToT>
using Association = typename AssociationT<FromT, ToT, false>;

template<typename FromT, typename ToT>
using MutableAssociation = typename AssociationT<FromT, ToT, true>;
```
The main interface for the user is that of the similarly templated `AssociationCollection`, which uses as the template parameters  the **immutable** (i.e. default) types, like
```cpp
#include "podio/AssociationCollection.h"
#include "edm4hep/MCParticle.h"
#include "edm4hep/ReconstructedParticle.h"

using MCRecoParticleAssociationCollection = podio::AssociationCollection<edm4hep::MCParticle, edm4hep::ReconstructedParticle>;
```
from here on the collection behaves the same as a normal collection (i.e. as if it would have been generated via a yaml definition file), i.e.
```cpp
auto mcRecoAssocs = MCRecoParticleAssociationCollection();
auto mcParticles = edm4hep::MCParticleCollection();
auto recoParticles = edm4hep::ReconstructedParticleCollection();

// fill mc and reco collections

auto assoc = mcRecoAssocs.create(); // now have a mutable association
assoc.setWeight(3.14f); // set a weight for the association
assoc.setFrom(mcParticles[0]); // set a MCParticle as the from part of the association
assoc.setTo(recoParticles[0]); // set a ReconstructedParticle as the to part of the association

// getting associations from store
auto& associations = store.get<MCRecoParticleAssociationCollection()>("associations");
for (auto a : associations) {
  // a is now an immutable association
  auto mc = a.getFrom(); // get the MCParticle as the from part of the association
  auto reco = a.getTo(); // get the ReconstructedParticle as the to part of the association
  auto weight = a.getWeight(); // get the weight
}
```
It is also possible to use the relations as subset collections as they use the same mechanism under the hood as collections of "proper" datatypes.

### Some internal technicalities
- Each generated datatype now has `DefT`, `MutT` and `CollT` type-aliases to get the default (immutable), mutable and collection types respectively from them. This is used internally for making sure that we return the correct types to the user in the end. In order to not have to check in all templates, there are a few `static_assert`s at the "entry points" to the associations to make sure we can assume that the `FromT` and `ToT` template parameters are the default types internally.
- The collection iterators use the same approach as the associations, i.e. there is a generalized template with a boolean parameter and two type-aliases to get to the immutable and mutable iterators. This avoids code duplication, but has the slight drawback that some `enable_if` code is necessary, to make sure that the conversion from mutable to immutable is possible but not vice-versa.

Includes changes from:
- [x] #253 
- [x]  #254
- [diff without these changes](https://github.com/tmadlener/podio/compare/clang-format-tidy...tmadlener:associations?expand=1)

### TODOs:
- [x] ROOT I/O not yet working (need to figure out how to make the dictionary generation pick up the templated class)
- [x] SIO backend needs manual intervention at the moment, as the user has to declare a `static` instantiation of each association they want to use
- [x] clang-tidy actually flags #174 as it apparently sees through all the template instantiations and is able to reason about memory access patterns here. For the moment this is silenced via a `NOLINT` directive in the `podio::ObjBase`
- [x] Finalize documentation and put into a markdown file. (see [here] (https://github.com/tmadlener/podio/blob/associations/doc/associations.md) for current version)
- [x] Templated `get/set` functionality (see [comment](https://github.com/AIDASoft/podio/pull/257#issuecomment-1036011167))
- [x] Needs #394 (cherry-picked to here)
- [x] Needs #399 (cherry-picked to here)
- [x] Needs #400 (cherry-picked to here)
- [x] Needs rework after #514 
- [x] Includes #548 
- [x] Includes #673 
- [x] Includes #687
- [x] Includes #689


tmadlener:
Should the association interface have "generic" `get` and `set` functions in addition to `get(To|From)` and `set(To|From)`? I.e. something that would look like:
```cpp
using namespace edm4hep;
using MCRecoParticleAssociation = podio::Association<MCParticle, ReconstructedParticle>;
auto assoc = MCRecoParticleAssociation();

// get the MCParticle
auto mcP = assoc.getFrom(); // implicitly refers to the MCParticle

auto mcP2 = assoc.get<MCParticle>(); // get the MCParticle explicitly
auto reco = assoc.get<ReconstructedParticle>(); // get the ReconstrucedParticle explicitly
```
The untemplated versions are nice to use where the association is somewhat "directed", while for "symmetric" associations `To` and `From` do not really make sense, I think.

andresailer:
```c++
auto mcP2 = assoc.get<MCParticle>(); 
```
Having to and from be the same type isn't prohibited, right? So this wouldn't work

Something like the tuple `get` (https://en.cppreference.com/w/cpp/utility/tuple/get)
```c++
auto mcP1 = assoc.get<0>(); 
auto mcP2 = assoc.get<1>(); 
auto weight = assoc.get<2>(); 
```

tmadlener:
> Having to and from be the same type isn't prohibited, right? So this wouldn't work

Good point. I didn' think about that. In principle we could forbid associations between the same type (by failing this at compile time), but there probably are legitimate use cases for that.

> Something like the tuple `get` 

We could probably do that, but then I don't see the additional value compared to `getTo`, `getFrom`, and `getWeight`, because you will still have to know which index refers to which part of your association, same as with the `From` and `To` scheme.

Alternatively we could only make the templated `get`/`set` functionality available for associations which have different types? Or we leave that out for now and see how far we get with simply `From`, `To`?

andresailer:
* Make naming consistent over all associations
* Potential to get confusing (without proper associations in podio, see https://github.com/AIDASoft/podio/pull/257)
    * Which member should be `from` and which one should be `to`?
    * "Higher level" should be `from` and "lower level" (i.e. MC) should be `to`
    * Want a templated version and maybe some utitlities to make working with associations easier (e.g. create a map on the fly from the contents of the collection like in LCIO)
cf. https://github.com/key4hep/EDM4hep/pull/134

tmadlener:
Using the proposed functionality of having a central entity that hands out collection buffers, the need for building all possible combinations up front can be alleviated at the cost of having to use a macro (and another function call in the SIO case) to do this automatically at compile time. Usability is still not as good as desired, and I would like to improve that. Also the python side of things still needs to be handled as that currently does not have the templated things compiled into a library which it could simply load.

tmadlener:
With #405 the name of the branches will also have to be considered in the whole "generatlity" aspect

doc/associations.md
```diff
@@ -0,0 +1,233 @@
...
+mind, effectively all mutating operations on `Association`s are defined using
+SFINAE using the following template structure (taking here `setFrom` as an
+example)
+
+```cpp
+template <typename FromU,
+          typename = std::enable_if_t<Mutable && 
+                                      std::is_same_v<detail::GetDefT<FromU>, FromT>>>
+void setFrom(FromU value);
+```
```
tmadlener, line 214: This needs a revision. Some of these have made it into mainline podio in a slightly different way since this was originally written

tmadlener:
Revived this one more time (maybe this is the last now). This time mainly making sure to incorporate changes that came possible with the `MaybeSharedPtr` and the introduction of some member type defs (#465). No new features were added wrt the last revival.

python/templates/MutableObject.h.jinja2
```diff
@@ -43,6 +43,12 @@ public:
   operator {{ class.bare_type }}() const;
 
 public:
+  // Some typedefs for the Association handling
+  using MutT = {{ class.namespace }}::Mutable{{ class.bare_type }};
+  using DefT = {{ class.namespace }}::{{ class.bare_type }};
+  using CollT = {{ class.namespace }}::{{ class.bare_type }}Collection;
+
+  static constexpr auto TypeName = "{{ class.namespace }}::Mutable{{ class.bare_type }}";
```
tmadlener, line 51:
```suggestion
```

Unnecessary

tests/read_test.h
```diff
@@ -28,6 +29,11 @@
 #include <type_traits>
 #include <vector>
 
+// Define an association that is used for the I/O tests
+using TestAssocCollection = podio::AssociationCollection<ExampleMC, ex42::ExampleWithARelation>;
+// Make sure the I/O parts are dynamically registered
+PODIO_DECLARE_ASSOCIATION(ExampleMC, ex42::ExampleWithARelation)
```
tmadlener, line 35: This could be unnecessary since we build a mini library now that has this.

tests/unittests/associations.cpp
```diff
@@ -0,0 +1,377 @@
...
+    REQUIRE(c == cluster);
+    REQUIRE(w == 1.0);
+  }
+}
+
+TEST_CASE("AssociationCollection basics", "[associations]") {
+  auto coll = TestAColl();
+
+  // TODO: basics without I/O
+}
```
tmadlener, line 170:
```suggestion
```
The basic collection functionality is covered by other tests as well

tests/write_frame.h
```diff
@@ -19,12 +19,18 @@
...
 #include "podio/Frame.h"
 #include "podio/UserDataCollection.h"
 
 #include <string>
 #include <tuple>
 
+// Define an association that is used for the I/O tests
+using TestAssocCollection = podio::AssociationCollection<ExampleMC, ex42::ExampleWithARelation>;
+// Make sure the I/O parts are dynamically registered
+PODIO_DECLARE_ASSOCIATION(ExampleMC, ex42::ExampleWithARelation)
```
tmadlener, line 32:
```suggestion
```
Same as for reading

tmadlener:
It looks like `clang-tidy` is a bit too cautious around some of the allocations that happen in the Associations. Since no actual leaks seem to happen when using sanitizers, I have opted to simply silence clang-tidy selectively for some of the tests.

tmadlener:
After the discussion in https://github.com/key4hep/EDM4hep/issues/344 I have renamed all `Association`s to `Link`s.

hegner:
@tmadlener - is it possible to separate some of the code cleanup into a separate PR?

tmadlener:
This is already pretty much isolated. There is #673 and there is also #687 as an independent change. I will have another look, to see if there is more that can be moved.

tests/read_frame_auxiliary.h
```diff
@@ -62,6 +62,8 @@ int test_frame_aux_info(const std::string& fileName) {
 
   auto collsToRead = collsToWrite;
   if (reader.currentFileVersion() < podio::version::Version{0, 16, 3}) {
+    collsToRead.erase(collsToRead.end() - 4, collsToRead.end());
+  } else if (reader.currentFileVersion() < podio::version::Version{1, 0, 99}) {
```
tmadlener, line 66:
```suggestion
  } else if (reader.currentFileVersion() < podio::version::Version{1, 1, 99}) {
```

tests/read_test.h
```diff
@@ -410,6 +413,43 @@ void processEvent(const podio::Frame& event, int eventNum, podio::version::Versi
       }
     }
   }
+
+  // ======================= Links ==========================
+  if (fileVersion >= podio::version::Version{1, 0, 99}) {
```
tmadlener, line 418:
```suggestion
  if (fileVersion >= podio::version::Version{1, 1, 99}) {
```

tests/unittests/links.cpp
```diff
@@ -0,0 +1,475 @@
...
+    REQUIRE(mutLink.getWeight() == 42.0f);
+    REQUIRE(mutLink.getFrom() == otherHit);
+    REQUIRE(mutLink.getTo() == otherCluster);
+  }
+
+  SECTION("Implicit conversion") {
+    // Use an immediately invoked lambda to check that the implicit conversion
+    // is working as desired
+    [hit, cluster](TestL link) { // NOLINT(performance-unnecessary-value-param):
+                                 // We want the value here to force teh
```
tmadlener, line 95:
```suggestion
                                 // We want the value here to force the
```

include/podio/detail/LinkCollectionIterator.h
```diff
@@ -0,0 +1,47 @@
...
+  bool operator!=(const LinkCollectionIteratorT& other) const {
+    return m_index != other.m_index; // TODO: may not be complete
+  }
+
+  LinkType operator*() {
+    m_object.m_obj = podio::utils::MaybeSharedPtr<LinkObjT>((*m_collection)[m_index]);
+    return m_object;
+  }
+
+  LinkType* operator->() {
```
hegner, line 30: what's the const qualifier here?

include/podio/detail/LinkObj.h
```diff
@@ -0,0 +1,53 @@
...
+  /// Destructor
+  ~LinkObj() {
+    delete m_from;
+    delete m_to;
+  }
+
+public:
+  podio::ObjectID id{};
+  LinkData data{1.0f};
+  FromT* m_from{nullptr};
```
hegner, line 47: do we want to use some smart pointer here instead of bare ones?

hegner (changes requested):

include/podio/detail/LinkObj.h
tmadlener: These pointers don't manage anything. Additionally, this is also what "equivalent" generated code would look like, so they can stay as bare ones.

hegner: OK. Fine. If that's the only delete I am fine.

include/podio/LinkCollection.h
```diff
@@ -0,0 +1,35 @@
...
+
+#include "podio/detail/LinkCollectionImpl.h"
+
+// Preprocessor helper macros for concatenating tokens at preprocessing times
+// Necessary because we use __COUNTER__ below for unique names of static
+// variables for values returned by registration function calls
+#define PODIO_PP_CONCAT_IMPL(x, y) x##y
+#define PODIO_PP_CONCAT(x, y) PODIO_PP_CONCAT_IMPL(x, y)
+
+#ifndef PODIO_ENABLE_SIO
```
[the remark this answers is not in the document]
tmadlener: I thought about putting the `PODIO_PP_CONCAT{,_IMPL}` into a separate header, but we don't have any other uses yet, so I left them here. Should we move them to `include/podio/detail/PreprocessorMacros.h`?

include/podio/detail/LinkCollectionIterator.h
tmadlener: They cannot be easily marked `const` because we switch the object that we return during the access (and there is no other place where we could do this). I don't think it's required by anything that these access operators are `const`.

hegner: OK. Fine then.

include/podio/LinkCollection.h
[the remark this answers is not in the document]
hegner: Yes, that could be helpful!

include/podio/LinkCollection.h
```diff
@@ -0,0 +1,35 @@
...
+#if PODIO_ENABLE_SIO && __has_include("podio/detail/LinkSIOBlock.h")
+  #include "podio/detail/LinkSIOBlock.h"
+  /// Main macro for declaring links. Takes care of the following things:
+  /// - Registering the necessary buffer creation functionality with the
+  ///   CollectionBufferFactory.
+  /// - Registering the necessary SIOBlock with the SIOBlock factory
+  #define PODIO_DECLARE_LINK(FromT, ToT)                                                                               \
+    const static auto PODIO_PP_CONCAT(REGISTERED_LINK_, __COUNTER__) =                                                 \
+        podio::detail::registerLinkCollection<FromT, ToT>(podio::detail::linkCollTypeName<FromT, ToT>());              \
+    const static auto PODIO_PP_CONCAT(LINK_SIO_BLOCK_, __COUNTER__) = podio::LinkSIOBlock<FromT, ToT>{};
```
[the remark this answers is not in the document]
tmadlener: There is no (at least no straight forward) way to have an `#if` inside a pre-processor macro. I think the best we could do here is to define separate macros for each purpose and then define the summary macro in terms of them?

include/podio/detail/LinkObj.h
tmadlener: Ah wait, I think I spoke to soon. Maybe these should be `unique_ptr`s after all, since they are indeed manually `delete`d. I was mixing things up with the `Obj*` that are inside these pointers.

include/podio/LinkCollection.h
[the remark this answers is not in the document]
tmadlener: Done.

include/podio/detail/LinkObj.h
tmadlener: This now done with `unique_ptr`s. Since I need the `Link`s and other generated code to behave the same, I have made the changes for the generated classes in #689 (which is now also included in this PR, so that has to be merged first).

include/podio/LinkCollection.h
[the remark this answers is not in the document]
tmadlener: I have redefined this "top level" macro in terms of two other macros that now contain one registration each. It doesn't really help us for the combinatorics, but at least we are just calling into macros at that point.

tests/datamodel/LinkCollections.h
```diff
@@ -0,0 +1,15 @@
...
+
+#include "podio/LinkCollection.h"
+
+#include "datamodel/ExampleClusterCollection.h"
+#include "datamodel/ExampleHitCollection.h"
+#include "datamodel/TypeWithEnergy.h"
+
+// Define an link that is used for the I/O tests
+using TestLinkCollection = podio::LinkCollection<ExampleHit, ExampleCluster>;
+// Define a link with an interace type that is used for I/O tests
```
m-fila, line 12:
```suggestion
// Define a link with an interface type that is used for I/O tests
```

tests/unittests/links.cpp
```diff
@@ -0,0 +1,475 @@
...
+  REQUIRE(link.get<TypeWithEnergy>() == hit2);
+
+  link.set(hit);
+  REQUIRE(link.get<TypeWithEnergy>() == hit);
+  REQUIRE(link.get<ExampleCluster>() == cluster);
+}
+
+#ifdef PODIO_JSON_OUTPUT
+TEST_CASE("Link JSON converstion", "[links][json]") {
+  const auto& [links, hits, clusters] = createLinkCollections();
```
m-fila, line 456:
```suggestion
TEST_CASE("Link JSON conversion", "[links][json]") {
```

tests/unittests/links.cpp
```diff
@@ -0,0 +1,475 @@
...
+      REQUIRE(a.getWeight() == index++);
+    }
+  }
+
+  SECTION("Conversion failures") {
+    // Cannot convert into a subset collection if elements already present
+    REQUIRE_THROWS_AS(links.setSubsetCollection(), std::logic_error);
+
+    // Connot convert a subset collection into a normal collection
+    REQUIRE_THROWS_AS(linkRefs.setSubsetCollection(false), std::logic_error);
```
m-fila, line 305:
```suggestion
    // Cannot convert a subset collection into a normal collection
```

include/podio/detail/Link.h
```diff
@@ -0,0 +1,357 @@
...
+    }
+  }
+
+  /// Templated version for setting an element of the link by type. Only
+  /// available for Links where FromT and ToT are **not the same type**,
+  /// and if the requested type is actually part of the Link. Will result
+  /// in a compilation error if any of these conditions is not met.
+  ///
+  /// @tparam T type of value (**infered!**)
+  /// @param value the element to set for this link.
```
m-fila, line 255:
```suggestion
  /// @tparam T type of value (**inferred!**)
```

include/podio/detail/Link.h
```diff
@@ -0,0 +1,357 @@
...
+#endif
+
+#include <ostream>
+#include <utility> // std::swap
+
+namespace podio {
+
+/// Generalized Link type for both Mutable and immutable (default)
+/// versions. User facing clases with the expected naming scheme are defined via
+/// template aliases in LinkFwd.h
```
m-fila, line 20:
```suggestion
/// versions. User facing classes with the expected naming scheme are defined via
```

include/podio/utilities/TypeHelpers.h
```diff
@@ -178,6 +181,20 @@ namespace detail {
   template <typename T>
   using hasObject_t = typename T::object_type;
 
+  /// Detector for checking the existance of a collection_type type member.
+  template <typename T>
```
m-fila, line 185:
```suggestion
  /// Detector for checking the existence of a collection_type type member.
```

tmadlener:
Is there anything left here to do? Otherwise I would merge this and then address issues in other PRs. Currently we have quite a few PRs that depend on this and that have very large diffs because they include this PR.

tmadlener:
I have no idea what makes the EDM4hep CI fail. The reason is that these workflows no longer discover the podio version that has just been built in the EDM4hep `find_package(podio 1.1 REQUIRED HINTS $ENV{PODIO})`, but rather the one from the underlying LCG stacks. However, nothing in this PR should change the behavior of that `find_package` call? Effectively, I just bump the patch version here, but that should not lead to any changes, I think. @jmcarcell @andresailer what am I missing here?

tmadlener:
It actually looks like the correct podio version is found, but the wrong python bindings(?). I cannot reproduce this in a local key4hep stack though.

jmcarcell:
Isn't the test correct and the file was written with the previous version of podio which doesn't match the current one anymore?

PR #258, opened 2022-02-03
Add preprocessor macros to retrieve versions

SanghyunKo:
BEGINRELEASENOTES
- Add preprocessor macros to retrieve versions

ENDRELEASENOTES

I've realized that retrieving versions for preprocessing is somehow quite tricky in Podio. Adding a preprocessor macro would make it more helpful to developers (support backward compatibility for lcg releases, .etc). I believe that the macro will be okay as long as we maintain `major.minor.patch` convention for the version (mimicked how ROOT does for versioning [[link](https://root.cern.ch/doc/master/RVersion_8h_source.html)]), but please let me know if there is any kind of issue you concern. Thanks.

### DIFF
--- a/podioVersion.h
+++ b/podioVersion.h
@@ -1,4 +1,9 @@
 #ifndef podioVersion_h
 #define podioVersion_h
 #define podio_VERSION @podio_VERSION@
+#define podio_VERSION_MAJOR @podio_VERSION_MAJOR@
+#define podio_VERSION_MINOR @podio_VERSION_MINOR@
+#define podio_VERSION_PATCH @podio_VERSION_PATCH@
+#define podio_VERSION_ENCODE(a,b,c) (((a) << 16) + ((b) << 8) + (c))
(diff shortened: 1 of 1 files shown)


tmadlener:
Hi Sanghyun,
thanks for this. I agree that this is quite cumbersome at the moment. There is a very similar pull request already: #238

I think the preprocessor macros are almost identical (apart from using 16 instead of 8 bits). Additionally that would also offer a `Version` class that can be used at runtime. Maybe you could check if something is missing there for your purposes?

SanghyunKo:
Hi @tmadlener, thanks for the comment. Indeed, I didn't notice there is already a similar PR for this. It looks more than enough for me already (and more ambitious in the sense of controlling the file's version, not just only the library's version). I'll close this PR.

The last question from my side is - do we target to merge #238 for v00-15? I hope so since this will be quite useful when we introduce `MutableClass` (#205) for the next release.

tmadlener:
> The last question from my side is - do we target to merge #238 for v00-15? I hope so since this will be quite useful when we introduce `MutableClass` (#205) for the next release.

Yes, I think we definitely aim for that.

SanghyunKo:
Thanks @tmadlener!

PR #259, opened 2022-02-09
Filter files with regex based on file name, not complete path

fdplacido:
BEGINRELEASENOTES
- Filter files with regex based on file name, not complete path

ENDRELEASENOTES

If the working directory contained the string `test_` as in doing `mkdir test_podio`, the regex would match all entries of files to install, installing none. This filtering matches based on filename.


hegner:
@fdplacido - thanks. That comes just in time before I was about to cut a new tag

PR #261, opened 2022-02-24
Make fixed width integer type arrays pass the validation

tmadlener:
BEGINRELEASENOTES
- Make the datamodel validation accept arrays of fixed width integer types.

ENDRELEASENOTES

This should have always been allowed, but the parsing was not filling the necessary information for the validation to pass in this case.


tmadlener:
@hegner can we also include this before we have a new tag?

PR #262, opened 2022-02-24
Add templated get/set functionality for GenericParameters and deprecate non-templated versions

tmadlener:
BEGINRELEASENOTES
- Make the getters and setters for the `GenericParameters` templated functions and add a deprecation warning for the untemplated ones.
- Define a `SupportedGenericDataTypes` tuple defining the types (and vectors of those) that can be stored in `GenericParameters`
- Add a `podio/utilities/TypeHelpers.h` header with some type handling helpers.

ENDRELEASENOTES

Removing the non-templated functions will result in some minor differences in the setting interface, mainly that the `vector` and single value functions now both use the `setValue` function instead of separate `setValue` and `setValues`. On the getter side there are a few more differences, most importantly that the functions that return `vector`s now have exactly the same interface as the single value getters. Previously the `vector` ones used a pre-allocated `vector` to fill the values into. Additionally, they returned a const-ref to the passed in vector, which leads to having two references with different mutability to the same vector. Overall the interface becomes more uniform and adding additional types, essentially requires to change some compile time configuration instead of implementing additional named overloads.

Previously:
```cpp
auto& md = store.getEventMetaData();
// set values
md.setValue("someInt", 42);
std::vector<std::string> strings = {"abc", "def", "ghi"};
md.setValues("someStrings", strings);

// get values from the meta data
const auto& strVal = md.getStringVal("key");  // return by const-ref
const auto intVal = md.getIntVal("someInt");  // return by value
std::vector<float> floatValues;
const auto& floatVals = md.getFloatVals("anotherKey", floatValues);  // need to pass in a pre-allocated vector
// Now floatValues and floatVals are references to the same vector with different mutability
```

Now:
```cpp
auto& md = store.getEventMetaData();
// set values
md.setValue("someInt", 42);
std::vector<std::string> strings = {"abc", "def", "ghi"};
md.setValue("someStrings", strings);

// get values
const auto& strVal = md.getValue<std::string>("key");  // return by const-ref
const auto intVal = md.getValue<int>("someInt");       // return by value
const auto& floatValues = md.getValue<std::vector<float>>("anotherKey");
// Only one const ref to the contents
```

This is a somewhat preparatory step for future developments that can however "stand on its own". To keep things working downstream the non-templated ones are now simply slated for deprecation.

@fdplacido Does something like this make conversions easier to handle or would you like to keep the non-templated functions?

~Needs other PRs first:~
- [x] targets `develop` for the moment for a cleaner diff.

### DIFF
--- a/include/podio/GenericParameters.h
+++ b/include/podio/GenericParameters.h
@@ -201,10 +201,10 @@ public:
 private:
   /// Get a reference to the internal map for a given type (necessary for SIO)
   template <typename T>
-  const MapType<T>& getMap() const {
-    if constexpr (std::is_same_v<T, int>) {
+  const MapType<detail::GetVectorType<T>>& getMap() const {
+    if constexpr (std::is_same_v<detail::GetVectorType<T>, int>) {
       return _intMap;
-    } else if constexpr (std::is_same_v<T, float>) {
+    } else if constexpr (std::is_same_v<detail::GetVectorType<T>, float>) {
       return _floatMap;
     } else {
       return _stringMap;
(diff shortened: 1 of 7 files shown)


tmadlener:
closing and re-opening, because github doesn't seem to pick up the latest commit (is that even possible?)

tmadlener:
Rebased to target `master` now that all pre-requisites have been merged.

fdplacido:
This would be fine for the converters. If any change is needed it would be minor: quick and easy to fix.

PR #265, opened 2022-03-18
CI: update compilers for dev builds

andresailer:
The clang10 builds are suppressed and no longer updated.

Other question is whether to replace LCG_99 with 101 (or wait until 102 comes out)

BEGINRELEASENOTES
-  CI: use clang12 and gcc11 for tests based on dev stacks

ENDRELEASENOTES


tmadlener (approved):

andresailer:
LCG_102 should come out within a week or two.

tmadlener:
Then, I will merge this now, and we can migrate to those with the rest, once they are available.

Issue #266, opened 2022-03-21
Allow setting default initializers on member definition


wdconinc:
Data type components are currently defined as, e.g.,
```yaml
    # Vector3D with floats
    edm4hep::Vector3f :
      Members:
        - float x
        - float y
        - float z
      ExtraCode:
        declaration: "
        Vector3f() : x(0),y(0),z(0) {}\n
        Vector3f(float xx, float yy, float zz) : x(xx),y(yy),z(zz) {}\n
        "
```

A common pattern is to set the default initialization of a type (e.g. `Vector3f() : x(0),y(0),z(0) {}`) which requires a constructors. Since C++11 we can use default initializer lists for non-static members, https://en.cppreference.com/w/cpp/language/data_members. This would allow (suggest?) a syntax in yaml:
```yaml
    # Vector3D with floats
    edm4hep::Vector3f :
      Members:
        - float x{0}
        - float y{0}
        - float z{0}
      ExtraCode:
        declaration: "
        Vector3f(float xx, float yy, float zz) : x(xx),y(yy),z(zz) {}\n
        "
```
In this example, chosen for its simplicitly, this would only get rid of a single default constructor. Other examples would be differently affected (e.g. a 4d covariance matrix with a default, diagonal, and full constructors).

tmadlener:
This looks like a very reasonable addition. Plus we will most likely need something like this for schema evolution in any case.

For clarification (and also because I couldn't find this in the documentation quickly): The data members of components and datatypes are default initialized in any case, so the `Vector3f` definition above will be generated to
```cpp
namespace edm4hep {
struct Vector3f {
  float x{};
  float y{};
  float z{};
  // ExtraCode ...
};
}
```

Adding the possibility to have user defined default values in the yaml file, should not be too complicated. From a syntax point of view, I think `{<value>}` is as good as ` = <value> `.

I suppose we would want an optional default (value) initialization in the yaml file, or would we mandate that users provide default values for all members?

wdconinc:
I think, in the interest of backwards compatibility, keeping it optional would be my preference.

tmadlener:
Just to better define the overall scope:
- Should the default initialization with user provided values also work for arrays? If we do, how far should we go in validation already on the python side?
- How far should we try and validate "reasonable" default values already on the python side? e.g. should we check that `int`s are actually initialized by an integer value and not for example a `float` value.

[...]

PR #268, opened 2022-03-23
[WIP] Make it possible for users to define default values in the datamodel yaml file

tmadlener:
BEGINRELEASENOTES
- Allow users to define default values for member variables, instead of default initializing all of them.
  - The syntax for specifying a default value is `- <type> <name> {<init-value>} // <description>`.
  - If the `type` is a builtin type or an array of builtin types, the initialization value is validated to yield valid c++. For other types this validation is skipped and it is the users responsibility to provide a valid initialization.
- Remove some of the python2 compatibility now that it is no longer necessary.

ENDRELEASENOTES

See #266 for some initial discussion on the supported use cases

Currently targets the `develop` branch, which simply includes #254 and #253 on top of the current `master` branch.

~This also needs #270 for one of the tests to work.~

- [ ] update documentation
- [ ] stricter validation?
- [x] check impact on time it takes to generate, e.g. edm4hep with the additional checks.

### DIFF
--- a/python/test_MemberParser.py
+++ b/python/test_MemberParser.py
@@ -140,6 +140,13 @@ class MemberParserTest(unittest.TestCase):
     self.assertEqual(parsed.name, 'f')
     self.assertEqual(parsed.default_val, '3.14f')
 
+    parsed = parser.parse(r'std::array<int, 3> array{1, 2, 3} // arrays can be initialized')
+    self.assertEqual(parsed.full_type, r'std::array<int, 3>')
+    self.assertEqual(parsed.default_val, '1, 2, 3')
+    self.assertEqual(parsed.name, 'array')
+
+    # These are cases where we cannot really decide whether the initialization
+    # is valid just from the member declaration. We let them pass here
     parsed = parser.parse('nsp::SomeValue val {42} // default values can have space')
     self.assertEqual(parsed.full_type, 'nsp::SomeValue')
     self.assertEqual(parsed.name, 'val')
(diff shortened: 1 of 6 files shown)


python/test_MemberParser.py
```diff
@@ -121,6 +123,44 @@ def test_parse_valid(self):  # pylint: disable=too-many-statements
...
+    self.assertEqual(parsed.default_val, '42')
+    self.assertEqual(parsed.namespace, 'nsp')
+    self.assertEqual(parsed.bare_type, 'SomeValue')
+
+    parsed = parser.parse(r'edm4hep::Vector3d v{1, 2, 3, 4} // for aggregates we do not validate init values')
+    self.assertEqual(parsed.full_type, 'edm4hep::Vector3d')
+    self.assertEqual(parsed.default_val, '1, 2, 3, 4')
+
+    parsed = parser.parse(r'AggType bogusInit{here, we can even put invalid c++}', False)
+    self.assertEqual(parsed.default_val, 'here, we can even put invalid c++')
```
tmadlener, line 162: Examples of things that we cannot really validate without further context, and that we currently pass through verbatim to the generation. Both of these would lead to compilation errors later.

python/test_MemberParser.py
```diff
@@ -148,12 +189,23 @@ def test_parse_invalid(self):
...
+        r'std::array<uint_fast16_t, 42> disallowedArray // arrays should not accept disallowed fixed width types',
+
+        # Default values cannot be empty
+        r'int emptyDefault{} // valid c++, but we want an explicit default value here',
+
+        # Some cases which should be caught by the -fsyntax-only checks
+        r'std::array<int, 1> array{1, 2, 3} // too many values provided',
+        # Invalid user default value initialization
+        r'int weirdDefault{whatever, even space} // invalid c++ should not pass',
+        r'int initFromFloat{3.14f} // implicit conversions are not allowed'
```
tmadlener, line 201: Examples of things we can actually catch during validation, because we have enough context from the member definition alone.

tmadlener:
The additional validation with the frequent calls to gcc have a quite severe impact on the runtime of the generation script actually. For edm4hep the generation time goes from 0.3 s to almost 3 s with the additional validation. That would be quite prohibitive for workflows were the generation happens frequently. Should there be a way to at least disable these checks for quicker iterations in such cases?

Issue #272, opened 2022-03-25
Deleted iterator copy constructor/assignment and `find`ing


wdconinc:
I suspect there are good reasons to delete the iterators in [python/template/macros/iterator/iterator.jinja2](https://github.com/AIDASoft/podio/blob/master/python/templates/macros/iterator.jinja2), but it makes searching through a collection a bit more cumbersome at times. I am wondering if there are ways in which some of the choices on iterator copy constructors and assignment operators deletion can be relaxed, or am I missing something I should be using instead?

The following are attempts at finding the first electron in an MCParticlesCollection. I am using gcc-11 with `-std=gnu++20`.

## Using `find_if` to get an iterator to the first electron
```c++
    const auto mc_first_electron = std::find_if(
      mcparts.begin(),
      mcparts.end(),
      [](const auto& p){ return p.getPDG() == 11; });
```
This fails because MCParticleCollectionIterator has a deleted copy constructor which is required (likely) to return a copy of the internal iterator inside std::find_if.

## Using a simple (old-style) iterator loop
If we can't copy an iterator, then let's avoid the copy.
```c++
    auto mc_first_electron = mcparts.end();
    for (auto p = mcparts.begin(); p != mcparts.end(); p++) {
      if (p->getPDG() == 11) {
        mc_first_electron = p;
        break;
      }
    }
```
This fails because iterator post-increment is not implemented. Ok, easy enough to change (and [arguably better](https://betterprogramming.pub/stop-using-i-in-your-loops-1f906520d548) though unlikely that an iterator takes up so much memory this matters here; still, it is good to instill best practices).
```c++
    auto mc_first_electron = mcparts.end();
    for (auto p = mcparts.begin(); p != mcparts.end(); ++p) {
      if (p->getPDG() == 11) {
        mc_first_electron = p;
        break;
      }
    }
```
This still fails but now because we cannot assign p to mc_first_electron since the assignment operator for iterators is deleted. Ok, we can avoid this assignment too, by avoiding a locally scoped iterator.
```c++
    auto mc_first_electron = mcparts.begin();
    for (; mc_first_electron != mcparts.end(); ++mc_first_electron) {
      if (mc_first_electron->getPDG() == 11) {
        break;
      }
    }
```
Ah, we have something that compiles.

## Checking that we actually found something
After the search, we want to make sure we found something (and return if not).
```c++
    if (mc_first_electron == mcparts.end()) {
      debug() << "No electron found" << endmsg;
[...]

PR #273, opened 2022-03-28
input_iterator type and unit tests

wdconinc:
Some changes to the iterators to make https://github.com/AIDASoft/podio/issues/272 work. This now enables all except for C++ ranges, and there are unit tests to go with it. C++20 unit tests still fail. This applies to all iterators, both mutable and const, but I think for input_iterators that is fine. For the forward iterators we probably need to split the iterators.jinja2 into two(?).

BEGINRELEASENOTES
- Allow collections to work with std::find

ENDRELEASENOTES


tests/CMakeLists.txt
```diff
@@ -66,7 +66,7 @@ else()
   FetchContent_Declare(
     Catch2
     GIT_REPOSITORY https://github.com/catchorg/Catch2.git
-    GIT_TAG        037ddbc75cc5e58b93cf5a010a94b32333ad824d
+    GIT_TAG        a4e4e82474f894d56cbe86e13a2d23bb1b46744d
```
wdconinc, line 69: This is pretty arbitrary, other than fixing the usual MINSIGSTKSZ issue by going to a newer catch2 version.

python/templates/macros/iterator.jinja2
```diff
@@ -23,6 +43,50 @@ private:
...
 {% endmacro %}
 
+{% macro reverse_iterator_declaration(class, prefix='') %}
+{% with iterator_type = class.bare_type + prefix + 'CollectionReverseIterator' %}
+class {{ iterator_type }} {
+public:
+  using difference_type = std::ptrdiff_t;
+  using value_type = {{ class.bare_type }};
+  using pointer = {{ class.bare_type }}*;
+  using reference = {{ class.bare_type }}&;
```
tmadlener, line 53: Also here the prefix is needed, I think.
```suggestion
  using value_type = {{ prefix }}{{ class.bare_type }};
  using pointer = {{ prefix }}{{ class.bare_type }}*;
  using reference = {{ prefix }}{{ class.bare_type }}&;
```

python/templates/macros/iterator.jinja2
```diff
@@ -2,18 +2,38 @@
 {% with iterator_type = class.bare_type + prefix + 'CollectionIterator' %}
 class {{ iterator_type }} {
 public:
-  {{ iterator_type }}(size_t index, const {{ class.bare_type }}ObjPointerContainer* collection) : m_index(index), m_object(nullptr), m_collection(collection) {}
+  using difference_type = std::ptrdiff_t;
+  using value_type = {{ class.bare_type }};
+  using pointer = {{ class.bare_type }}*;
+  using reference = {{ class.bare_type }}&;
```
tmadlener, line 8: Shouldn't these be prefixed with `Mutable` for the mutable iterator?
```suggestion
  using value_type = {{ prefix }}{{ class.bare_type }};
  using pointer = {{ prefix }}{{ class.bare_type }}*;
  using reference = {{ prefix }}{{ class.bare_type }}&;
```

python/templates/Collection.cc.jinja2
```diff
@@ -1,6 +1,7 @@
 {% import "macros/utils.jinja2" as utils %}
 {% import "macros/collections.jinja2" as macros %}
 {% from "macros/iterator.jinja2" import iterator_definitions %}
+{% from "macros/iterator.jinja2" import reverse_iterator_definitions %}
```
tmadlener, line 4: I think this should also work (although I haven't tested it).
```suggestion
{% from "macros/iterator.jinja2" import iterator_definitions, reverse_iterator_definitions %}
```

tmadlener (commented):
Thanks @wdconinc. Very nice work (and it seems like you didn't have much problems with the templates).

A few "higher level" questions:
- What is currently still missing to making the iterators satisfy the forward iterator concept?
- Do we still need input iterators if we can support forward iterators? If I understand the iterator nesting correctly, than the latter is a superset of the former functionality wise, right?
- Do the unittests need I/O at the moment? Or could they be put into Catch2 test cases?
- Would it be possible to add same static checks to the unittests, that essentially just asser that our iterators fulfill the concepts that we want?

wdconinc:
Thanks for the comments. I had a `static_assert` on is_forward_iterator that was failing due to failing is_constructible, so something is still going wrong (it might have been in part because of where the `static_assert` was, since some definitions were in other code units, so it was probably bound to fail). Maybe just adding the definitions (I think it was the iterator reference/dereference and increment/decrement) into the class decaration would allow for a `static_assert`.

Otherwise, I do think we have all the bits and pieces for forward_iterators and probably even birdirectional_iterator.

I added reverse iterators, not because they are required for these concepts, but because they are typically implemented and may be useful. However, there are some issues with the `rend` sentinel since we use a `size_t` for the index. A -- beyond the end just wraps around and needs more testing.

It is true that a lot of this can go into unit tests. I was thinking that some of the memory issues would be picked up better by a full IO test, but then I realized that those are all disabled in the sanitizers so it's not quite as useful and unittests would be just fine.

We should likely add a C++20 workflow to the testing, if that is possible given the LCG upstream.

tmadlener:
> Maybe just adding the definitions (I think it was the iterator reference/dereference and increment/decrement) into the class decaration would allow for a `static_assert`.

Ah, so if I understand, you were trying to `static_assert` in the generated code directly? I would have simply done something like we do for some other type checks as well in the unittests, e.g. (there are a few more involved ones for const-correctness further down)
https://github.com/AIDASoft/podio/blob/1dc1723646bd0551b839dbfa27eb277877c73c0e/tests/unittest.cpp#L197-L198

> However, there are some issues with the rend sentinel since we use a `size_t` for the index. A -- beyond the end just wraps around and needs more testing.

That is a good point, but I think we could put `static_cast<size_t>(-1)` as a sentinel value there. I don't think we technically support collections with that size, because the `ObjectID` uses a signed integer for handling the relations during I/O, so `numeric_limits<int>::max()` is probably the largest collection we can have before things start to break. We had a bit of discussion about that in #170 if you are interested.

wdconinc:
Apparently, I think, one of the issues I'm currently running into to satisfy `static_assert(std::ranges::bidirectional_range<ExampleMCCollection>);` is the c++ standard defect report http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2325r3.html (views should not be required to be default constructible), which we do not satisfy for iterators. This has been resolved in gcc-12, but not 11.

wdconinc:
For views, we may need one more change: iterators must allow for a const dereference, which is not how we are defining them.  What do you think of a mutable `m_object`?
```c++
  {{ prefix }}{{ class.bare_type }} operator*() const {
    m_object.m_obj = (*m_collection)[m_index];
    return m_object;
  };
private:
  size_t m_index;
  mutable {{ prefix }}{{ class.bare_type }} m_object;
  const {{ class.bare_type }}ObjPointerContainer* m_collection;
```
With that change, I think

wdconinc:
The unit tests pass the more advanced tests, but fail the following ones (doesn't fill correctly, doesn't find correctly), so that needs a bit of investigation.
```c++
  // fill from const
  collection.clear();
  collection.create();
  auto const_hit = ExampleHit(0x42ULL, 0., 0., 0., 0.);
  //FIXME this should work but we may need a conversion constructor!
  //std::fill(collection.begin(), collection.end(), const_hit);
  //REQUIRE(collection.begin()->cellID() == 0x42ULL);
  // fill from mutable
  collection.clear();
  collection.create();
  auto mutable_hit = MutableExampleHit(0x42ULL, 0., 0., 0., 0.);
  std::fill(collection.begin(), collection.end(), mutable_hit);
  //FIXME
  //REQUIRE(collection.begin()->cellID() == 0x42ULL);

  // find in collection
  collection.clear();
  collection.create();
  collection.create(0x42ULL, 0., 0., 0., 0.);
  auto found = std::find_if(collection.begin(), collection.begin(),
    [](const auto& h){ return h.cellID() == 0x42ULL; });
  STATIC_REQUIRE(std::is_same_v<decltype(found), ExampleHitMutableCollectionIterator>);
  REQUIRE(found != collection.end());
  //FIXME
  //REQUIRE(found->cellID() == 0x42ULL);
```

Other outstanding questions:
- Any concerns with mutable for the `m_object` pointer in the iterator?

wdconinc:
I guess the fact that `std::fill` doesn't work is expected in the podio philosophy of not modifying objects when they are inside a container. Still confused by why `std::find` fails...

CMakeLists.txt
```diff
@@ -36,7 +36,7 @@ endif()
 # ``-DCMAKE_CXX_STANDARD=<standard>`` when invoking CMake
 set(CMAKE_CXX_STANDARD 17 CACHE STRING "")
 
-if(NOT CMAKE_CXX_STANDARD MATCHES "17|20")
+if(NOT CMAKE_CXX_STANDARD MATCHES "17|20|23")
```
tmadlener, line 39: :) I like the optimism. I have nothing against adding this here, but we are probably effectively locked to 17 until ROOT supports 20.

python/templates/Collection.h.jinja2
```diff
@@ -39,8 +45,22 @@ class {{ class.bare_type }}Collection : public podio::CollectionBase {
...
+  using const_reverse_iterator = {{ class.bare_type }}CollectionReverseIterator;
+  using reverse_iterator = {{ class.bare_type }}MutableCollectionReverseIterator;
 
   {{ class.bare_type }}Collection();
+  #if __cpp_lib_ranges
+  {{ class.bare_type }}Collection(std::ranges::input_range auto&& range)
+  requires(std::convertible_to<std::ranges::range_value_t<decltype(range)>, {{ class.bare_type }}>)
+  : m_isValid(false), m_isPrepared(false), m_isSubsetColl(false), m_collectionID(0), m_storage() {
+    if constexpr (std::ranges::sized_range<decltype(range)>) {
+      reserve(std::ranges::size(range));
```
tmadlener, line 57: Just for my understanding, this would call `Collection::reserve` in practice, right? Because that does not exist, unless it is introduced in this PR.

python/templates/macros/iterator.jinja2
```diff
@@ -2,31 +2,101 @@
...
+  {{ iterator_type }}(size_t index, const {{ class.bare_type }}ObjPointerContainer* collection)
+    : m_index(index), m_object(nullptr), m_collection(collection) {};
+
+  {{ iterator_type }}(const {{ iterator_type }}&) = default;
+  {{ iterator_type }}& operator=(const {{ iterator_type }}&) = default;
+
+  bool operator!=(const {{ iterator_type}}& x) const {
+    return m_index != x.m_index; // TODO: may not be complete
+  }
+  bool operator==(const {{ iterator_type}}& x) const {
```
tmadlener, line 70:
```suggestion
  bool operator!=(const {{ iterator_type }}& x) const {
    return m_index != x.m_index; // TODO: may not be complete
  }
  bool operator==(const {{ iterator_type }}& x) const {
```
nitpicky style comment

tmadlener:
Regarding making the `m_object` `mutable`, I think we could do without that actually since we are in any case returning a value and could in principle skip filling the internal one first. Unless I am missing something here and we actually need to update the internal state in `operator*` for other functionality. As far as I can see we don't have to do that, right? So I think the following should work:
```cpp
  {{ prefix }}{{ class.bare_type }} operator*() const {
    return {(*m_collection)[m_index])};
  };
```
(I am not entirely sure if the return like this works or if the `{{ prefix }}{{ class.bare_type }}` has to be spelled out.)

CMakeLists.txt
wdconinc: I think this may be true, to some extent. We use podio in the c++20 framework from EIC, and whatever c++20 support is in the headers is available to our code, even if only c++17 functionality is compiled into the libraries. So, I think that here in the CMakeLists.txt indeed the flag is a bit premature but it was a way to ensure the unit tests get built with c++23 for the zip views :-)

wdconinc: I guess that just means we should keep c++20 functionality in headers only.

python/templates/Collection.h.jinja2
wdconinc: Yes, I was reading through the full changeset to determine if I have tests for all new functionality, and this code branch doesn't have a test and (as you point out) will likely fail when passed a `sized_range`. My suggestion would be to simply remove the reserve and just let the usual push_back do that when needed.

tests/unittest.cpp
```diff
@@ -411,6 +420,124 @@ TEST_CASE("const correct iterators on collections", "[const-correctness]") {
                  MutableExampleCluster*>); // CollectionIterator should only give access to mutable objects
 }
 
+TEST_CASE("Iterator concepts", "[iterator-concepts]") {
+  // a collection to play with
+  auto collection = ExampleHitCollection();
+
+  #ifdef __cpp_lib_ranges
```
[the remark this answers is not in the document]
wdconinc: That's a good suggestion. I am not sure if the `__cpp_lib_ranges` else is a good enough definition of pre-c++20 but I'll run some tests. At this point I don't have something that works yet...

wdconinc:
> Regarding making the `m_object` `mutable`, I think we could do without that actually since we are in any case returning a value and could in principle skip filling the internal one first. Unless I am missing something here and we actually need to update the internal state in `operator*` for other functionality. As far as I can see we don't have to do that, right? So I think the following should work:
> 
> ```c++
>   {{ prefix }}{{ class.bare_type }} operator*() const {
>     return {(*m_collection)[m_index])};
>   };
> ```

I am not sure how strong the equality preserving constraint on dereferencing iterators is. In this case dereferencing the same (unincremented) iterator twice would result in two different objects (both pointing to the same internal data), not the same object (which we have now due to assignment). I am not sure if our implementation would treat those as equal.

I'll run some tests on this.

tmadlener:
>  In this case dereferencing the same (unincremented) iterator twice would result in two different objects (both pointing to the same internal data), not the same object (which we have now due to assignment). I am not sure if our implementation would treat those as equal.

If this falls back to using `operator==` of the returned object, than we should be OK, since that makes two objects equal to each other if they point to the same data:
https://github.com/AIDASoft/podio/blob/1dc1723646bd0551b839dbfa27eb277877c73c0e/python/templates/macros/declarations.jinja2#L90

CMakeLists.txt
wdconinc: The main issue with ROOT appears to be when doing e.g. `.include ExampleMCCollection.h` for a c++20 header in a c++17 ROOT. But if the c++20 functionality in the header is suitably guarded against non-c++20 compilers, then we should be fine.

tmadlener:
I think some of the CI failures are not related to the changes in this PR, e.g.:
https://github.com/AIDASoft/podio/runs/5754875143?check_suite_focus=true#step:4:247

They should be fixed once #253 and #254 are merged. You can try to target the `develop` branch to see how things look with them merged.

I will merge those this week, also to unclog some of the other PRs that are essentially depending on them.

tests/unittest.cpp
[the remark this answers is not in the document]
wdconinc: Finally figured this one out with iterator traits. Correct c++17 syntax (valid for c++20 too) is:
```c++
STATIC_ASSERT(std::is_same_v<std::iterator_traits<decltype(collection)::iterator>::iterator_category, std::bidirectional_iterator_tag>)
```
etc

tests/unittest.cpp
```diff
@@ -411,6 +420,130 @@ TEST_CASE("const correct iterators on collections", "[const-correctness]") {
...
+  // find in collection
+  collection.clear();
+  collection.create();
+  collection.create(0x42ULL, 0., 0., 0., 0.);
+  auto found = std::find_if(collection.begin(), collection.begin(),
+    [](const auto& h){ return h.cellID() == 0x42ULL; });
+  STATIC_REQUIRE(std::is_same_v<decltype(found), ExampleHitMutableCollectionIterator>);
+  REQUIRE(found != collection.end());
+  //FIXME
+  //REQUIRE(found->cellID() == 0x42ULL);
```
tmadlener, line 484: Is this still commented because it is not working yet, or is this leftover from previous iterations / developing?

tmadlener (commented):
Thanks for the latest additions. For the read tests we usually just throw a `std::runtime_error` if something is not as expected (at least for the time being). However, I am not entirely sure whether it is possible for all the tests in `read_iterators.cpp` or `read_views.cpp` to easily formulate conditions. In any case, since the iterators are directly using collection resources after I/O I wouldn't expect any problems there.

Overall this looks good to me, I just have one minor comment inline for one of the unittests.

tests/unittest.cpp
wdconinc: This one does still need fixing indeed. Not sure yet what's going on with this.

wdconinc:
> Thanks for the latest additions. For the read tests we usually just throw a `std::runtime_error` if something is not as expected (at least for the time being). However, I am not entirely sure whether it is possible for all the tests in `read_iterators.cpp` or `read_views.cpp` to easily formulate conditions. In any case, since the iterators are directly using collection resources after I/O I wouldn't expect any problems there.

I was going to remove those IO tests read_iterators and read_views again since they don't really test anything beyond the unit tests.

tests/unittest.cpp
wdconinc: `std::find_if` was just a stupid mistake (which took me way too long to catch) where the search boundaries were both `begin()`...

wdconinc:
The single remaining issue in the unit tests (commented out) is with the `std::fill` and is essentially the following:
```c++
for (auto i = collection.begin(); i != collection.end(); i++) {
  *i = MutableExampleHit(0x42ULL, 0., 0., 0., 0.); 
}
```
This compiles but fails to modify the collection because `operator*` returns a valid lvalue reference for the purpose of the output_iterator (and forward_iterator, bidirectional_iterator), but it is not resulting in a proper `std::fill` functionality since I think the assignment happens to a copy and not to the objects pointed to in the container. I am not sure if this is addressable (no pun intended).

It is unfortunate that `std::fill` 'seems to work' but just leaves the collection unchanged.

I think the line in the loop calls, in sequence and for the lhs only:
```c++
MutableExampleHit ExampleHitMutableCollectionIterator::operator*() const {
  return {(*m_collection)[m_index]};
}

MutableExampleHit::MutableExampleHit( ExampleHitObj* obj) : m_obj(obj) {
  if (m_obj) m_obj->acquire();
}

MutableExampleHit& MutableExampleHit::operator=(MutableExampleHit other) {
  swap(*this, other);
  return *this;
}
```

With mutable `m_object` in the iterator dereference we don't get any different results but the chain is different:
```c++
MutableExampleHit ExampleHitMutableCollectionIterator::operator*() const {
  m_object.m_obj = (*m_collection)[m_index];
  return m_object;
}

MutableExampleHit& MutableExampleHit::operator=(MutableExampleHit other) {
  swap(*this, other);
  return *this;
}
```

tmadlener:
`std::fill` seems to be a tricky one to solve indeed. I can't see an easy way to propagate the swapping of the internal pointers in the objects to be reflected in the collection. But since `std::fill` only requires a `forward_iterator`, I also don't see an easy way to make it fail at compile time.

Can we `= delete` the overloads for our generated collections? Something along the lines of
```cpp
namespace std {
template<typename T>
std::fill(CollectionIterator, CollectionIterator, const T&) = delete;
}
```
(and similar for all of the other iterators as well). I know opening `namespace std` is not exactly considered best practice, but we could at least make things fail to compile for `std::fill` instead of silently not doing the right thing.

I have played a bit with a normal `std::vector` and it seems to work there: https://godbolt.org/z/ohrP333Er

Or would you want `std::fill` to work properly?

wdconinc:
Deleting the `std::fill` is an interesting suggestion. I wonder which other STL algorithms rely on assigning to dereferenced pointers, and whether we would need to delete them all... I could imagine that a large fraction of the [modifying sequence operations](https://en.cppreference.com/w/cpp/algorithm/ranges#Modifying_sequence_operations) are likely affected.

wdconinc:
So, I guess I'm ok with std::fill failing until someone figures out how to make it work, as long as we are somehow clear about this.

tmadlener:
> I could imagine that a large fraction of the [modifying sequence operations](https://en.cppreference.com/w/cpp/algorithm/ranges#Modifying_sequence_operations) are likely affected.

Right. That is a lot to delete...

> So, I guess I'm ok with std::fill failing until someone figures out how to make it work, as long as we are somehow clear about this.

Failing in the sense of "compiling but not working" or actually deleting it to make it fail at compile time? What we can always do (at least at this point) is to properly document what is and what is not possible with our iterators. Which would essentially boil down to "everything that leaves the collection unchanged", I suppose.

wdconinc:
Maybe we should ensure that a dereferenced iterator lvalue is always a const object (even for mutable iterators!) so the assignment implied in the `std::fill` would have to fail.

tmadlener:
Then we would have come almost full circle, as that was the case up to #193 ;)

The problem I see with that is that would imply that we would have to make `operator->` also return an immutable object, in order to not have `operator*` and `operator->` have different return types. However, if `operator->` returns an immutable object, than the following no longer works.
```cpp
auto mcs = edm4hep::MCParticleCollection();
// fill with whatever
for (auto& p : mcs) {
  p.setEnergy(3.14f);
}
```
This would mean that for mutating access users would have to fall back on an indexed loop.

wdconinc:
I will leave it to you to decide how to proceed. I don't think we can resolve the `std::fill` issues without a lot(?) more discussion and I have no clear ideas about it right now (i.e. nothing else I would want to try). I think the main use case is likely interacting with read-only collections and not necessarily filling them with `std::fill`.

The only other possible loose end (as far as I think we identified) was the size_t decrement-from-zero for the reverse iterator increment operator. I have added a unit test which runs a reverse iterator loop, and that is doing the right thing, so I think this is a non-issue and we don't need to implement specific sentinel type.

tmadlener:
I would move on with what we have at the moment. That would at least make the non-mutating algorithms work, and I suppose that is the main use case. We can document the fact that `std::fill` (and siblings / cousins) do not work as intended. Could you add a small paragraph to [doc/examples.md](https://github.com/AIDASoft/podio/blob/master/doc/examples.md)? (Could also highlight some of the ranges things that are now possible, or just link to the unittests that show that)

Finally, can I ask you for two more (small) things:
- Could you merge the latest changes from master again to fix the pre-commit workflow?
- Could you put the currently commented unit test for `std::fill` into its own `TEST_CASE` and label it `[!shouldfail]`? ([Catch2 tags](https://github.com/catchorg/Catch2/blob/devel/docs/test-cases-and-sections.md#special-tags)) That would at least make sure we compile it, and we will catch it when we "accidentally" fix it at some point.

wdconinc:
No luck on the pre-commit. I can just run clang-format by hand and make the changes.

tmadlener:
Slightly late with a reply here, sorry. We had a bit of discussion about this during yesterdays [EDM4hep meeting](https://indico.cern.ch/event/1142576/). The consensus was that having things compile but not work as intended is not something that we actually want to have. One of the main arguments apart from unexpected results in user code was that it might break in unexpected places, when algorithms use some of these under the hood. In such cases it will be very challenging to debug.

I am not entirely sure how we should continue from here to be honest. I would really like to have this in podio in some form, also because there has gone quite some work into this already.

One possibility that we talked about was to "detach" this iterator issues from the podio collections (which have a few limitations due to their main purpose being I/O). E.g. having an easy way of getting a `vector` of immutable objects from the collection on which you could then do all sorts of things, but which no longer have any connection to the original collection.

I am not sure if you have any preferences or ideas in such a direction?

wdconinc:
The inefficiency of having to (essentially) deep copy a collection is probably prohibitive for the use cases that this is ultimately trying to address (e.g. parallel unsequenced loops with `std::for_each(std::execution::par_unseq, c.begin(), c.end(), my_lambda)` inside gaudi algorithms). So, having a way to explicitly create a shallow copy (which would turn the original collection IO unsafe) may be an approach.

wdconinc:
I have to point out that being able to compile code that has incorrect results is already there in `master`. This PR doesn't introduce anything new in that sense (though I agree that it may increase expectations among users that some algorithms will work when they don't, and that we should fail to compile for things that don't work). Using the current `master` branch the following fairly straightforward example compiles but fails to do what was intended:
```cpp
  auto collection = ExampleClusterCollection();
  collection.create();
  for (auto i = collection.begin(); i != collection.end(); ++i) {
    *i = MutableExampleCluster(42);
  }
  REQUIRE(collection.begin()->energy() == 42);
```

wdconinc:
Just checking in on this after encountering the motivation for its development in production code (`find_if` not working on a read-only container).

tmadlener:
`std::find_if` should technically work with an input iterator, right? So maybe lowering from a bidirectional to an input iterator could solve this without seemingly making other things work (e.g. `std::fill`) that should not work?

wdconinc:
Let me see whether I can reduce the scope to input iterator only as part of this PR.

wdconinc:
I've restricted the iterators to just input_iterators, with explicit unit tests to check they are not more than that. Due to the limitation in the iterators, the ranges cannot be used with reverse views so those have been removed from the unit tests.

tmadlener (commented):
Thanks for updating this. I have essentially just minor comments relating to tagging all related tests with one common tag and one placement of an `#if`.

From the provided unit tests it looks like the `std::fill` case still compiles, but then fails at runtime, right? I would have assumed our iterators not being forward iterators would make `std::fill` fail at compile time, but that doesn't seem to be the case. I am not sure I understand why at this point.

wdconinc:
> From the provided unit tests it looks like the std::fill case still compiles, but then fails at runtime, right?

I think this still relates to https://github.com/AIDASoft/podio/issues/281. To prevent `std::fill` on compile we would probably have to remove the Mutable iterators for regular const collections, which are the ones in master already. This PR doesn't really change that.

wdconinc:
Possible next ideas to explore:
- delete `std::fill` and all other algorithms that result in incorrect behavior,
- delete the underlying assignment operator that operates on dereferenced pointer lvalues.

wdconinc:
I'm thinking the following is what we may need in `python/templates/macros/declarations.jinja2`:
```cpp
{% if prefix == 'Mutable' %}
  /// copy-assignment operator
  {{  full_type }}& operator=({{ type }} other);
  {{  full_type }}& operator=({{ full_type }} other);
{% else %}
  /// copy-assignment operator deleted (use Mutable type)
  {{  full_type }}& operator=({{ full_type }} other) = delete;
{% endif %}
```

This would remove assignments like `object1 = object2`, but allow `mutableobject1 = mutableobject2` and `mutableobject1 = object2`. Implementation of this highlights a few other false-positives in the unittests (`tests/read_test.h` where we reuse the `p` in `auto p = mcps[0]` which is an Object, not MutableObject). Ultimately this would also remove this assignment in `std::fill` or the equivalent assignment in
```cpp
auto i = collection.begin();
*i = MutableExampleHit;
```
since const-correctness of `i` and `begin()` would block assignment.

Ultimately this should mean that `std::fill` will work on MutableObjectCollection, but would fail at compilation on ObjectCollection.

wdconinc:
Nevermind, removing assignment violates [`std::assignable_from<T&, T>`](http://en.cppreference.com/w/cpp/concepts/assignable_from) which is required by [`std::input_or_output_iterator`](https://en.cppreference.com/w/cpp/iterator/input_or_output_iterator), so removing assignment is not the right approach (though we are allowed to make it fail).

wdconinc:
Other things I tried today that didn't pan out:
- delete begin and force use of cbegin with as_const (cbegin is required to be what begin returns on const),
- throw exceptions in problematic code (too many false positives, and didn't even block std::fill).

tmadlener:
I just tried to explicitly generate `delete`d overloads for `std::fill` for all of the iterators and that seems to do the trick and calls to `std::fill` no longer compile.

[changes on top of this PR](https://github.com/wdconinc/podio/compare/input-forward-iterators...tmadlener:podio:no-compile-fill?expand=1)

There is also some additional work in there which would in principle allow us to do `static_assert`s to make sure that calling `std::fill` really doesn't work. It is pretty esoteric, and I essentially just wanted to see if I can get something like that to work, so we could just pull in parts that generate the `delete`s for the overloads.

wdconinc:
Do we actually need a mutable iterator on collections? Can we just make it entirely non-mutable?

Right now that will fail the following constructs:
```cpp
  for (auto&& i : coll) {
    i.energy(42); // make sure that we can indeed change the energy here for
                  // non-const collections
  }
  REQUIRE(hit1.energy() == 42);
```
Are these used in the wild? It only fails at the `unittest` stage right now, with these specific cases. Does it make sense to force users to treat collections as always const?

tmadlener:
> Are these used in the wild?

Good question, I don't know. We could at probably flush out most of the use cases with a deprecation warning to see how many there are.

On the other hand, I am not entirely sure if I like the idea of removing the iterators, since operations done through the iterators are working, but assignments through dereferenced iterators are failing.

hegner:
@tmadlener - should we resurrect this?

tmadlener:
I am not entirely sure how easy this is to resurrect, given that doing it “properly” would probably mean quite a bit of work. I will have to look at the discussion in here again and see what the current status is. Maybe there are things that can be salvaged for now and we come back to proper ranges support at a latter point

hegner:
OK. Thanks

PR #276, opened 2022-03-29
Remove support for having std::string members in datatypes and components

### LINKED ISSUE #267: std::string should be removed from the list of allowed builtin types
Currently it is possible to create datatypes (and components) that have a `std::string` member, because we explicitly consider it as a builtin type:
https://github.com/AIDASoft/podio/blob/d00f8cf8c76d5e405bb8bc7afae3ffa8a82b8dfd/python/generator_utils.py#L43-L46

Are there any actual users of this feature, other than the internal test data model? I think this should be removed, so that we truly have PODs only.

tmadlener:
BEGINRELEASENOTES
- Remove support for having std::string members in datatypes and components, as they break PODness and it seems that this feature was not in use in any case.
- Make ROOTReader slightly more robust against missing datatypes in dictionaries when reading files.

ENDRELEASENOTES

Fixes #267 
- [x] Targeting `develop` for now, for cleaner diff.

### DIFF
--- a/python/generator_utils.py
+++ b/python/generator_utils.py
@@ -41,14 +41,13 @@ class DefinitionError(Exception):
 BUILTIN_TYPES = ["int", "long", "float", "double",
                  "unsigned int", "unsigned", "unsigned long",
                  "char", "short", "bool", "long long",
-                 "unsigned long long", "std::string"]
+                 "unsigned long long"]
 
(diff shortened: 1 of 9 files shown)


tmadlener:
Rebased to target `master` after recent merges.

PR #278, opened 2022-03-30
Frame in memory interface

tmadlener:
BEGINRELEASENOTES
- Introduce the `podio::Frame` as a generalized (event) data container with the necessary prerequisites to extend this basic implementation with all the planned functionality later.
  - This version does not yet include any I/O functionality but only introduces the thread-safe in-memory interface.

ENDRELEASENOTES

This is a first version of the in-memory interface of the `Frame` proposal. I am putting this here to gather early feedback, mainly on the user interface and on some of the proposed functionality. At this point only the basic skeleton for the `podio::Frame` is present and much of the foreseen functionality is still missing. However, the interface should be somewhat stable already as most of that functionality should be hidden, or not observable through the interface. One thing that could still influence the interface could be the choice of policies that we want to offer. I have put some of my proposals below.

Having this discussion early enough might make it possible to split this feature into several pull requests instead of having one extremely large one.

### `podio::Frame` interface

The Frame is implemented via type-erasure. This allows us to have different policies but only one user facing class with value semantics and the possibility to construct from different raw data types [^1]. The current public interface of the Frame looks essentially like this:
```cpp
class Frame {
public:
  template<typename CollT>
  const CollT& get(const std::string& name) const;

  template<typename CollT>
  const CollT& put(CollT&& coll, const std::string& name);

  template<typename T>
  void putParameter(const std::string& key, T value);
  
  template<typename T>
  const T& getParameter(const std::string& key) const;
};
```

Some notes (about the actual implementation that are not visible in the above version):
- The `put` and `get` methods use `std::enable_if` to only be available for collections added by podio. 
- The `put` method uses `std::enable_if` to force users to explicitly pass a temporary or relinquish ownership via `std::move`.
- The `{put|get}Parameter` functionality provides the extra data / meta data functionality and is using the `GenericParameters` class under the hood.
  - The templates are using `std::enable_if` to only be available for the types that are actually supported (`int`, `float`, `std::string` and `std::vectors` of those)
  - The return value of `getParameter` is either a `const T&` or a `T` (for `int` and `float`).
  - There are a few dedicated overloads for `putParameter` to enhance the usability, e.g. for passing raw string literals or initializer lists.

### Planned policies
There are a few policies that I have already in mind, but this is still open for discussion
- Unpacking policy that controls how the raw data that are passed are unpacked. E.g. could be eagerly on Frame construction or lazily on demand. Would be completely invisible for users.
- Handling of collisions when putting collections into the Frame. E.g. could throw an exception, silently drop the new collection or replace the existing collection. Could potentially also be part of the `put` interface to have more fine grained control, instead of having a "global" policy
- Handling of missing collections. E.g. could throw an exception or simply return an empty new collection. Could potentially become part of the `get` interface to have more fine grained control, instead of a "global" policy.
- Locking policy. E.g. could drop all (externally observable) locking if the Frame is only used from one thread or to read only, or could go with the default locking which might have an impact on single threaded use.

[^1]: This first proposal is still missing these features at the moment.

### Preliminary TODO list
- [x] Policies -> **postponed to later PR**
- [x] I/O functionality -> Done in #287 
- [x] Thread safe get/set functionality
- [x] Thread safe parameters access
- [x] documentation
- [x] Combined c++20 concepts and c++17 compatible implementation of type constraints? Not sure how easily possible that is. Probably can be added later. -> **postponed to later PR**
- [x] Release notes
- [x] ~~Targets develop~~
- [x] Need #262 ~on top of develop~ [diff of once that is merged](https://github.com/tmadlener/podio/compare/templated-generic-param-access...tmadlener:frame?expand=1)

### DIFF
--- a/tests/frame.cpp
+++ b/tests/frame.cpp
@@ -227,7 +227,8 @@ TEST_CASE("Frame movability", "[frame][move-semantics]") {
   }
 
   SECTION("Use after move construction") {
-    auto otherFrame = std::move(frame);
+    auto otherFrame = std::move(frame); // NOLINT(clang-analyzer-cplusplus.Move) clang-tidy and the Catch2 sections
+                                        // setup do not go along here
     otherFrame.putParameter("aString", "Can add strings after move-constructing");
     REQUIRE(otherFrame.getParameter<std::string>("aString") == "Can add strings after move-constructing");
 
--- a/tests/frame.cpp
+++ b/tests/frame.cpp
@@ -3,6 +3,7 @@
 #include "catch2/catch_test_macros.hpp"
 
 #include "datamodel/ExampleClusterCollection.h"
+#include "datamodel/ExampleHitCollection.h"
 
 TEST_CASE("Frame collections", "[frame][basics]") {
   auto event = podio::Frame();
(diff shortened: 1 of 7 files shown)


tests/frame.cpp
```diff
@@ -0,0 +1,333 @@
...
+  for (int i = 0; i < nThreads; ++i) {
+    threads.emplace_back([&frame, i]() {
+      auto clusters = ExampleClusterCollection();
+      clusters.create(i * 3.14);
+      clusters.create(i * 3.14);
+      frame.put(std::move(clusters), makeName("clusters", i));
+
+      // Retrieve a few collections in between and do iust a very basic testing
+      auto& existingClu = frame.get<ExampleClusterCollection>("clusters");
+      REQUIRE(existingClu.size() == 2);
```
tmadlener, line 148: catch2 macros are not yet threadafe:
https://github.com/catchorg/Catch2/blob/devel/docs/limitations.md#thread-safe-assertions

tmadlener:
Closing this, as it is fully contained in #287 and will be merged as an “atomic” commit introducing the Frame

PR #279, opened 2022-03-31
Unset ROOT_INCLUDE_PATH for tests

tmadlener:
BEGINRELEASENOTES
- Fix test environment to work again in newest Key4hep release by unsetting `ROOT_INCLUDE_PATH` in the test environment to avoid potential clashes with existing other installations in the environment.
- Add CI build against the Key4hep nightlies.
- Switch to use the Catch2 installation from Key4hep for the workflows.

ENDRELEASENOTES

See key4hep/key4hep-spack#336 for some related discussion

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -22,6 +22,9 @@ function(CREATE_PODIO_TEST sourcefile additional_libs)
   target_link_libraries(${name} TestDataModel ${additional_libs})
   set_property(TEST ${name} PROPERTY ENVIRONMENT
       LD_LIBRARY_PATH=${CMAKE_CURRENT_BINARY_DIR}:${CMAKE_BINARY_DIR}/src:$ENV{LD_LIBRARY_PATH}
+      # Clear the ROOT_INCLUDE_PATH for the tests, to avoid potential conflicts
+      # with existing headers from other installations
+      ROOT_INCLUDE_PATH=
     )
 endfunction()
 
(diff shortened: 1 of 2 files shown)


vvolkl:
Looks good, though I'm still puzzled why this was triggered by the formatting changes ...

tmadlener:
Some of the clang-tidy fixes are not just "simple" formatting fixes but also contain some fixes along the lines of
```cpp
// before
virtual ~IReader() {}
// after
virtual ~IReader() = default;
```
Some of the include guards change as well in that commit, so that might also play into this.

And I am not entirely sure how strict ROOT actually is with these dictionary payloads, i.e. if they have to be identical character by character or if (actual) formatting changes would be allowed.

andresailer:
The header guards were changed by clang-format, that might cause root to parse effectively the same header twice.

tests/CMakeLists.txt
```diff
@@ -22,6 +22,9 @@ function(CREATE_PODIO_TEST sourcefile additional_libs)
   target_link_libraries(${name} TestDataModel ${additional_libs})
   set_property(TEST ${name} PROPERTY ENVIRONMENT
       LD_LIBRARY_PATH=${CMAKE_CURRENT_BINARY_DIR}:${CMAKE_BINARY_DIR}/src:$ENV{LD_LIBRARY_PATH}
+      # Clear the ROOT_INCLUDE_PATH for the tests, to avoid potential conflicts
+      # with existing headers from other installations
+      ROOT_INCLUDE_PATH=
```
vvolkl, line 27: Do you actually have to clear it or will it be enough to prefix it with the correct path?

tmadlener: It would in principle work with prefixing `${CMAKE_SOURCE_DIR}/tests/datamodel`. But my thought was that clearing it gives a simpler test environment in the end, and this would be one less thing to worry about.

tmadlener:
I also just realized that at the moment that this wouldn't exhibit any problems as on master things are still working as we do not yet have the clang-tidy changes.

tmadlener:
Unless there are any further requests, I would merge this and then #254 today.

PR #282, opened 2022-04-20
Add option to toggle the formatting via clang-format in the generator cmake macro

tmadlener:
BEGINRELEASENOTES
- Add a `PODIO_USE_CLANG_FORMAT` option to the cmake configuration to toggle the autodiscovery of `clang-format` and a `.clang-format` configuration file. This option is also available for downstream packages that use podio to generate their EDM. If set to `OFF` podio will not try to see whether `clang-format` and a `.clang-format` file are available and will also not try to format the code accordingly. By default this option is `ON`, but this gives users the possibility to turn the feature off in case of problems.

ENDRELEASENOTES

The main reason to introduce this override is the following: `--fallback-style` first appears with LLVM 4.0.0, but the default version on CentOS7 appears to be 3.4.2. #254 introduced the automatic detection of `clang-format` and the presence of a `.clang-format` file in the cmake macros that are used for the code generation to switch on the automatic formatting during code generation. This now breaks on CentOS7 where the default LLVM is installed. Having this option makes it possible to simply skip the auto discovery and subsequently the formatting.


cmake/podioMacros.cmake
```diff
@@ -129,21 +129,27 @@ endfunction()
 # this is essentially a no-op, and should not cause re-compilation.
 #---------------------------------------------------------------------------------------------------
 function(PODIO_GENERATE_DATAMODEL datamodel YAML_FILE RETURN_HEADERS RETURN_SOURCES)
-  CMAKE_PARSE_ARGUMENTS(ARG "" "OUTPUT_FOLDER" "IO_BACKEND_HANDLERS" ${ARGN})
+  CMAKE_PARSE_ARGUMENTS(ARG "" "OUTPUT_FOLDER;FORMAT" "IO_BACKEND_HANDLERS" ${ARGN})
```
andresailer, line 132: Add `FORMAT` to the docstring before this function

cmake/podioMacros.cmake
```diff
@@ -129,21 +129,27 @@ endfunction()
...
   IF(NOT ARG_OUTPUT_FOLDER)
     SET(ARG_OUTPUT_FOLDER ${CMAKE_CURRENT_SOURCE_DIR})
   ENDIF()
   IF(NOT ARG_IO_BACKEND_HANDLERS)
     # At least build the ROOT selection.xml by default for now
     SET(ARG_IO_BACKEND_HANDLERS "ROOT")
   ENDIF()
+  IF ("${ARG_FORMAT}" STREQUAL "")
```
andresailer, line 140: Not giving an argument to FORMAT will produce an error in cmake, won't it?

tmadlener: Just checked with cmake@3.16.3 (Ubuntu) and all of the following calls seem to work as intended

- Passing in the value of a variable (passing a "false-y" value here is the only way to turn disable auto formatting)
```cmake
PODIO_GENERATE_DATAMODEL(datamodel datalayout.yaml headers sources
  FORMAT ${FORMAT_GENCODE} IO_BACKEND_HANDLERS ${PODIO_IO_HANDLERS})
```

- Simply passing `FORMAT` without any additional argument (will fall back to the default `ON`)
```cmake
PODIO_GENERATE_DATAMODEL(datamodel datalayout.yaml headers sources
  FORMAT IO_BACKEND_HANDLERS ${PODIO_IO_HANDLERS})
```

vvolkl:
So any systems with the wrong `clang-format` versions have to explicitly disable this feature in order for the datamodel to build? Is it possible to query the version of `clang-format` and fix its invocation accordingly?

tmadlener:
In principle yes. This PR essentially gives the opportunity to do that without too much difficulty. The problem is that apart from the flags that are available for `clang-format` also the output of `clang-format --version` has changed, so parsing the version from that output becomes slightly cumbersome as well. If we want to go down that route, I think it might be easier to just try and run `clang-format` first with the detected `.clang-format` config and see if that works (that could be done inside the code generator).
 
In the end this should really only be a problem for rather old versions of clang-format and fairly recent versions should be fine. One other option would be to turn the defaults around, so that users have to explicitly opt in to the whole thing.

cmake/podioMacros.cmake
andresailer: Ok, thanks for checking!

tmadlener:
@mirguest, @vvolkl updated this to a simpler approach of introducing a `PODIO_USE_CLANG_FORMAT` option that is also available in downstream packages because it gets imported there via the cmake macros.

Still not fully automatic detection of `clang-format` capabilities, but I think the easiest we can do at the moment.

tmadlener:
It looks like that on newer Ubuntu versions `clang-format` sets up proper symlinks to be invokable via `clang-format`. Unfortunately parsing the version from `clang-format --version` also requires quite a bit of work, and then we still have the problem that it is not yet entirely clear to me which versions brought which capabilities. I would suggest to merge this as is and implement a more elegant solution later if necessary.

tmadlener:
> The check for the clang-format version is fairly straightforward, I just check the output of `clang-format --fallback-style`

Looks good to me. The only thing we will not catch in this case is if features requested in the `.clang-format` file are not yet supported by `clang-format`. But in that case I think it is a bit more forgiving and just issue warnings.

PR #283, opened 2022-04-20
Make it possible to define default initialization values in the yaml definition

### LINKED ISSUE #266: Allow setting default initializers on member definition

tmadlener:
BEGINRELEASENOTES
- Allow users to define default values for member variables, instead of default initializing all of them.
  - The syntax for specifying a default value is `- <type> <name>{<init-value>} // <description>`.
  - The passed value is not validated in any way. **Apart from a very basic syntax check, there is no validation that the provided default initialization values are actually valid**. This means that generated code might not compile.
- Remove some of the python2 compatibility and do some cleanup 

ENDRELEASENOTES

Supersedes #268 
Fixes #266 

It would be possible to run the individual members through a syntax-only check of the compiler to see if the initialization is actually valid. This would catch probably almost all issues before we generate any code. However, this has proven to increase generation time quite significantly (see https://github.com/AIDASoft/podio/pull/268#issuecomment-1076425296). Since invalid default values will in any case be caught by the compiler after code generation, we decided to skip validation of default values and keep this as a sort of "expert feature".

A few of the things that are not valid but pass validation have been put into the tests as examples (and documentation).

### DIFF
--- a/doc/datamodel_syntax.md
+++ b/doc/datamodel_syntax.md
@@ -75,7 +75,16 @@ The definition of a member is done in the `Members` section in the form:
       <type> <name> // <comment>
 ```
 
-where `type` can be any buildin-type or a `component`.
+where `type` can be any builtin-type or a `component`.
+
+It is also possible to specify default values for members via
+```yaml
+    Members:
+      <type> <name>{<default-value>} // <comment>
+```
+Note that in this case it is extremely expensive to check whether the provided `default-value` results in valid c++.
+Hence, there is only a very basic syntax check, but no actual type check, and wrong default values will be caught only when trying to compile the datamodel.
+
 
 ### Definition of references between objects:
 There can be one-to-one-relations and one-to-many relations being stored in a particular class. This happens either in the `OneToOneRelations` or `OneToManyRelations` section of the data definition. The definition has again the form:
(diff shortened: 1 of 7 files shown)


tmadlener:
@wdconinc would this cover all your intended use cases?

PR #286, opened 2022-05-23
Make sure vector member buffers point to the correct place even if a collection has been moved

tmadlener:
BEGINRELEASENOTES
- Make sure that vector member buffers for writing point to the correct place even if a collection has been moved, by resetting them when the buffers are requested.
- Add checks for this to the unittests, as this is sort of an interface for I/O backends.

ENDRELEASENOTES


tests/unittest.cpp
```diff
@@ -524,9 +526,9 @@ auto createCollections(const size_t nElements = 3u) {
 }
 
 // Helper functionality to keep the tests below with a common setup a bit shorter
-void checkCollections(const ExampleHitCollection& hits, const ExampleClusterCollection& clusters,
-                      const ExampleWithVectorMemberCollection& vectors,
-                      const podio::UserDataCollection<float>& userData, const size_t nElements = 3u) {
+void checkCollections(/*const*/ ExampleHitCollection& hits, /*const*/ ExampleClusterCollection& clusters,
+                      /*const*/ ExampleWithVectorMemberCollection& vectors,
+                      /*const*/ podio::UserDataCollection<float>& userData, /*const*/ size_t nElements = 3u) {
```
tmadlener, line 531: These should all be `const`, but currently `prepareForWrite` is not thread-safe and thus not marked `const`.

tests/unittest.cpp
```diff
@@ -543,14 +545,74 @@ void checkCollections(const ExampleHitCollection& hits, const ExampleClusterColl
...
+      const auto& hitPOD = (*hitPODBuffers)[iHit];
+      REQUIRE(hitPOD.cellID == static_cast<unsigned long long>(iHit));
+      REQUIRE(hitPOD.energy == 100.f * iHit);
+    }
+  } else {
+    // NOTE: Just basic tests here really. Checking the contents as well here
+    // requires information that is not easily available at the moment
+    REQUIRE(hitBuffers.data == nullptr);
+    REQUIRE(hitBuffers.references->size() == 1);
+    REQUIRE(hitBuffers.vectorMembers->empty());
```
tmadlener, line 573: This is always true for subset collections and already checked in the `"Subset collection basics"` test case.

tests/unittest.cpp
```diff
@@ -543,14 +545,74 @@ void checkCollections(const ExampleHitCollection& hits, const ExampleClusterColl
...
+      const auto& hitID = (*hitRelationBuffer)[iHit];
+      REQUIRE(hitID.index == (int)iHit);
+      REQUIRE(static_cast<unsigned>(hitID.collectionID) == hits.getID());
+    }
+  } else {
+    // NOTE: Just basic tests here really. Checking the contents as well here
+    // requires information that is not easily available at the moment
+    REQUIRE(clusterBuffers.data == nullptr);
+    REQUIRE(clusterBuffers.references->size() == 1);
+    REQUIRE(clusterBuffers.vectorMembers->empty());
```
tmadlener, line 594: Always true for subset collections; see above.

tests/unittest.cpp
```diff
@@ -543,14 +545,74 @@ void checkCollections(const ExampleHitCollection& hits, const ExampleClusterColl
...
+    for (int iC = 0; iC < (int)nElements; ++iC) {
+      REQUIRE((*countBuffer)[iC * 2] == iC);
+      REQUIRE((*countBuffer)[iC * 2 + 1] == 42 + iC);
+    }
+  } else {
+    // NOTE: Just basic tests here really. Checking the contents as well here
+    // requires information that is not easily available at the moment
+    REQUIRE(vecMemBuffers.data == nullptr);
+    REQUIRE(vecMemBuffers.references->size() == 1);
+    REQUIRE(vecMemBuffers.vectorMembers->empty());
```
tmadlener, line 614: Same here.

tests/unittest.cpp
tmadlener: Once #295 is merged, these should be possible as `const` again.

tmadlener: `getBuffers` is still not marked `const` preventing this. Fixing that in a separate PR.

tmadlener: Tracked in #302

PR #287, opened 2022-05-23
Prototype version of Frame I/O

tmadlener:
BEGINRELEASENOTES
- Introduce the `podio::Frame` as a generalized, thread-safe (event) data container.
  - This first version offers all necessary functionality and an almost finalized interface, i.e. we plan to keep this as stable as possible, but we might still change things if it turns out that there are better ways to do some things
  - For details about the basic interface and the underlying design considerations please consult the corresponding [documentation](https://github.com/AIDASoft/podio/doc/frame.md)
- **This will be the only way to work with podio data starting from version 1.0**
  - For now the current I/O implementations remain in place unchanged, but they will be deprecated (and removed) in the not too distant future

ENDRELEASENOTES

This is a prototype version of Frame based I/O with the necessary changes to the `Frame` and the collections that are necessary to enable this. **In its current form all existing I/O functionality remains intact.** This current version supports the following:
- Writing and reading via ROOT. **The resulting files are different than the ones written by the `ROOTWriter`**.
  - The reading side should already be able to handle different Frame categories ~, while the writing part mainly has all the preparations for, but not actual support yet.~
- Writing and reading via SIO. **The resulting files are different than the ones written by the `SIOWriter`**.
  - For SIO the written Frames are *self contained* in the sense that they store all the necessary information for each frame which makes the writer/reader almost stateless.

From a usage perspective the `Frame` behaves very similar to the current `EventStore` for reading purposes. This can be seen by the few changes that have been necessary to `read_test.h` to accommodate for the simultaneous usage with both. The major differences between the two at this point are:
- Interface to get the *parameters* (*event meta data* in the `EventStore`).
- No direct support yet for *collection meta data* in the `Frame`.

For writing there are some conceptual differences that make a common implementation of the test code rather hard:
- The `EventStore` hands out a reference to the collection and makes sure to clear it for every event
- The `Frame` starts empty in every event and the collections are newly created and added to the `Frame`

### Some more technicalities
Quite a few changes had to be introduced to make everything work, the most important ones are:
- Make it possible to instantiate collections from buffers and to create collections "from thin air". In the current implementation this is done by introducing `CollectionReadBuffers` that have a `createCollection` method that is dynamically assigned during runtime (and the necessary code is generated by code generation). This is necessary to separate the reading of the buffers (and their potential schema evolution) from the actual creation of the collections later.
  - This also includes the possibility to create "empty" CollectionBuffers that can then be filled by I/O backends when reading. 
- The new root file format will contain a `TTree` for every `Frame` category. Additionally, there is one meta data tree that holds all the information that is necessary for reading everything.
  - For ROOT the contents of any given category is fixed by the contents of the first element of a category that is written, as that is used to initialize the `TTree` and all the branches.
- There are quite a few "quick and dirty" solutions to get a working prototype ready:
  - ~Collections still have to be *registered for write* in the `ROOTFrameReader`. It is not yet entirely clear to me how the best interface looks for this.~ In the latest iteration the *collections to write* have become an argument to `writeFrame` (with a default to write all available collections).
  - The current implementation of the buffer and collection creation could probably be implemented slightly cleaner
  - The `Frame` has quite some additional (public) API surface at the moment for the writer to get access to the necessary data.

### TODOs
- [x] tests for multithreaded use of Frames created from file -> postpone to later PR (this needs quite some additional "setup work", that would mainly pollute this PR even further.
- [x] SIO writing
- [x] Writing different categories
- [x] Test how interface works in the "real world" (aka k4SimDelphes, see key4hep/k4SimDelphes#85)
- [x] Unpacking policies. The current implementation only has lazy unpacking -> policies are foreseen in the current design but will be added later
- [x] release notes
- [x] documentation
- [x] builds on top of #278
- [x] Includes changes from #285, #286 #294, #295


include/podio/SIOFrameReader.h
```diff
@@ -0,0 +1,54 @@
...
+namespace podio {
+
+class CollectionIDTable;
+
+class SIOFrameReader {
+
+public:
+  SIOFrameReader();
+  ~SIOFrameReader() = default;
+
```
tmadlener, line 23: Delete copy ctors.

include/podio/SIORawData.h
```diff
@@ -0,0 +1,79 @@
...
+#include <memory>
+#include <numeric>
+#include <optional>
+#include <string>
+#include <vector>
+
+namespace podio {
+class SIORawData {
+  // using CollIDPtr = std::shared_ptr<const podio::CollectionIDTable>;
+
```
tmadlener, line 21:
```suggestion
```

include/podio/ROOTFrameReader.h
```diff
@@ -0,0 +1,97 @@
...
+   */
+  struct CategoryInfo {
+    /// constructor from chain for more convenient map instertion
+    CategoryInfo(std::unique_ptr<TChain>&& c) : chain(std::move(c)) {
+    }
+    std::unique_ptr<TChain> chain{nullptr};                              ///< The TChain with the data
+    unsigned entry{0};                                                   ///< The next entry to read
+    std::vector<std::pair<std::string, CollectionInfo>> storedClasses{}; ///< The stored collections in this category
+    std::vector<root_utils::CollectionBranches> branches{};              ///< The branches for this category
+    std::shared_ptr<CollectionIDTable> table{nullptr};                   ///< The collection ID table for this category
```
tmadlener, line 77:
```suggestion
    std::shared_ptr<const CollectionIDTable> table{nullptr};                   ///< The collection ID table for this category
```

This is `const` in the `ROOTRawData` so it should also be `const` here.

include/podio/ROOTReader.h
```diff
@@ -26,6 +25,8 @@ class CollectionBase;
 class Registry;
 class CollectionIDTable;
 class GenericParameters;
+struct CollectionBuffers;
+
```
tmadlener, line 29: This does no longer exist? Should probably be `CollectionReadBuffers`. Is the forward declaration necessary?

include/podio/ROOTFrameReader.h
```diff
@@ -0,0 +1,97 @@
...
+  // non-copyable
+  ROOTFrameReader(const ROOTFrameReader&) = delete;
+  ROOTFrameReader& operator=(const ROOTFrameReader&) = delete;
+
+  void openFile(const std::string& filename);
+
+  void openFiles(const std::vector<std::string>& filenames);
+
+  /// Read all collections requested
+  std::unique_ptr<podio::ROOTRawData> readNextFrame(const std::string& category);
```
hegner, line 50: what behaviour do you want to have in case of end of iteration? The doc string should reflect that

include/podio/ROOTFrameReader.h
```diff
@@ -0,0 +1,97 @@
...
+
+  /// Returns number of entries for the given category
+  unsigned getEntries(const std::string& category) const;
```
hegner, line 53: the term category is a bit jargon. You should explain a bit better what that is/

include/podio/ROOTFrameReader.h
```diff
@@ -0,0 +1,97 @@
...
+
+  podio::version::Version currentFileVersion() const {
+    return m_fileVersion;
+  }
+
+  // Information about the data vector as wall as the collection class type
+  // and the index in the collection branches cache vector
+  // TODO: Make this private again
```
hegner, line 61: make it private ;-)

include/podio/ROOTFrameReader.h
```diff
@@ -0,0 +1,97 @@
...
+  using CollectionInfo = std::tuple<const TClass*, const TClass*, size_t>;
+
+private:
+  /**
+   * Helper struct to group together all the necessary state to read / process a
+   * given category.
+   */
+  struct CategoryInfo {
```
hegner, line 69: again - explain "category"

include/podio/ROOTFrameReader.h
```diff
@@ -0,0 +1,97 @@
...
+    /// constructor from chain for more convenient map instertion
```
hegner, line 70: typo

include/podio/SIOFrameWriter.h
```diff
@@ -0,0 +1,41 @@
...
+
+#include <string>
+#include <utility>
+#include <vector>
+
+namespace podio {
+
+class Frame;
+
+class SIOFrameWriter {
```
hegner, line 16: would there be a use case to introduce an AbstractFrameWriter ?

src/rootUtils.h
```diff
@@ -16,13 +17,54 @@
 #include <vector>
 
 namespace podio::root_utils {
+/**
+ * The name of the meta data tree in podio ROOT files. This tree mainly stores
+ * meta data that is necessary for ROOT based I/O.
+ */
+constexpr static auto metaTreeName = "podio_metadata";
```
hegner, line 24: not sure such information is ROOT specific. It is part of the file schema and could maybe go elsewhere

hegner (changes requested):
Looks very good. I only have minor comments so far.
What we should make sure is to have a decent internal documentation about the concepts there.

Could you work on the TODOs listed in the PR? Namely some docs and release notes? The unpacking we can put into another ticket. Thanks!

include/podio/ROOTFrameReader.h
tmadlener: Turns out it is not entirely trivial to make it private, since it is used by a free function in the implementation. Moved it into a `detail` namespace now to at least "hide" it somewhat, since it is an implementation detail in the end.

include/podio/SIORawData.h
```diff
@@ -0,0 +1,79 @@
+#ifndef PODIO_SIORAWDATA_H
```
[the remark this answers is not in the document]
tmadlener: Just to have it in writing as well: Each backend defines it's RawData type and is also responsible for implementing the necessary interface for the interaction with the Frame. The design philosophy essentially was to separate reading the data from file and doing operations on that file later on. For the SIO backend this enables simply reading compressed record blocks first and then doing all the necessary unpacking / decompressing later (and on a different thread if necessary).

include/podio/ROOTFrameReader.h
```diff
@@ -0,0 +1,120 @@
...
+   */
+  std::unique_ptr<podio::ROOTFrameData> readNextEntry(const std::string& name);
+
+  /// Returns number of entries for the given name
+  unsigned getEntries(const std::string& name) const;
+
+  podio::version::Version currentFileVersion() const {
+    return m_fileVersion;
+  }
+
```
forthommel, line 69:
```suggestion
/// Returns stored collections in this category
std::map<std::string, detail::CollectionInfo> storedClasses(const std::string& catName);
```

src/ROOTFrameReader.cc
```diff
@@ -0,0 +1,275 @@
...
+      brNames.emplace_back(name.substr(0, fUnder));
+    }
+  }
+
+  std::sort(brNames.begin(), brNames.end());
+  brNames.erase(std::unique(brNames.begin(), brNames.end()), brNames.end());
+
+  return brNames;
+}
+
```
forthommel, line 184:
```suggestion

std::map<std::string, detail::CollectionInfo> ROOTFrameReader::storedClasses(const std::string& catName) {
  std::map<std::string, detail::CollectionInfo> colls_info;
  if (auto& catInfo = getCategoryInfo(catName); catInfo.table) {
    for (const auto& [col, col_info] : catInfo.storedClasses)
      colls_info[col] = col_info;
  }
  return colls_info;
}

```

forthommel: @tmadlener Very nice work around these frames! May I suggest this addition to ease their integration into consuming workflows? This allows the access to the `TClass` objects as retrieved from the `TChain` on readout, and thus allows retrieving a human- (and `RDataSource`-) readable version of each collection type.

forthommel (changes requested):

src/ROOTFrameReader.cc
tmadlener: Hi @forthommel,

The `detail::CollectionInfo` are currently really an implementation detail of the `ROOTFrameReader` and I am not entirely sure if they should be exposed in an interface in their current form. Definitely not while they are still in the `detail` namespace. What is the information that is actually necessary for `RDataSource` to work nicely here? I would prefer a solution that only exposes what is really necessary instead of spilling everything here.

forthommel: Hi @tmadlener, and thanks for the feedback!
What my scheme would need is the collection/category type with name `catName`, as stored in the `TChain`. Unless it is already exposed elsewhere in the API and I missed it?

tmadlener: There is nothing in the API yet that exposes any of those details, as the currently foreseen EDM API does not need anything like this. Additionally, it gives podio the freedom to treat I/O implementations as a detail instead of making e.g. the Branch layout in the root files part of the API.

IIUC, then you would need the XYZData (resp. vectors of them), right?

I would suggest to move that discussion/implementation to another PR entirely, because this one is (too) large already and we would actually like to keep this one in its current form to make it possible to implement Schema evolution on top of this before iterating further.

tmadlener:
@hegner as discussed briefly today:
- Added documentation and some release notes (the link in the release notes will become valid once the PR is actually merged)
- Rebased this onto the latest master and fix some minor issues with CMake to make sure everything compiles again (these were introduced in #323)

As far as a first version goes this is complete and can be reviewed, resp. considered stable for schema evolution purposes.

hegner (approved):

hegner (approved):

PR #291, opened 2022-05-28
Replace obj_needs_destructor by more general concept

### LINKED ISSUE #288: Replace `obj_needs_destructor` by more general concept
Currently we only generate a destructor for the `Obj` objects if it is necessary, e.g. here:

https://github.com/AIDASoft/podio/blob/469b739c855841ed7821d2a43e3f8c779a471e02/python/templates/Obj.h.jinja2#L35-L39

which is populated in the generator, here:

https://github.com/AIDASoft/podio/blob/469b739c855841ed7821d2a43e3f8c779a471e02/python/podio_class_generator.py#L259-L260

The condition in the generator describes a rather general class of datatypes, i.e. ones that have no relations or vector members. I think the name of `obj_needs_destructor` should be changed to something that better reflects this more general condition. E.g. `is_trivial_type` (or something similar).

soumilbaldota:
BEGINRELEASENOTES

- The name of obj_needs_destructor should be changed to something that better reflects this more general condition. E.g. is_trivial_type

- The condition in the generator describes a rather general class of datatypes, i.e. ones that have no relations or vector members.

ENDRELEASENOTES

Fixes #288

### DIFF
--- a/doc/templates.md
+++ b/doc/templates.md
@@ -100,7 +100,8 @@ The following keys / variables are filled for each datatype
 | `forward_declarations`        | The forward declarations for the user facing classes header files. This is a nested dict, where the keys are namespaces and the leaf values are classes. |
 | `forward_declarations_obj`     | The forward declarations for the `Obj` classes header files.                                                                                             |
 | `is_pod`                      | Flag value indicating whether the `Data` class of this datatype is a POD or if it contains an STL member                                                 |
-| `obj_needs_destructor`        | Flag indicating whether the `Obj` class needs a dedicated destructor for the datatype or whether `= default` is enough.                                  |
+| `obj_needs_destructor`        | Flag indicating whether the `Obj` class needs a dedicated destructor for the datatype or whether `= default` is enough.          
+| `is_trivial_type`        | Describes a general class of data types i.e. ones which have no relations or vector members.                            |
 | `ostream_collection_settings` | A dict with a single `header_contents` key that is necessary for the output stream overload implementation of collections                                |
 
 
--- a/doc/templates.md
+++ b/doc/templates.md
@@ -100,7 +100,6 @@ The following keys / variables are filled for each datatype
 | `forward_declarations`        | The forward declarations for the user facing classes header files. This is a nested dict, where the keys are namespaces and the leaf values are classes. |
 | `forward_declarations_obj`     | The forward declarations for the `Obj` classes header files.                                                                                             |
 | `is_pod`                      | Flag value indicating whether the `Data` class of this datatype is a POD or if it contains an STL member                                                 |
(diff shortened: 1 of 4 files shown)


doc/templates.md
```diff
@@ -100,7 +100,8 @@ The following keys / variables are filled for each datatype
 | `forward_declarations`        | The forward declarations for the user facing classes header files. This is a nested dict, where the keys are namespaces and the leaf values are classes. |
 | `forward_declarations_obj`     | The forward declarations for the `Obj` classes header files.                                                                                             |
 | `is_pod`                      | Flag value indicating whether the `Data` class of this datatype is a POD or if it contains an STL member                                                 |
-| `obj_needs_destructor`        | Flag indicating whether the `Obj` class needs a dedicated destructor for the datatype or whether `= default` is enough.                                  |
+| `obj_needs_destructor`        | Flag indicating whether the `Obj` class needs a dedicated destructor for the datatype or whether `= default` is enough.
```
tmadlener, line 103:
```suggestion
```

This has been removed from the code, so we can also remove it from the documentation.

doc/templates.md
```diff
@@ -100,7 +100,8 @@ The following keys / variables are filled for each datatype
 | `forward_declarations`        | The forward declarations for the user facing classes header files. This is a nested dict, where the keys are namespaces and the leaf values are classes. |
 | `forward_declarations_obj`     | The forward declarations for the `Obj` classes header files.                                                                                             |
 | `is_pod`                      | Flag value indicating whether the `Data` class of this datatype is a POD or if it contains an STL member                                                 |
-| `obj_needs_destructor`        | Flag indicating whether the `Obj` class needs a dedicated destructor for the datatype or whether `= default` is enough.                                  |
+| `obj_needs_destructor`        | Flag indicating whether the `Obj` class needs a dedicated destructor for the datatype or whether `= default` is enough.          
+| `is_trivial_type`        | Describes a general class of data types i.e. ones which have no relations or vector members.                            |
```
tmadlener, line 104:
```suggestion
| `is_trivial_type`        | Flag that indicates that this is a *trivial* data type, i.e. one without relations or vector members.                            |
```
Just to clarify that this is a boolean value.

tmadlener (changes requested):
Hi @soumilbaldota,

Thank you very much for this PR. I have just two minor comments concerning the documentation (see below).

As you can see we have quite a few checks running in continuous integration (CI). These are essentially building the code with the changes from the PR and then running all the tests (as you can also do "offline"). Different workflows run this on different operating systems and using different compilers. There is an additional workflow: `pre-commit` that runs a few style checks and linters. PRs are in general only merged if all tests pass. (There sometimes are intermittent hick-ups which can be fixed by running them again, but we will identify those cases and trigger re-running for you).

Could you please fix the suggestions below. This should also make the `pre-commit` workflow succeed.

tmadlener (approved):

PR #296, opened 2022-06-03
running pre-commit locally

soumilbaldota:
BEGINRELEASENOTES
- Add instructions on how to run pre-commit locally to the documentation

ENDRELEASENOTES

### DIFF
--- a/doc/advanced_topics.md
+++ b/doc/advanced_topics.md
@@ -76,11 +76,9 @@ To implement your own transient event store, the only requirement is to set the 
 
 ## Running pre-commit
 
- - Install [homebrew](https://brew.sh/)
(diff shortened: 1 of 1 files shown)


tmadlener:
Thanks for adding this. I have a few minor comments:
- I think the installation via `pip` is probably more widespread, as it is probably more widespread than `pip` (especially if you already have a python installation)
- Instead of pointing to the `brew` homepage, could you instead point to the pre-commit homepage (https://pre-commit.com/)? That effectively has all the installation and usage instructions on their landing page.
- There seem to be some whitespace issues that fail the pre-commit workflow. If you run that locally, I think it automatically fixes these (we have not enabled this in CI though).

PR #310, opened 2022-06-28
Parsing Julia types from c++ types in yaml description

soumilbaldota:
BEGINRELEASENOTES
- Adding a new member to Member Variable -> julia_type
- Adding a function to parse the cpp type to a julia type
- Adding tests to test_MemberParser.py

ENDRELEASENOTES


python/generator_utils.py
```diff
@@ -86,6 +86,40 @@ def __str__(self):
...
+      if self.cpp_type.startswith('std::'):
+        self.cpp_type = self.cpp_type[5:]
+      if self.cpp_type in ALLOWED_FIXED_WIDTH_TYPES:
+        if ALL_FIXED_WIDTH_TYPES_RGX.match(self.cpp_type):
+          regex_string = re.search("(8|16|32|64)", self.cpp_type).group()
+          if self.cpp_type.startswith('u'):
+            self.cpp_type = ("UInt" + regex_string)
+          elif self.cpp_type.startswith('i'):
+            self.cpp_type = ("Int" + regex_string)
+          return self.cpp_type
```
tmadlener, line 113: This could potentially be simplified by adapting the `ALL_FIXED_WIDTH_TYPES_RGX`. In principle the number of bits is already inside a capture group, so it should available. Looking at the definition, only the `u` is not yet inside a capture group.

python/generator_utils.py
```diff
@@ -86,6 +86,40 @@ def __str__(self):
 
 class MemberVariable:
   """Simple class to hold information about a member variable"""
+
+  def parse_julia_type(self, **kwargs):
```
tmadlener, line 90: From a quick glance, this looks like it does not necessarily need to be a member function of the `MemberVariable` class. I would make it a free function and then simply call that on the assignment to the `self.julia_type`.

Additionally, I think it should take only one argument, or at least not a `**kwargs`, but rather named arguments instead.

python/generator_utils.py
```diff
@@ -86,6 +86,40 @@ def __str__(self):
...
+    """Parse the given c++ type to a Julia type"""
+    self.cpp_type = kwargs.pop('cpp_type', None)
+    is_array = kwargs.pop('is_array', False)
+
+    builtin_types_map = {"int": "Int32", "float": "Float32", "double": "Float64",
+                         "bool": "Bool", "long": "Int32", "unsigned int": "UInt32",
+                         "unsigned long": "UInt32", "char": "Char", "short": "Int16",
+                         "long long": "Int64", "unsigned long long": "UInt64", "std::string": "String"}
```
tmadlener, line 98:
```suggestion
                         "long long": "Int64", "unsigned long long": "UInt64"}
```
Since `std::string` has been marked deprecated in #305 I don't think we need to support that in the Julia interface. This might also make some of the handling below a bit easier to do.

tmadlener (commented):
Hi @soumilbaldota,

thanks for adding this. This looks pretty good already. I have a few comments inline that would potentially make the `parse_julia_type` function a bit easier to read as it should hopefully remove a few `if`s.

python/generator_utils.py
```diff
@@ -28,6 +28,31 @@ def _prefix_name(name, prefix):
...
+    if cpp_type.startswith('std::'):
+      cpp_type = cpp_type[5:]
+    if cpp_type in ALLOWED_FIXED_WIDTH_TYPES:
+      regex_string = re.split("(u|)int(8|16|32|64)_t", cpp_type)
+      cpp_type = regex_string[1].upper() + "Int" + regex_string[2]
+      return cpp_type
+
+  else:
+    array_type = get_julia_type(cpp_type=array_type)
+    return f"MVector{'{'}{array_size}, {array_type}{'}'}"
```
tmadlener, line 51:
```suggestion
    return f"MVector{{{array_size}, {array_type}}}"
```
slightly more readable with the same result

tmadlener:
The function looks good to me (apart from the nitpicky comment about the f-string).

I see that you have essentially duplicated the file structure of the c++ templates for the julia templates. I am not sure if we need all of those in the end, so I would suggest to add the julia template files as we need them.

pylint seems to be complaining about some formatting: https://github.com/AIDASoft/podio/runs/7131646471?check_suite_focus=true#step:4:280

tmadlener:
Could you also add some checks to the MemberParser tests, to make sure that the parsed julia types are as expected?

These tests are here 
https://github.com/AIDASoft/podio/blob/9d078f82d366e9363cbf2e941e3647badbcba690/python/test_MemberParser.py#L15

The `parsed` in these tests is a `MemberVariable` so testing the `julia_type` should be almost trivial.

python/test_MemberParser.py
```diff
@@ -87,38 +97,44 @@ def test_parse_valid(self):  # pylint: disable=too-many-statements
...
     self.assertEqual(parsed.description, 'hopefully')
+    self.assertEqual(parsed.julia_type, r'TopLevelNamespaceType')
 
     parsed = parser.parse(r'std::array<::GlobalType, 1> anArray // with a top level type')
     self.assertEqual(parsed.full_type, r'std::array<::GlobalType, 1>')
     self.assertEqual(parsed.name, r'anArray')
     self.assertEqual(parsed.description, r'with a top level type')
     self.assertTrue(not parsed.is_builtin_array)
     self.assertEqual(parsed.array_type, r'::GlobalType')
+    self.assertEqual(parsed.julia_type, r'MVector{1, ::GlobalType}')
```
soumilbaldota, line 121: Should this look like:
`self.assertEqual(parsed.julia_type, r'MVector{1, ::GlobalType}')`

tmadlener (approved):
Hi @soumilbaldota 

this looks good to me. Is there anything still missing from your side, that you would like to still have in this PR?

Otherwise I would suggest to have the next step in a new PR.

soumilbaldota:
Hi @tmadlener ,
There is nothing more on this pull request. It can be merged now.

PR #311, opened 2022-07-04
Julia preprocessing

soumilbaldota:
BEGINRELEASENOTES
- Adding a julia include (include_jl) in the dictionary passed to the fill_templates

ENDRELEASENOTES


python/templates/Julia.jl.jinja2
```diff
@@ -0,0 +1,18 @@
+{% for include in includes_jl %}
+{{ include }}
+{% endfor %}
+
+mutable struct {{ class.bare_type }}
+{% for member in Members %}
+	{{member.name}}::{{member.julia_type }}
```
tmadlener, line 7:
```suggestion
	{{ member.name }}::{{ member.julia_type }}
```
This is not really necessary and also not written down anywhere, but for the other templates we tried to follow this style where we had spaces before/after the `{{`/`}}`. It is slightly more readable, and in languages that use `{}` it makes it slightly easier to spot what is a template replacement and what is not.

python/podio_class_generator.py
```diff
@@ -299,21 +310,26 @@ def _preprocess_for_class(self, datatype):
...
 
   def _preprocess_for_collection(self, datatype):
     """Do the necessary preprocessing for the collection"""
-    includes_cc, includes = set(), set()
+    includes_cc, includes, includes_jl = set(), set(), set()
 
     for relation in datatype['OneToManyRelations'] + datatype['OneToOneRelations']:
       if datatype['class'].bare_type != relation.bare_type:
         includes_cc.add(self._build_include(relation.bare_type + 'Collection'))
+        includes_jl.add(self._build_include(relation.bare_type + 'Collection', julia=True))
```
tmadlener, line 323: Is this necessary? Respectively can you use the generated files with this? I am asking because as far as I can see there is no `XYZCollection.jl` that we generate here(?)

python/podio_class_generator.py
```diff
@@ -359,29 +375,35 @@ def _preprocess_datatype(self, name, definition):
     data = deepcopy(definition)
     data['class'] = DataType(name)
     data['includes_data'] = self._get_member_includes(definition["Members"])
+    data['includes_data_jl'] = self._get_member_includes(definition["Members"], julia=True)
```
tmadlener, line 378: Is this necessary? In the templates I don't see that we use `includes_data_jl`

python/podio_class_generator.py
```diff
@@ -299,21 +310,26 @@ def _preprocess_for_class(self, datatype):
...
+        includes_jl.add(self._build_include(relation.bare_type + 'Collection', julia=True))
         includes.add(self._build_include(relation.bare_type))
+        includes_jl.add(self._build_include(relation.bare_type, julia=True))
 
     if datatype['VectorMembers']:
       includes_cc.add('#include <numeric>')
 
     datatype['includes_coll_cc'] = self._sort_includes(includes_cc)
     datatype['includes_coll_data'] = self._sort_includes(includes)
+    datatype['includes_jl'].update(self._sort_includes(includes_jl))
```
tmadlener, line 332: Does the `_sort_includes` work as expected with the Julia includes?

python/templates/Julia.jl.jinja2
```diff
@@ -0,0 +1,18 @@
...
+{% endfor %}
+{% for relation in OneToManyRelations %}
+	{{relation.name}}::{{relation.julia_type }}
```
tmadlener, line 10:
```suggestion
	{{relation.name}}::Vector{ {{ relation.julia_type }} }
```
Since we can have an arbitrary number of relations here

tmadlener (commented):
Hi @soumilbaldota,

I have a few comments (some are inline) that are easier to handle right at the beginning as they are probably less work to address at the moment. What you have until now definitely goes into the right direction, I just would do some things slightly differently, and some things might not be necessary for the Julia part. However, you are currently working on this on certainly have more insights than me, so if I missed something let me know.

- The name of the template is currently `Julia.jl.jinja2`. I think here you have to find a slightly more descriptive name, because there might be more Julia template files and in principle the fact that it is a Julia template is already encoded in the `.jl` suffix. A good name here should reflect that the template is used for a mutable Julia type/struct.
- You have put the necessary preprocessing steps for the Julia template into the different existing preprocessing functions for the c++ code. Here I think it would make sense to define an entirely new `_preprocess_for_julia` (or similar) method, so that all the necessary steps can be found in one place. This makes it much easier to maintain in the future.
- There are some things that look like they are there simply to reflect the c++ side of code generation. I am not entirely sure they are actually necessary. Making the julia preprocessing a separate method would make it easier to actually spot these and only handle what is actually necessary. (i.e. only add things here if they are needed in the templates)

python/podio_class_generator.py
tmadlener: I see they are used in the templates now. Is there a reason why they have to go to a different list than the usual `includes_jl`?

tmadlener:
This looks pretty nice already. Is the generated code working? Can you add just a few tests using the example datamodel that comes with podio. I don't think it has to cover everything at this point, but e.g. something using the `ExampleCluster`, `ExampleHit`, and `ExampleMC` would be nice. Maybe we can run them in the CI (although I am not sure if any of our workflow environment actually comes with Julia)

I have one question / comment. Maybe you have a stronger opinion here than me. From the current status it looks like the `components` use the same template as the `datatypes`. I don't currently see a technical reason not do so for Julia, since essentially we do not really differentiate between the two (yet?). Do you think it would make sense to introduce a template for components for Julia? This could then obviously be made much simpler than the one for datatypes, since by definition components cannot have any sort of relation or vector members. Or would you rather leave it with one template for the moment and introduce a separate one when it is actually needed?

soumilbaldota:
@tmadlener 

> I don't think it has to cover everything at this point, but e.g. something using the ExampleCluster, ExampleHit, and ExampleMC would be nice. 

The generated code is working for them and I have run it with [datalayout.yaml](https://github.com/AIDASoft/podio/blob/master/tests/datalayout.yaml).

>  Or would you rather leave it with one template for the moment and introduce a separate one when it is actually needed?

I think I can include this later also.

Currently I am facing an issue that when I run class_generator with [edm4hep.yaml](https://github.com/key4hep/EDM4hep/blob/master/edm4hep.yaml). It generates the `Vertex.jl` and `ReconstructedParticle.jl` but they are interdependent and try to include each other resulting in a dead loop. I am confused on how to fix them.
The code generated looks fine as per the prototype made earlier.

python/podio_class_generator.py
soumilbaldota: I think they can go in the same list. I will change it in the next commit.

tmadlener:
> Currently I am facing an issue that when I run class_generator with [edm4hep.yaml](https://github.com/key4hep/EDM4hep/blob/master/edm4hep.yaml). It generates the `Vertex.jl` and `ReconstructedParticle.jl` but they are interdependent and try to include each other resulting in a circular inclusion. I am confused on how to fix them.
EDIT: Will need some forward declarations . 
The code generated looks fine as per the prototype made earlier.

Ah that is an excellent double check to see if things are working as expected and whether we have overlooked something. Are forward declarations possible in Julia? Or do we have to think about a different solution to make this work with datatypes that refer to each other?

soumilbaldota:
>  Are forward declarations possible in Julia? 

They are not. 
If you have a suitable way or idea to fix it please let me know.
And the workarounds I have found are not ideal for our use case (two separate files.)
- (https://github.com/JuliaLang/julia/issues/269#issuecomment-68421745)
- https://github.com/JuliaLang/julia/issues/269#issuecomment-408254437

The below only works when both the structs are in the same file.

```julia
mutable struct Vertex{T}
	associatedParticle::Union{T ,Nothing}
end
mutable struct ReconstructedParticle{T}
	startVertex::Union{Nothing, T }
end
function Vertex()
	return Vertex{ReconstructedParticle}(
	nothing,
	)
end
function ReconstructedParticle()
	return ReconstructedParticle{Vertex}(
	nothing,
	)
end
```

tmadlener:
> If you have a suitable way or idea to fix it please let me know.

I will have to think about it as well. Maybe there is an elegant solution using abstract types or something, but from the top of my head I am not entirely sure. Maybe @hegner has a clever idea for this.

Another thing that we have not really dealt with in our prototype is the fact that the yaml definition allows to introduce a namespace (in c++), e.g. `edm4hep` for EDM4hep. Here I think we should be able to use Julia modules to achieve a very similar effect. Could you have a look at that and see what is possible in that regard? One thing that would need clarification from my side is whether we are able to "re-open" a module in different files. That is something that is possible in c++, but for Julia I am not entirely sure.

python/podio_class_generator.py
```diff
@@ -222,15 +229,40 @@ def _process_datatype(self, name, definition):
...
+            includes_jl.add(self._build_include(relation.bare_type, julia=True, is_struct=True))
+          # if datatype['class'].bare_type != relation.bare_type:
+          #   includes_jl.add(self._build_include(relation.bare_type + 'Collection', julia=True))
+    try:
+      includes_jl.remove(self._build_include(datatype['class'].bare_type, julia=True, is_struct=True))
+    except KeyError:
+      pass
+    datatype['includes_jl']['constructor'].update((includes_jl))
+
+  def _get_params(self, datatype):
```
tmadlener, line 256:
```suggestion
  def _get_julia_params(self, datatype):
```
Just to make it clear that this is necessary for Julia.

Could you also put a docstring here that very briefly explains what this function does and why it is necessary?

python/templates/Constructor.jl.jinja2
```diff
@@ -0,0 +1,26 @@
+include("{{ class.bare_type }}Struct.jl")
+{% for include in includes_jl['constructor'] %}
+{{ include }}
+{% endfor %}
+function {{ class.bare_type }}()
+	return {{ class.bare_type }}{% if params_jl %}{ {% for relation in params_jl %} {{ relation }}, {% endfor %} }{% endif %}(
+	{% for member in Members %}
+	{% if member.is_array %}
+	{{ member.julia_type }}(undef),
```
tmadlener, line 9: Can we initialize this with another default (that makes sense)?

tmadlener (commented):
Nice approach and implementation for the templates. I have only a few minor comments inline and would also want to ask you whether you could add a couple of tests that use the [example datamodel](https://github.com/AIDASoft/podio/tests/datalayout.yaml) that comes with podio to see if things work with that. At least for the most important use cases like relations.

Could you also have a look at the CI failures and fix any obvious problems. Let me know if you need help with that.

python/podio_class_generator.py
soumilbaldota: I have already included the docstring.

python/templates/Constructor.jl.jinja2
soumilbaldota: we can initialize it using other methods but, undef is better due to [better perfmance](https://stackoverflow.com/questions/58430581/julia-what-is-undef-in-array-in-julia) as it just allocates the memory required instead of initializing each variable.

soumilbaldota:
/home/runner/work/podio/podio/tests/src/ExampleWithOneRelationSIOBlock.cc:13:29: error: unused parameter 'device' [-Werror,-Wunused-parameter]
void handlePODDataSIO(devT& device, ::ExampleWithOneRelationData* data, size_t size) {
                            ^
/home/runner/work/podio/podio/tests/src/ExampleWithOneRelationSIOBlock.cc:13:67: error: unused parameter 'data' [-Werror,-Wunused-parameter]
void handlePODDataSIO(devT& device, ::ExampleWithOneRelationData* data, size_t size) {
                                                                  ^
2 errors generated.




The above two errors are making the pre-commit fail but i dont know why.

python/templates/Constructor.jl.jinja2
tmadlener: That is a very good point. I didn't think about that.

python/podio_class_generator.py
```diff
@@ -222,15 +229,42 @@ def _process_datatype(self, name, definition):
...
+    try:
+      includes_jl.remove(self._build_include(datatype['class'].bare_type, julia=True, is_struct=True))
+    except KeyError:
+      pass
+    datatype['includes_jl']['constructor'].update((includes_jl))
+
+  def _get_julia_params(self, datatype):
+    """Get the parameteric types for MutableStructs"""
+    params = set()
+    for relation in datatype['OneToManyRelations'] + datatype['VectorMembers'] + datatype['OneToOneRelations']:
```
tmadlener, line 259:
```suggestion
    """Get the relations as parameteric types for MutableStructs"""
    params = set()
    for relation in datatype['OneToManyRelations'] + datatype['OneToOneRelations']:
```
Just to make the docstring a bit more clear. 
Also I am not sure we actually discussed this during prototyping, but looking at the code now, I just realize that the `VectorMembers` are not needed here, because they can only be builtin types, or components. This means that there should really be no way to end up with a cyclic dependency.

python/podio_class_generator.py
tmadlener: Ah right, Looks like I was just commenting while you were fixing the pre-commit issue for the docstring.

python/podio_class_generator.py
```diff
@@ -359,29 +393,38 @@ def _preprocess_datatype(self, name, definition):
     data = deepcopy(definition)
     data['class'] = DataType(name)
     data['includes_data'] = self._get_member_includes(definition["Members"])
-    data['is_pod'] = self._is_pod_type(definition["Members"])
+    data['includes_jl'] = {'constructor': self._get_member_includes(definition["Members"], julia=True),
+                           'struct': self._get_member_includes(definition["Members"], julia=True),
+                           'is_pod': self._is_pod_type(definition['Members'])}
```
tmadlener, line 398:
```suggestion
                           'struct': self._get_member_includes(definition["Members"], julia=True)}
    data['is_pod'] = self._is_pod_type(definition['Members'])
```
This should fix the c++ warnings you see, as `is_pod` needs to be available at the top level in the templates

python/podio_class_generator.py
```diff
@@ -222,15 +229,38 @@ def _process_datatype(self, name, definition):
...
+        includes_jl.add(self._build_include(relation.bare_type, julia=True, is_struct=True))
+        # if datatype['class'].bare_type != relation.bare_type:
+        #   includes_jl.add(self._build_include(relation.bare_type + 'Collection', julia=True))
+    for member in datatype['VectorMembers']:
+      if self._needs_include(member) and not member.is_builtin:
+        includes_jl_struct.add(self._build_include(member.bare_type, julia=True))
+    datatype['includes_jl']['constructor'].update((includes_jl))
+    datatype['includes_jl']['struct'].update((includes_jl_struct))
+
+  def _get_julia_params(self, datatype):
```
tmadlener, line 252:
```suggestion
  @staticmethod
  def _get_julia_params(datatype):
```
Easy fix for the pre-commit complaint

tmadlener (changes requested):
One thing I just realized is that the current version is not working for me, because the generated Julia code is also passed through `clang-format`, which messes with the formatting and makes the generated Julia code not work in the end.

You will have to add a check here
https://github.com/AIDASoft/podio/blob/a8efda986f776892e90e7aafa4958aaf18a1e6c7/python/podio_class_generator.py#L154

that makes the Julia files not being run through `clang-format`. Should be as simple as checking for the file-ending I hope.

soumilbaldota:
I have added Collections in a separate template to Constructor because we are only making Collections of Datatypes and not Components and hence called fill_templates only in process_datatype()

tests/unittest.jl
```diff
@@ -0,0 +1,78 @@
+include("../Tmp/data/ExampleMC.jl")
+include("../Tmp/data/ExampleWithVectorMember.jl")
+include("../Tmp/data/ExampleForCyclicDependency1.jl")
+include("../Tmp/data/ExampleForCyclicDependency2.jl")
+include("../Tmp/data/ExampleWithOneRelation.jl")
+include("../Tmp/data/ExampleCluster.jl")
+
+using Test
+@testset "All" begin
```
tmadlener, line 9:
```suggestion
@testset "Julia Bindings" begin
```
Just a slightly less generic way

tmadlener (commented):
Excellent. Could you just add one or two tests that also use the collections? Obviously since they are `Vector{T}` at the moment, there is probably nothing that breaks them, but if we need a more complicated implementation in the future we would at least immediately see where they break.

python/templates/Constructor.jl.jinja2
```diff
@@ -0,0 +1,26 @@
+include("{{ class.bare_type }}Struct.jl")
+{% for include in includes_jl['constructor'] %}
+{{ include }}
+{% endfor %}
+function {{ class.bare_type }}()
+	return {{ class.bare_type }}{% if params_jl %}{ {% for relation in params_jl %} {{ relation }}, {% endfor %} }{% endif %}(
```
tmadlener, line 6:
```suggestion
	return {{ class.bare_type }}{% if params_jl %}{ {{ params_jl | join(',') }} }{% endif %}(
```
I just realized that this can probably be simplified like suggested. This makes this template slightly more readable and has the additional advantage that we are not left with a trailing comma.

We are already using it in a few places, e.g. here, where we even pick out an attribute for each element of the list:
https://github.com/AIDASoft/podio/blob/a8efda986f776892e90e7aafa4958aaf18a1e6c7/python/templates/macros/declarations.jinja2#L62

python/templates/JuliaCollection.jl.jinja2
```diff
@@ -0,0 +1,2 @@
+include("{{ class.bare_type }}.jl")
+{{ class.bare_type }}Collection = Vector{ {{ class.bare_type }}{% if params_jl %}{ {% for relation in params_jl %} {{ relation }}, {% endfor %} }{% endif %} }
```
tmadlener, line 2:
```suggestion
{{ class.bare_type }}Collection = Vector{ {{ class.bare_type }}{% if params_jl %}{ {{ params_jl | join(',') }} }{% endif %} }
```

python/templates/MutableStruct.jl.jinja2
```diff
@@ -0,0 +1,18 @@
+{% for include in includes_jl['struct'] %}
+{{ include }}
+{% endfor %}
+mutable struct {{ class.bare_type }}{% if params_jl %}{ {% for relation in params_jl %} {{ relation }}T, {% endfor %} }{% endif %}
```
tmadlener, line 4:
```suggestion
mutable struct {{ class.bare_type }}{% if params_jl %}{ {{ params_jl | join('T, ') }} }{% endif %}
```

tmadlener (approved):

PR #316, opened 2022-07-27
Remove the macOS CI workflow because they will no longer work

tmadlener:
BEGINRELEASENOTES
- Remove macOS CI workflows because github hosted runners will deprecate macOS 10.15 ([announcement](https://github.com/actions/virtual-environments/issues/5583)) and later versions of macOS no longer support fuse and as a consequence CVMFS.

ENDRELEASENOTES


andresailer (approved):

tmadlener:
See also discussion in cvmfs-contrib/github-action-cvmfs#17

PR #317, opened 2022-07-29
Allow for extending datamodels with an upstream datamodel

### LINKED ISSUE #263: Allow easier extension of data models (e.g. EIC builds on edm4hep)
In EIC we have found ourselves unable to build _on_ edm4hep, and instead necessitated to _reimplement_ edm4hep in an eicd namespace. While we can create new data types and combine them in the same file with edm4hep data types, this ultimately does not allow that much freedom since we wish to maintian a 'chain of custody' with relations to the originating objects. Pretty quickly that means we cannot extricate ourselves from a web of edm4hep data types.

While it is to some extent intentional that inheritance of podio data models is restricted to avoid fragmentation with incompatible extensions of data models, this is now leading to duplication of the same data model in a different namespace where there is no connection any more to the originating data model. That would appear on the face of it to be a worse situation. As we understand this is not limited to EIC vis-a-vis edm4hep. 

We wonder if a more flexible structure that allows some inheriting of data model components can be allowed.

tmadlener:
BEGINRELEASENOTES
- Make it possible to pass an upstream datamodel to the class generator such that datatypes and components defined there can be used in an unrelated datamodel. This makes it possible to extend datamodels and to prototype new datatypes with the aim of upstreaming them eventually without having to redefine all the necessary components.
- Refactor the internals of the config reader / class generator slightly to make it possible to hold multiple datamodels in memory

ENDRELEASENOTES

Fixes #263

Turned out to be a little more work than originally anticipated due to some additional refactoring that was necessary.

@vvolkl @wdconinc @hegner does this solve our issues with extending datamodels?

### DIFF
--- a/cmake/podioMacros.cmake
+++ b/cmake/podioMacros.cmake
@@ -121,6 +121,8 @@ set_property(CACHE PODIO_USE_CLANG_FORMAT PROPERTY STRINGS AUTO ON OFF)
 #   Parameters:
 #      OUTPUT_FOLDER        OPTIONAL: The folder in which the output files should be placed
 #                           Default is ${CMAKE_CURRENT_SOURCE_DIR}
+#      UPSTREAM_EDM         OPTIONAL: The upstream edm and its package name that are passed to the
+#                           generator via --upstream-edm
 #      IO_BACKEND_HANDLERS  OPTIONAL: The I/O backend handlers that should be generated. The list is
 #                           passed directly to podio_class_generator.py and validated there
 #                           Default is ROOT
@@ -130,10 +132,15 @@ set_property(CACHE PODIO_USE_CLANG_FORMAT PROPERTY STRINGS AUTO ON OFF)
 # this is essentially a no-op, and should not cause re-compilation.
 #---------------------------------------------------------------------------------------------------
 function(PODIO_GENERATE_DATAMODEL datamodel YAML_FILE RETURN_HEADERS RETURN_SOURCES)
-  CMAKE_PARSE_ARGUMENTS(ARG "" "OUTPUT_FOLDER" "IO_BACKEND_HANDLERS" ${ARGN})
+  CMAKE_PARSE_ARGUMENTS(ARG "" "OUTPUT_FOLDER;UPSTREAM_EDM" "IO_BACKEND_HANDLERS" ${ARGN})
   IF(NOT ARG_OUTPUT_FOLDER)
     SET(ARG_OUTPUT_FOLDER ${CMAKE_CURRENT_SOURCE_DIR})
   ENDIF()
+  SET(UPSTREAM_EDM_ARG "")
+  IF (ARG_UPSTREAM_EDM)
+    SET(UPSTREAM_EDM_ARG "--upstream-edm=${ARG_UPSTREAM_EDM}")
+  ENDIF()
+
   IF(NOT ARG_IO_BACKEND_HANDLERS)
     # At least build the ROOT selection.xml by default for now
     SET(ARG_IO_BACKEND_HANDLERS "ROOT")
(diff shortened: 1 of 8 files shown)


hegner:
Great! Another item removed from my todo list. :-)
Need to do a few minor adjustments for the schema evolution part, but that is fine

tmadlener:
> Need to do a few minor adjustments for the schema evolution part, but that is fine

Shall we do them in this PR still or rather together with schema evolution?

hegner:
Though the work is not that complicated, it needs some agreement on policies. Take the following:

-   edm4hep v1 to v2 evolution needs (for the non-trivial) cases edm4hep_v1.yaml and edm4hep_v2.yaml

Once we have a data model that builds upon edm4hep and e.g. uses a component that non-trivially changed in edm4hep - how do we handle that? In its current form, it would require to explicitly provide all of 

- edm4hep_v1.yaml , edm4hep_v2.yaml , myEdm_vN.yaml, myEdm_vN+1.yaml 

Which may not be that user friendly. In particular if you have multiple levels there. So we have to find a way to automate some of the lookups without having a need for explicit command line passing.

Besides that - I made as well a change to hold two datamodels in memory at the same time. Will look into which version to keep.

tmadlener:
Naively I would have said podio doesn't try to do schema evolution across datamodels and we should potentially advertise this possibility more as a way to do prototyping for new datatypes rather than as a full fledged means to extend arbitrary datamodels. Let's discuss this at the meeting next Tuesday (should be with a US friendly timeslot).

-----
> Besides that - I made as well a change to hold two datamodels in memory at the same time. Will look into which version to keep.

Mine is as trivial as it gets, so if you need anything more than what is essentially a `struct` for keeping together a few pieces of data it will not be very useful ;)

wdconinc:
I'm looking into some tests with this (extending EDM4hep with our inclusive kinematics additions).

This requires the upstream yaml file. Should podio then install that yaml file for data models?

I'd imagine something like this in our CMakeLists.txt:
```
find_package(EDM4HEP 0.4.1 REQUIRED)
PODIO_GENERATE_DATAMODEL(eicd eic_data.yaml headers sources
  UPSTREAM_EDM edm4hep:${EDM4HEP_YAML}/edm4hep.yaml
  OUTPUT_FOLDER ${CMAKE_CURRENT_BINARY_DIR}
  )
```
where the yaml file is in `share/doc/EDM4HEP` or so? (not sure what FHS says about this)

Thoughts?

tmadlener:
I would put the responsibility of installing the upstream yaml file on the upstream datamodel (e.g. EDM4hep). That would also give datamodel authors at least some control over whether they want their datamodel easily extended or not.

Putting the yaml file into `<prefix>/share` sounds like the most intuitive place to put it. At least in Key4hep we usually have a `$<upper-case-package-name>` environment variable that points there in our spack recipes. (although EDM4hep is notably not doing it yet...). There are potentially more elegant ways via target properties in cmake and then simply assuming that we can locate it relative from a path we can get from there. In the end the former relies a bit more on a certain environment, while the latter is a line or two more in cmake, but we can let cmake do the heavy lifting of locating EDM4hep. I have no strong opinion on either.

wdconinc:
Our tests indicate this works exactly as hoped. In particular, this will allow us to
- revert to using EDM4hep vector components in derived data model definitions,
- create associations between EDM4hep collections and derived data model collections.

tmadlener:
Thanks for checking and letting as know.

@hegner, do you think the schema evolution bits can be added later without changing behavior? In that case we could consider merging this rather early, as this would solve at least two issues of not being able to extend EDM4hep at the moment.

tmadlener:
Unless there are any urgent changes that still need to go into this one, I would merge this in order for the nightly builds to pick this up, so that some of the pull requests that depend on this can run their tests in their CI workflows.

PR #318, opened 2022-08-03
fix: explicitly use SELECTION_XML in OUTPUT_FOLDER

wdconinc:
After checking that `${ARG_OUTPUT_FOLDER}/${SELECTION_XML}` exists, we should not simply pass `${SELECTION_XML}` to `PODIO_GENERATE_DICTIONARY`, but pass the full path for when `${ARG_OUTPUT_FOLDER}` is not `${CMAKE_CURRENT_SOURCE_DIR}`.

BEGINRELEASENOTES
- Bugfix for data models in OUTPUT_FOLDER not equal to source dir in root dictionary generation cmake macro.

ENDRELEASENOTES


tmadlener:
Good catch. Thank you. 
Tow minor things (I can also fix those, just let me know. I should be able to push directly to the PR branch usually)
- Could you also update the docstring of the macro to include that `SELECTION_XML` is relative to `OUTPUT_FOLDER`? 
- The podio internal tests of the extenstion datamodel also do not seem to get this correct here: https://github.com/AIDASoft/podio/blob/d0d4f671b8f5e41d6e11c07fe54c6c8d93d32930/tests/CMakeLists.txt#L199-L200

wdconinc:
> Could you also update the docstring of the macro to include that SELECTION_XML is relative to OUTPUT_FOLDER?

We could also allow both relative and absolute paths to selection file and use cmake to figure which it is, similar to https://github.com/AIDASoft/podio/blob/master/cmake/podioMacros.cmake#L29-L34.

The other approach could be to change the test for existence of the selection.xml in https://github.com/AIDASoft/podio/blob/master/cmake/podioMacros.cmake#L273-L276 to something like
```cmake
  IF (NOT EXISTS ${ARG_OUTPUT_FOLDER}/${SELECTION_XML}
  AND NOT EXISTS ${SELECTION_XML})
    MESSAGE(STATUS "Not adding the ROOT dictionary to the targets because the selection.xml has not been generated")
    RETURN()
  ENDIF()
```

I'll let you fix this in whichever way you think best.

tmadlener:
Thanks for the `IS_ABSOLUTE` hint. Now `SELECTION_XML` can either be absolute or relative to `OUTPUT_FOLDER`. Also updated the documentation and made the status message a bit more informative.

tmadlener:
`dev4/x86_64-centos7-gcc11-opt` seems to be broken entirely, we pick up the default cmake from CentOS7.

cmake/podioMacros.cmake
```diff
@@ -259,7 +259,7 @@ endfunction()
 #    dict_name            Name of the dictionary
 #    CORE_LIB             The core datamodel library (e.g. from PODIO_ADD_DATAMODEL_CORE_LIB)
 #    HEADERS              The list of all header files generated by PODIO_GENERATE_DATAMODEL
-#    SELECTION_XML        The selection.xml file genertaed by PODIO_GENERATE_DATAMODEL
+#    SELECTION_XML        The selection.xml file genertaed by PODIO_GENERATE_DATAMODEL (either an absolut path or relative to OUTPUT_FOLDER)
```
tmadlener, line 262:
```suggestion
#    SELECTION_XML        The selection.xml file genertaed by PODIO_GENERATE_DATAMODEL (either an absolute path or relative to OUTPUT_FOLDER)
```

tmadlener (approved):

Issue #314, opened 2022-07-27
EventStore + ROOTReader ownership of CollectionIDTable


faustus123:
- OS version: all
 - Compiler version: all
 - PODIO version: v00-14-01
 
This took me quite a while to figure out.

What I am trying to do is read events into multiple EventStore objects using the same ROOTReader. The reads are being done sequentially. The idea is once an event is read in, I can pass the EventStore into one thread and then I read another event into a different EventStore object for a different thread. This, however gives seg. faults. The issue (or at least the main one right now) is the following:

In ROOTReader::OpenFiles a _CollectionIDTable_ is instantiated and the _ROOTReader_ object keeps a pointer to it.

When EventStore::setReader is called, it gets that pointer and places it in a unique_ptr. This essentially gives ownership to the EventStore even though ROOTReader still has a pointer that it uses.

Deleting the EventStore object will free the CollectionIDTable leaving the ROOTReader with an invalid pointer. Calling EventStore::setReader using a new event store will cause a second unique_ptr to exist to the CollectionIDTable created by ROOTReader. 

Without knowing all of the details, I can't say what the best fix is. I would guess either using shared_ptr everywhere, or just make the EventStore use a raw pointer rather than unique_ptr.

Oh, and one other suggestion. If ROOTReader placed things in protected: sections rather than private: I would be able to handle this sort of thing in my code. I know it is not really good programming form, but would give me a way to tweak things to figure it out and keep rolling for the moment until an official fix came out. Again, just a thought.

tmadlener:
The main issue is a recurring one: The `EventStore` that is shipped with podio was really only meant as an example implementation of how one could achieve the functionality and it has since unfortunately long outlived its original scope. We try to rectify this (and other things) with the `Frame` approach, where each `Frame` can stand on its own and different `Frame`s can easily be put onto different threads.

For this we have a prototype implementation in #287. This is essentially ready for testing as it is. The main reason why it has not been merged yet is because we are still working on getting schema evolution into this approach as well and we don't want to change the interface too often.

tmadlener:
Closing, as this is no longer relevant with the removal of the `EventStore`

PR #323, opened 2022-08-16
Add `podio-dump` tool to dump contents of podio files

tmadlener:
BEGINRELEASENOTES
- Add a `podio-dump` python script (installed to `<prefix>/bin` that can be used to dump event contents to stdout. By default prints an overview over the collections and their types, but can also be used to dump full events, via the `-d` or `--detailed` flag. Use `--help` to get all available options and their descriptions.
- To allow `podio-dump` to work with all available backends also add support for reading SIO via the `PythonEventStore`.
  - Split off the necessary c++ functionality into a separate `podioPythonStore` library (+ necessary ROOT dictionaries).
- Add a `print` function to the collections for easier dumping from the python side.
- Add a `print` function to the `GenericParameters`
- Make `goToEvent` is a part of the `IReader` interface and correctly implemented it for the `SIOReader`.

ENDRELEASENOTES

Supersedes #212, #242 and #240


include/podio/SIOReader.h
```diff
@@ -88,7 +90,7 @@ class SIOReader : public IReader {
 
   typedef std::pair<CollectionBase*, std::string> Input;
   std::vector<Input> m_inputs{};
-  CollectionIDTable* m_table{nullptr}; // will be owned by the EventStore
+  std::shared_ptr<CollectionIDTable> m_table{nullptr}; // will be owned by the EventStore
```
andresailer, line 93: With the shared_ptr the ownership comment doesn't make much sense

python/EventStore.py
```diff
@@ -73,6 +73,14 @@ def get(self, name):
       coll.__getitem__ = getitem
     return coll
 
+  def collections(self):
+    """Get all collection names."""
```
andresailer, line 77:
```suggestion
    """Return list of all collection names."""
```

python/EventStore.py
```diff
@@ -73,6 +73,14 @@ def get(self, name):
...
+    return [str(c) for c in self.current_store.getCollectionNames()]
+
+  def metadata(self):
+    """Get the metadata of the current event"""
```
andresailer, line 81: What is the type that is returned here?

python/test_EventStore.py
```diff
@@ -2,9 +2,17 @@
 
 from EventStore import EventStore
 
+#pylint: disable=no-member # pylint does not realize that the store is populated by inheriting classes
 
 class EventStoreBaseTestCase:
```
andresailer, line 7: Called it a `Mixin`, and pylint will know.
https://pylint.pycqa.org/en/2.5/faq.html#how-do-i-avoid-access-to-undefined-member-messages-in-my-mixin-classes

andresailer (commented):
Thanks for having a go at this so quickly.
I am mostly wondering about the collectionIDTable interface change.

include/podio/EventStore.h
```diff
@@ -112,16 +112,16 @@ class EventStore : public ICollectionProvider, public IMetaDataProvider {
...
+    m_table = std::move(table);
   }
 
   // members
   mutable std::set<int> m_retrievedIDs{};
   mutable CollContainer m_collections{};
   mutable std::vector<CollectionBase*> m_cachedCollections{};
   IReader* m_reader{nullptr};
-  std::unique_ptr<CollectionIDTable> m_table;
+  std::shared_ptr<CollectionIDTable> m_table;
```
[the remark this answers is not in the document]
tmadlener: Essentially this is needed for the read tests, which now also try to jump to some events and make sure that works. In the course of this the reader is only created once, but the `EventStore` is created multiple times. If the `EventStore` owns this uniquely, the first `EventStore` to go out of scope also destroys the table that is held by the reader.

The current design is overall a bit backwards I think. See also #314

Having (unique) ownership with the readers doesn't really work either, because `EventStore`s can be created without a reader, and in that case they have to own the table, because otherwise ownership would have to go to the writers, which again breaks, because there can be multiple writers to each `EventStore`.

This should all be solved much more cleanly with the Frame...

include/podio/EventStore.h
```diff
@@ -82,8 +82,8 @@ class EventStore : public ICollectionProvider, public IMetaDataProvider {
   /// set the reader
   void setReader(IReader* reader);
 
-  CollectionIDTable* getCollectionIDTable() const {
-    return m_table.get();
+  std::shared_ptr<CollectionIDTable> getCollectionIDTable() const {
```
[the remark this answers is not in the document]
tmadlener: Good point. Maybe we can leave this, but still have a `shared_ptr` internally.

tmadlener:
Just realized that the changes to the `IReader` break k4LCIOReader.

PR #333, opened 2022-09-20
Initialize unique_ptr in initializer list for nvcc

tmadlener:
BEGINRELEASENOTES
- Initialize the `unique_ptr<mutex>` in the constructor initializer list instead of in the member variable declaration. This is more likely a bug in nvcc (or maybe a c++17 feature not yet supported by nvcc). Fixes key4hep/k4Clue#34
- Pass `--disable-new-dtags` to the linker when using `PODIO_SET_RPATH`, to set `RPATH` and not `RUNPATH` in the binaries.
- Pin the ubuntu version for runners that build on ubuntu to not accidentally go out of sync with the underlying LCG releases.
- Disable the podio tests in the edm4hep workflows (see #359).

ENDRELEASENOTES

### DIFF
--- a/cmake/podioBuild.cmake
+++ b/cmake/podioBuild.cmake
@@ -27,6 +27,10 @@ macro(podio_set_rpath)
     if("${isSystemDir}" STREQUAL "-1")
       set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/${LIBDIR}")
     endif("${isSystemDir}" STREQUAL "-1")
+    # Make sure to actualy set RPATH and not RUNPATH by disabling the new dtags
+    # explicitly. Set executable and shared library linker flags for this
+    set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--disable-new-dtags")
+    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,--disable-new-dtags")
   else()
     set(CMAKE_SKIP_INSTALL_RPATH TRUE)           # skip the full RPATH for the install tree
   endif()
--- a/python/templates/Collection.cc.jinja2
+++ b/python/templates/Collection.cc.jinja2
@@ -17,7 +17,7 @@
 
 {% with collection_type = class.bare_type + 'Collection' %}
 {{ collection_type }}::{{ collection_type }}() :
-  m_isValid(false), m_isPrepared(false), m_isSubsetColl(false), m_collectionID(0), m_storage() {}
+  m_isValid(false), m_isPrepared(false), m_isSubsetColl(false), m_collectionID(0), m_storageMtx(std::make_unique<std::mutex>()), m_storage() {}
 
 {{ collection_type }}::{{ collection_type }}({{ collection_type }}Data&& data, bool isSubsetColl) :
   m_isValid(false), m_isPrepared(false), m_isSubsetColl(isSubsetColl), m_collectionID(0), m_storage(std::move(data)) {}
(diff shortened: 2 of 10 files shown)


tmadlener:
Pipeline for seeing if these changes actually do fix the nvcc problem: https://gitlab.cern.ch/key4hep/k4-deploy/-/pipelines/4516064

tmadlener:
For some of those it might become a bit more work than for the collections, because if you cannot directly initialize in the declaration the constructor can no longer by `default`ed.

vvolkl:
Seems like it wasn't too bad: with https://github.com/tmadlener/podio/pull/4 k4Clue builds ok - there's one segfault in the gaudiwrapper test but that's probably because I didn't rebuild k4fwcore with the modified podio.

ebrondol:
The issue was tested with CUDA 12.0 (released this week) and the problem is still there. 
Unfortunately, we need this workaround.

python/templates/Collection.cc.jinja2
```diff
@@ -17,7 +17,7 @@
 
 {% with collection_type = class.bare_type + 'Collection' %}
 {{ collection_type }}::{{ collection_type }}() :
-  m_isValid(false), m_isPrepared(false), m_isSubsetColl(false), m_collectionID(0), m_storage() {}
+  m_isValid(false), m_isPrepared(false), m_isSubsetColl(false), m_collectionID(0), m_storageMtx(std::make_unique<std::mutex>()), m_storage() {}
 
 {{ collection_type }}::{{ collection_type }}({{ collection_type }}Data&& data, bool isSubsetColl) :
```
tmadlener, line 22: This constructor is currently missing the initialization of the mutex.

tmadlener (commented):
I will rebase this ontop of the current master, and try to have another look and make sure I got all the constructors....

tmadlener:
I have to check the ubuntu builds again. I am not yet entirely sure why they break

vvolkl:
I can confirm that even after the rebase this PR still fixes k4CLUE builds.

cmake/podioBuild.cmake
```diff
@@ -27,6 +27,10 @@ macro(podio_set_rpath)
     if("${isSystemDir}" STREQUAL "-1")
       set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/${LIBDIR}")
     endif("${isSystemDir}" STREQUAL "-1")
+    # Make sure to actualy set RPATH and not RUNPATH by disabling the new dtags
+    # explicitly. Set executable and shared library linker flags for this
+    set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--disable-new-dtags")
+    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,--disable-new-dtags")
```
tmadlener, line 33: And I hope this "fixes" the Ubuntu CI builds. The problem is that on Ubuntu setting `RPATH`, actually sets `RUNPATH` which has lower precedence than `LD_LIBRARY_PATH`. Debian based systems, seem to have different default linker flags, which set `RPATH` instead of `RUNPATH`.

Following a few stackoverflow threads, there seems to be no real consensus which one is the "correct" one. In any case these changes **force `RPATH`, i.e. there is now way to override via the environment**. At least for Key4hep it should not really be a problem, since we always only have one podio in the environment.

tmadlener:
All workflows running on `ubuntu-20.04` are currently failing to install cvmfs, because https://ecsft.cern.ch/dist/cvmfs/cvmfs-release/cvmfs-release-latest_all.deb seems to be unavailable at the moment.

tmadlener:
I have disabled the podio tests in the edm4hep workflows for the moment, because using `PODIO_SET_RPATH` does not seem to work in downstream edm4hep in this environment, even when also using `RPATH` to build there. I will open an issue to keep track of that, but in order to unblock this, I figured disabling the podio tests (which are run in any case in other workflows) is the easiest way forward.

PR #335, opened 2022-09-27
Update CI workflows and add EDM4hep workflow

tmadlener:
BEGINRELEASENOTES
- Update the `github-action-cvmfs` and `run-lcg-view` actions to their latest available version to pick up the latest improvements (caching of dependencies, log groups)
- Introduce log groups in github actions for easier to interpret outputs
- Switch to LCG_102 for lcg based build environments
- Add a workflow that builds and tests EDM4hep after building podio

ENDRELEASENOTES

- [x] Should we also use this opportunity to update some of the environments to more recent LCG releases?
- [x] Should we also introduce (at least) one workflow that also builds EDM4hep@master to make sure we do not break it accidentally?

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -10,8 +10,8 @@ jobs:
     if: github.repository == 'AIDASoft/podio'
     steps:
     - uses: actions/checkout@v2
-    - uses: cvmfs-contrib/github-action-cvmfs@v2
-    - uses: aidasoft/run-lcg-view@v2
+    - uses: cvmfs-contrib/github-action-cvmfs@v3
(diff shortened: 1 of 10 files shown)


andresailer:
a) makes sense to update
b) sounds good as well

tmadlener:
I moved one ubuntu and one centos7 workflow based on LCG_102 to a dedicated `edm4hep` workflow, where podio is built and tested first, and afterwards EDM4hep is fetched, built and tested as well. Testing podio in these workflows makes it possible to remove them from their respective original workflows. Additionally this workflow builds Catch2 v3.1.0 first to avoid having to build it twice (for podio and EDM4hep).

PR #336, opened 2022-09-27
Cmake tweaks

andresailer:
Just some things I found slightly annoying

BEGINRELEASENOTES
- podioConfig.cmake: silence warning about cmake policy CMP00012
- CMake: explicitly look for catch2 version 3 and fail at cmake instead of compile step

ENDRELEASENOTES


tmadlener (approved):
Thanks.

PR #337, opened 2022-09-30
Make the notebook pattern use vectors instead of arrays

### LINKED ISSUE #332: Enhancement: vectorized accessor to variable-length full array (constexpr)
In the vectorized access (notebook) pattern, we are currently required to pass the number of entries we wish to retrieve, as a constexpr on compile time, e.g. [`unittest.cpp`](https://github.com/AIDASoft/podio/blob/v00-15/tests/unittest.cpp#L186)
```cpp
  auto energies = hits.energy<10>();
```
With C++20, we have now compile-time access to `hits.size()` with [`constexpr size_type std::vector::size() const noexcept;`](https://en.cppreference.com/w/cpp/container/vector/size) (though not for `deque`, which is used for pointers and is what `hits.size()` actually uses) which would allow something like
```cpp
  auto energies = hits.energy<hits.size()>();
```
At this point, why not take the next step and implement a default `arraysize` in `template<{{ member.full_type }}, arraysize = m_storage.entries.size()>`, resulting in the following (also) working:
```cpp
  auto energies = hits.energy();
```
This interface would seem to allow for some clarity in downstream code that may be beneficial.

Before going too far into a rabbit hole of template code:
- Using a `deque` for the object pointers (instead of a `vector`) may be unavoidable, but we may be able to use `size()` on the `m_data` instead of the `entries` in `CollectionData.h`, or even implement a `constexpr size_type CollectionData::size() const;`.
- How do you feel about C++20 features that require special handling (we can't just `decltype` the return type of `std::vector::size()` since that doesn't include the `constexpr`-ness)? Does this lead to a jinja2 branch for `cxxstd==20` or something?
- Is all this worth it? Ultimately we're still running an AoS loop to get to the SoA representation...

tmadlener:
BEGINRELEASENOTES
- Make the notebook pattern functionality return `std::vector`s instead of `std::array` to avoid having to specify a static size. Fixes #332 
- **Backwards incompatible change** as the return type as well as the call signature for the notebook pattern change.

ENDRELEASENOTES

@wdconinc very simple approach, but should in principle fix the original issue, I think.

### DIFF
--- a/doc/examples.md
+++ b/doc/examples.md
@@ -88,12 +88,14 @@ copy of only the data that are needed for a particular calculation. This
 pattern is supported by providing access like
 
 ```cpp
-    auto x_array = hits.x<10>(); // returning values of
-    auto y_array = hits.y<10>(); // the first 10 elements
+    auto x_array = hits.x();   // returning all values
+    auto y_array = hits.y(10); // or only the first 10 elements
 ```
 
-The resulting `std::array` can then be used in (auto-)vectorizable code.
-If less objects than requested are contained in the collection, the remaining entries are default initialized.
+The resulting `std::vector` can then be used in (auto-)vectorizable code.
+Passing in a size argument is optional; If no argument is passed all elements will be returned,
+if an argument is passed only as many elements as requested will be returned.
+If the collection holds less elements than are requested, only as elements as are available will be returned.
 
 ### EventStore functionality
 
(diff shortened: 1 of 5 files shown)


wdconinc:
Away for the long weekend here but will try to test on Monday.

tmadlener:
Ah, no worries. Monday is off in Germany as well ;)

wdconinc:
Looks good to me.

PR #340, opened 2022-10-21
Adding reading specific entry from frame

kjvbrt:
BEGINRELEASENOTES
- Adding reading of specific entry from frame
ENDRELEASENOTES

### DIFF
--- a/include/podio/ROOTFrameReader.h
+++ b/include/podio/ROOTFrameReader.h
@@ -60,6 +60,13 @@ public:
    */
   std::unique_ptr<podio::ROOTFrameData> readNextEntry(const std::string& name);
 
+  /**
+   * Read specific event. In case the frame can't be constructed or there is
+   * not such event, this returns a nullptr.
+   */
+  std::unique_ptr<podio::ROOTFrameData> readEntry(const std::string& name,
+                                                  const unsigned long entNum);
+
   /// Returns number of entries for the given name
   unsigned getEntries(const std::string& name) const;
 
--- a/include/podio/ROOTFrameReader.h
+++ b/include/podio/ROOTFrameReader.h
@@ -65,8 +65,7 @@ public:
    * the given name. In case the entry does not exist for this name or in case
    * there is no data for this name, this returns a nullptr.
    */
-  std::unique_ptr<podio::ROOTFrameData> readEntry(const std::string& name,
-                                                  const unsigned entry);
+  std::unique_ptr<podio::ROOTFrameData> readEntry(const std::string& name, const unsigned entry);
 
   /// Returns number of entries for the given name
   unsigned getEntries(const std::string& name) const;
@@ -112,8 +111,8 @@ private:
   GenericParameters readEventMetaData(CategoryInfo& catInfo);
 
   /**
-   * Read the data entry specified in the passed CategoryInfo, and increase the 
-   * counter aferwards. In case the requested entry is larger than the 
+   * Read the data entry specified in the passed CategoryInfo, and increase the
+   * counter aferwards. In case the requested entry is larger than the
    * available number of entries, return a nullptr.
    */
   std::unique_ptr<podio::ROOTFrameData> readEntry(ROOTFrameReader::CategoryInfo& catInfo);
--- a/include/podio/ROOTFrameReader.h
+++ b/include/podio/ROOTFrameReader.h
@@ -112,9 +112,9 @@ private:
   GenericParameters readEventMetaData(CategoryInfo& catInfo);
 
   /**
-   * Read the data entry from which a Frame can be constructed for the given
-   * name. In case the entry does not exist for this category, this returns a
-   * nullptr.
+   * Read the data entry specified in the passed CategoryInfo, and increase the 
+   * counter aferwards. In case the requested entry is larger than the 
+   * available number of entries, return a nullptr.
    */
   std::unique_ptr<podio::ROOTFrameData> readEntry(ROOTFrameReader::CategoryInfo& catInfo);
 
--- a/include/podio/ROOTFrameReader.h
+++ b/include/podio/ROOTFrameReader.h
@@ -61,11 +61,12 @@ public:
   std::unique_ptr<podio::ROOTFrameData> readNextEntry(const std::string& name);
 
   /**
-   * Read specific event. In case the frame can't be constructed or there is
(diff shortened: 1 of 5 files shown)


include/podio/ROOTFrameReader.h
```diff
@@ -103,6 +111,13 @@ class ROOTFrameReader {
 
   GenericParameters readEventMetaData(CategoryInfo& catInfo);
 
+  /**
+   * Read the data entry from which a Frame can be constructed for the given
+   * name. In case the entry does not exist for this category, this returns a
+   * nullptr.
```
tmadlener, line 117:
```suggestion
   * Read the data entry specified in the passed CategoryInfo, and increase the 
   * counter aferwards. In case the requested entry is larger than the 
   * available number of entries, return a nullptr.
```

to reflect what this actually does

src/ROOTFrameReader.cc
```diff
@@ -37,6 +37,22 @@ std::unique_ptr<ROOTFrameData> ROOTFrameReader::readNextEntry(const std::string&
...
+
+  return readEntry(catInfo);
+}
+
+std::unique_ptr<ROOTFrameData> ROOTFrameReader::readEntry(const std::string& name,
+                                                          const unsigned entNum) {
+  auto& catInfo = getCategoryInfo(name);
+  if (!catInfo.chain) {
+    return nullptr;
+  }
```
tmadlener, line 49: I think this check could also be moved to the private `readEntry` as that would avoid having it duplicated in two places.

tests/read_frame.h
```diff
@@ -52,6 +52,15 @@ int read_frames(const std::string& filename) {
...
   }
 
+  // Reading specific entry
+  {
+    auto frame = podio::Frame(reader.readEntry("events", 4));
+    processEvent(frame, 4, reader.currentFileVersion());
+
+    auto otherFrame = podio::Frame(reader.readEntry("other_events", 4));
+    processEvent(otherFrame, 4 + 100, reader.currentFileVersion());
+  }
```
tmadlener, line 62: Can you add two more tests here?
- Calling `readNextEntry` after a specific Frame has been read to "document" that it will continue from the previously read frame
- Calling `readEntry` again but with a smaller entry number to make sure that "jumping back" also works.

src/SIOFrameReader.cc
```diff
@@ -66,6 +66,12 @@ std::unique_ptr<SIOFrameData> SIOFrameReader::readNextEntry(const std::string& n
                                         tableInfo._uncompressed_length);
 }
 
+std::unique_ptr<SIOFrameData> SIOFrameReader::readEntry(const std::string& name, const unsingned entry);
+  m_nameCtr[name] = entry;
```
tmadlener, line 70: This is a very clever trick. Can you add a short comment that says that this will indeed work (in all cases) and that all the necessary checks are actually done in `readNextEntry`, because otherwise we will probably be wondering about this everytime we come across it.

src/SIOFrameReader.cc
```diff
@@ -66,6 +66,12 @@ std::unique_ptr<SIOFrameData> SIOFrameReader::readNextEntry(const std::string& n
                                         tableInfo._uncompressed_length);
 }
 
+std::unique_ptr<SIOFrameData> SIOFrameReader::readEntry(const std::string& name, const unsingned entry);
```
tmadlener, line 69:
```suggestion
std::unique_ptr<SIOFrameData> SIOFrameReader::readEntry(const std::string& name, const unsigned entry);
```
typo

include/podio/SIOFrameReader.h
```diff
@@ -32,6 +32,14 @@ class SIOFrameReader {
    */
   std::unique_ptr<podio::SIOFrameData> readNextEntry(const std::string& name);
 
+  /**
+   * Read the specified data entry from which a Frame can be constructed for
+   * the given name. In case the entry does not exist for this name or in
+   * case there is no data for this name, this returns a nullptr.
+   */
+  std::unique_ptr<podio::SIOFrameData> readEntry(const std::string& name
```
tmadlener, line 40:
```suggestion
  std::unique_ptr<podio::SIOFrameData> readEntry(const std::string& name,
```

tmadlener (commented):
Thanks @kjvbrt,

I have only a few minor comments inline.

src/ROOTFrameReader.cc
kjvbrt: Moved to private method

tests/read_frame.h
kjvbrt: Add :)

src/SIOFrameReader.cc
kjvbrt: fixed

include/podio/SIOFrameReader.h
kjvbrt: fixed

src/SIOFrameReader.cc
kjvbrt: Added a comment

src/ROOTFrameReader.cc
```diff
@@ -37,6 +37,22 @@ std::unique_ptr<ROOTFrameData> ROOTFrameReader::readNextEntry(const std::string&
   if (!catInfo.chain) {
     return nullptr;
   }
```
tmadlener, line 39:
```suggestion
```
This is a duplicate check now that it is handled by `readEntry`

tests/read_frame.h
```diff
@@ -52,6 +52,35 @@ int read_frames(const std::string& filename) {
...
+
+  // Reading next entry after jump should continue the sequence
+  if (!reader.readNextEntry("events")) {
+    std::cerr << "Trying to read next entry after jump, should read event 5 out of 10" << std::endl;
+    return 1;
+  }
```
tmadlener, line 68:
```suggestion
  {
    auto nextFrame = podio::Frame(reader.readNextEntry("events"));
    processEvent(nextFrame, 5, reader.currentFileVersion());
  }
```
I was thinking more along the lines of actually checking whether this is really the correct next event here. Similar for the other cases.

Having said that, a check that reading past the available entries indeed returns a `nullptr` also seems like a good idea to have here.

tmadlener:
`dev3` seems to have a failed installation at the moment and fails because no recent enough CMake version is available. Merging this regardless.

Issue #169, opened 2021-01-12
Possibility of more intelligible branch names for relations


tmadlener:
Currently `OneToManyRelations` and `OneToOneRelations` are simply stored with a `#{index}` post-fix, where the `index` runs from 0 to the number of total relations, starting with the `OneToManyRelations`. This makes it very hard to interpret the branches in the produced root output file. A similar problem exists for the `VectorMembers`, where the post fix is `_{index}`.

When working directly with the root files it might be nicer to have more intelligible branch names, e.g. something like `#{relation_name}`.

Opening this issue here to not forget about it and also to maybe gather feedback into what the best solution would be here. Currently we run the danger that people "circumventing" the EventStore rely upon the order of the relations, when implementing their solution, but this order is basically an implementation detail. Some people are already using this, so it might be good to either make it very explicit that this is indeed an implementation detail that can change without outside warning, or to make this part of the "API" by using more verbose branch names.

One of the open questions that would need to be answered first is whether it is actually possible to use other branch names at all. It is not yet entirely clear to me whether this is possible and how much work it would be to implement it.

andresailer:
Relates to #86

tmadlener:
Another point to consider is that `RDataFrame` needs column names that are valid c++ identifiers. Currently this is handled by something like
```python
(df.Alias("Muon0", "Muon#0.index")  # give this an an Alias that is a valid c++ identifier 
   .Define("doStuff(Muon0)") )      # now can use the branch
```

Issue #85, opened 2020-04-28
podio_class_generator generates strange names for VectorMembers


vvolkl:
Opened by @zoujh in EDM4hep: https://github.com/key4hep/EDM4hep/issues/54

PR #343, opened 2022-10-27
[WIP] Python bindings for Frame

tmadlener:
BEGINRELEASENOTES
- Add python bindings for `Frame` based I/O
  - Available from `podio.root_io` and `podio.sio_io`, where a `Reader` and a `Writer` is implemented for each.
- Add necessary functionality for python bindings to C++ API
  - untyped `Frame::get` method for getting collections
  - functionality to inspect file and `Frame` contents more easily
- Reorganize python code into structure that follows the usual python packaging conventions a bit more closely
  - Introduce the `podio` module. Make CMake generate the `__init__.py` with the correct version
  - Move everything except the generator script into `module`. Additionally also keep an `EventStore` wrapper to not break existing code. 
- Refactor the `CMakeLists.txt` that is responsible for building the core and all required I/O libraries

ENDRELEASENOTES

- [ ] Actual python bindings for the `Frame`
- [ ] more tests

## Details
These are a few implementation and technical details

### python `Reader` implementation / interface
Due to the essentially equivalent interfaces of the c++ readers, the readers are implemented via the `BaseReaderMixin` that essentially defines the complete interface. The backend specific readers simply have to instantiate the approriate c++ reader via PyROOT. The tests follow the same strategy via the `ReaderTestCaseMixin` base class.

The python version of the readers offers a slightly more pythonic access pattern to some of the functionality by some thin wrapper code. The following helper classes facilitate this
- `FrameCategoryIterator`: the iterator class that grants access to all `Frame`s of a given category, returned by `Reader.get`

### `CMakeLists.txt` refactoring
To enable python bindings more easily, all built shared libraries now have an accompanying dictionary, exposing the necessary c++ classes. To avoid even more code repetition I introduced a `PODIO_ADD_LIB_AND_DICT` function that handles most of the boilerplate.


python/podio/base_reader.py
```diff
@@ -0,0 +1,44 @@
...
+
+  def get(self, category):
+    """Get an iterator with access functionality for a given category.
+
+    Args:
+        category (str): The name of the desired category
+
+    Returns:
+      FrameCategoryIterator: The iterator granting access to all Frames of the
+        desired category
```
tmadlener, line 42:
```suggestion
        FrameCategoryIterator: The iterator granting access to all Frames of the
            desired category
```

python/podio/frame.py
```diff
@@ -0,0 +1,127 @@
...
+ROOT.gInterpreter.LoadFile('podio/Frame.h')  # noqa: E402
+from ROOT import podio  # noqa: E402 # pylint: disable=wrong-import-position
+
+
+def _determine_supported_parameter_types():
+  """Determine the supported types for the parameters.
+
+  Returns:
+      tuple (str): the tuple with the string representation of all **c++**
+        classes that are supported
```
tmadlener, line 18:
```suggestion
          classes that are supported
```

python/podio/frame.py
```diff
@@ -0,0 +1,127 @@
...
+
+    Returns:
+        dict: A dictionary mapping each key to the corresponding c++ type
+    """
+    params = self._frame.getGenericParametersForWrite()  # this is the iffy bit
+    keys_dict = {}
+    for par_type in SUPPORTED_PARAMETER_TYPES:
+      keys = params.getKeys[par_type]()
+      # Make sure to convert to a python string here to not have a dangling
+      # referenc here for the key. Also make the type an std::vector so that we
```
tmadlener, line 122:
```suggestion
      # reference here for the key. Also make the type an std::vector so that we
```

python/podio/frame.py
```diff
@@ -0,0 +1,127 @@
...
+      # always call the getter that obtains the full vector and we decide later
+      # in the python layer whether we return a single value or a list of values
+      keys_dict.update({str(k): f'std::vector<{par_type}>' for k in keys})
```
tmadlener, line 125: This will overwrite potentially already existing keys if there are parameters that have the same key but different types. Although this is not necessarily a good idea, it is possible (and valid) to do in the current implementation of the underlying `GenericParameters`

python/podio/test_Frame.py
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env python3
+"""Unit tests for python bindings of podio::Frame"""
+
+import unittest
+
+from podio.frame import Frame
+# using root_io as that should be present
```
tmadlener, line 7:
```suggestion
# using root_io as that should always be present regardless of which backends are built
```

python/podio/test_Reader.py
```diff
@@ -0,0 +1,67 @@
...
+    with self.assertRaises(IndexError):
+      _ = frames[-11]
+
+    # Again only rudimentary checks
+    frame = frames[7]
+    self.assertEqual(frame.get_parameter('UserEventName'), ' event_number_107')
+    # jumping back again also works
+    frame = frames[3]
+    self.assertEqual(frame.get_parameter('UserEventName'), ' event_number_103')
+
```
tmadlener, line 47: Here a valid negative indexed test case is missing.

tmadlener:
@wdconinc @gaede @andresailer @vvolkl @hegner any strong feelings on the python interface here? Anything obvious still missing?

wdconinc:
Does it make sense to include (for documentation purposes?) an example or test case with uproot, awkward, or numpy?

tmadlener:
> Does it make sense to include (for documentation purposes?) an example or test case with uproot, awkward, or numpy?

In principle I think it would. I am a bit hesitant to put it into documentation at the moment, because I am afraid, there will be a few changes still to the format of the root files, before we have a v1.0. E.g., one of the things that I would still like to tackle before that is, #169.

The one main issue I see with this, is that documenting this "officially" will tie podio very tightly into an expected file format for root. So I really want to be sure that things are stable, before we essentially make that a sort of interface as well.

wdconinc:
I agree with having a stable interface before 'advertising' this. Issues #169 and #85 linked therein are indeed both very relevant for podio usability in python.

PR #345, opened 2022-11-02
Add a Frame reader for legacy files

tmadlener:
BEGINRELEASENOTES
- Add a `ROOTLegacyReader` and a `SIOLegacyReader` that read files that have been written prior to #287 into `podio::Frame`s and offers the same interface as the frame readers
  - Also including python bindings for it

ENDRELEASENOTES

The implementations have been largely lifted from my git history, where a prototype for this has existed. The current version is essentially simply brushed up from there and with the same interface as the other frame readers. Given that the actual frame readers are largely based on these prototypes some of the internals are very similar, but at some points have to work around some issues introduced by the slightly different file formats.

- [x] Do we want / need an `SIOLegacyReader`?
- [x] needs #343 (relies on CMake refactoring and python binding basics)

### DIFF
--- a/include/podio/ROOTLegacyReader.h
+++ b/include/podio/ROOTLegacyReader.h
@@ -1,0 +1,125 @@
+#ifndef PODIO_ROOTLEGACYREADER_H
+#define PODIO_ROOTLEGACYREADER_H
+
+#include "podio/CollectionBranches.h"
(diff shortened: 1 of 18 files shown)


tmadlener:
merging this now as there are no conflicts with #341 and we want to exercise this for a software tutorial

PR #351, opened 2022-11-16
Fix a few smaller issues with Frame python bindings and podio-dump

tmadlener:
BEGINRELEASENOTES
-  Fix bug in Frame python bindings where empty collections were considered as non-existing. Replacing the original check relying on some implicit boolean conversions (which also caught empty collections) to an explicit check against `nullptr`.
- Make `podio-dump` more robust in installations without SIO support, by guarding the corresponding import.

ENDRELEASENOTES


hegner:
@tmadlener - would be great if you could not always self-merge here ;-)

PR #354, opened 2022-11-30
Make scripts POSIX compliant and fix mix of tabs and spaces

jmcarcell:
BEGINRELEASENOTES

- Make `env.sh` setup script POSIX compliant to run in shells other than bash
  - Change `==` to `=`
  - Change tabs to spaces (two) to avoid mix of spaces and tabs for indenting
  - Add `<prefix>/include` to `ROOT_INCLUDE_PATH` (as it is required since #343)

ENDRELEASENOTES

For the second one it's a bit of a matter of taste and I personally prefer spaces to tabs but it's true that there was a mix of tabs and spaces and it looked ugly, now it will be consistent in every editor.

### DIFF
--- a/env.sh
+++ b/env.sh
@@ -25,7 +25,7 @@ elif [[ "$unamestr" = 'Darwin' ]]; then
     fi
 fi
 
-if [ "$(echo $PYTHONPATH | grep -o $PODIO/python)" = "" ]; then
+if ! echo $PYTHONPATH | grep -o $PODIO/python > /dev/null 2>&1; then
   export PYTHONPATH=$PODIO/python:$PYTHONPATH
 fi
 
--- a/env.sh
+++ b/env.sh
@@ -9,8 +9,7 @@ fi
 
 unamestr=`uname`
 if [[ "$unamestr" = 'Linux' ]]; then
-  echo $LD_LIBRARY_PATH | grep $PODIO/lib >& /dev/null
-  if [ $? = "1" ]; then
+  if [ "$(echo $LD_LIBRARY_PATH | grep -o $PODIO/lib)" = "" ]; then
     # RedHat based put the libraries into lib64
     if [ -d $PODIO/lib64 ]; then
       export LD_LIBRARY_PATH=$PODIO/lib64:$LD_LIBRARY_PATH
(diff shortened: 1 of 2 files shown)


tmadlener:
Hi @jmarcell,
thanks for this. Since you are already working on this, could you also add `<prefix>/include` to the `ROOT_INCLUDE_PATH` in `env.sh`? This has become necessary with #343

env.sh
```diff
@@ -8,25 +8,27 @@ if [ -z "$PODIO" ]; then
 fi
 
 unamestr=`uname`
-if [[ "$unamestr" == 'Linux' ]]; then
-	echo $LD_LIBRARY_PATH | grep $PODIO/lib >& /dev/null
-	if [ $? == "1" ]; then
+if [[ "$unamestr" = 'Linux' ]]; then
+  if [ "$(echo $LD_LIBRARY_PATH | grep -o $PODIO/lib)" = "" ]; then
```
tmadlener, line 12:
```suggestion
  if ! echo $LD_LIBRARY_PATH | grep $PODIO/lib > /dev/null 2>&1; then
```
I think this is the most idiomatic (POSIX compliant) one liner to check this. It uses the non-zero exit code of `grep` if it doesn't find a match, and we simply redirect all output to `/dev/null` to not clutter the output. Also avoids creating a subshell.

env.sh
```diff
@@ -8,25 +8,27 @@ if [ -z "$PODIO" ]; then
...
-	fi
-elif [[ "$unamestr" == 'Darwin' ]]; then
+  fi
+elif [[ "$unamestr" = 'Darwin' ]]; then
   # This currenty does not work on OS X as DYLD_LIBRARY_PATH is ignored
   # in recent OS X versions
-	echo $DYLD_LIBRARY_PATH | grep $PODIO/lib >& /dev/null
-	if [ $? == "1" ]; then
-    	export DYLD_LIBRARY_PATH=$PODIO/lib:$DYLD_LIBRARY_PATH
+  if [ "$(echo $DYLD_LIBRARY_PATH | grep -o $PODIO/lib)" = "" ]; then
```
tmadlener, line 23:
```suggestion
  if ! echo $DYLD_LIBRARY_PATH | grep -o $PODIO/lib > /dev/null 2>&1; then
```

env.sh
```diff
@@ -8,25 +8,27 @@ if [ -z "$PODIO" ]; then
...
+      export DYLD_LIBRARY_PATH=$PODIO/lib:$DYLD_LIBRARY_PATH
     fi
 fi
-echo $PYTHONPATH | grep $PODIO/python >& /dev/null
-if [ $? == "1" ]; then
-	export PYTHONPATH=$PODIO/python:$PYTHONPATH
+
+if [ "$(echo $PYTHONPATH | grep -o $PODIO/python)" = "" ]; then
```
tmadlener, line 28:
```suggestion
if ! echo $PYTHONPATH | grep -o $PODIO/python > /dev/null 2>&1; then
```

env.sh
```diff
@@ -8,25 +8,27 @@ if [ -z "$PODIO" ]; then
...
+  export PYTHONPATH=$PODIO/python:$PYTHONPATH
+fi
+
+if [ "$(echo $ROOT_INCLUDE_PATH | grep -o $PODIO/include)" = "" ]; then
```
tmadlener, line 32:
```suggestion
if ! echo $ROOT_INCLUDE_PATH | grep -o $PODIO/include > /dev/null 2>&1; then
```

tmadlener (commented):
Thanks for taking care of the `ROOT_INCLUDE_PATH` additions. Putting the `grep` checks into one line is also a very good idea. For those I have suggested a slightly different way, see inline.

tmadlener (approved):

PR #355, opened 2022-12-01
Deal with missing files in the readers

### LINKED ISSUE #353: podio readers do not handle non-existant files gracefully
None of the currently existing readers handle non-existant files gracefully. All root related readers simply segfault, all sio related readers throw an `sio::exception` which is never caught.

For root the output looks similar to this:
```
Error in <TFile::TFile>: file non-existant.root does not exist

 *** Break *** segmentation violation
[... potentially a complete backtrace ...]
```

For sio the output looks similar to this
```
terminate called after throwing an instance of 'sio::exception'
  what():  <path/to/sio-build-source>/src/api.cc (l.72) in read_record_info: ifstream is not open! [not_open
```

I think both of these should be handled better.

jmcarcell:
Fix #353. The readers now check if the files exist before opening them (checked using the new C++17 std library `filesystem`) and I added a warning message that tells you that the file doesn't exist. As it is now it will be the responsibility of the user to check that the opening of the files didn't go well before doing anything else. For the readers that can multiple files it should work if any of the files is readable even if there are some that do not exist or are missing but I haven't tested that.

BEGINRELEASENOTES
- Change the readers so that when the file is missing they won't crash

ENDRELEASENOTES

This is the output of the tests when the file is missing:

```
$ tests/read                 
Warning: file exampl.root does not exist.
No files could be found...

$ tests/read_frame       
Warning: file exampl_frame.root does not exist 
No files could be found...
The podio build version could not be read back correctly. (expected:0.16.0, actual: 0.0.0)

$ tests/read_frame_legacy_root
Warning: file exampl.root does not exist.
No files could be found...
The podio build version could not be read back correctly. (expected:0.16.0, actual: 0.0.0)

$ tests/read_sio              
Warning: file example.sio does not exist

$ tests/read_frame_sio        
Warning: file example_frame.sio does not exist 
The podio build version could not be read back correctly. (expected:0.16.0, actual: 0.0.0)

$  tests/read_frame_legacy_sio 
Warning: file exampl.sio does not exist 
The podio build version could not be read back correctly. (expected:0.16.0, actual: 0.0.0)
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,6 +67,9 @@ ADD_CLANG_TIDY()
 #--- Declare options -----------------------------------------------------------
 option(CREATE_DOC "Whether or not to create doxygen doc target." OFF)
 option(ENABLE_SIO "Build SIO I/O support" OFF)
+if (ENABLE_SIO)
+  add_compile_definitions(PODIO_WITH_SIO)
+endif()
 
 
 #--- Declare ROOT dependency ---------------------------------------------------
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,9 +67,6 @@ ADD_CLANG_TIDY()
 #--- Declare options -----------------------------------------------------------
 option(CREATE_DOC "Whether or not to create doxygen doc target." OFF)
 option(ENABLE_SIO "Build SIO I/O support" OFF)
-if (ENABLE_SIO)
-  add_compile_definitions(PODIO_WITH_SIO)
-endif()
 
 
 #--- Declare ROOT dependency ---------------------------------------------------
--- a/src/ROOTFrameReader.cc
+++ b/src/ROOTFrameReader.cc
@@ -13,6 +13,7 @@
 #include "TTreeCache.h"
 
 #include <unordered_map>
+#include <filesystem>
 
 namespace podio {
 
@@ -197,6 +198,19 @@ void ROOTFrameReader::openFile(const std::string& filename) {
 }
 
 void ROOTFrameReader::openFiles(const std::vector<std::string>& filenames) {
+  bool any_file_exists = false;
+  for (const auto& fn : filenames) {
+    if (!std::filesystem::exists(fn)) {
+      std::cout << "Warning: file " << fn << " does not exist " << std::endl;
+    }
+    else {
+      any_file_exists = true;
+    }
+  }
+  if (!any_file_exists) {
+    std::cout << "No files could be found..." << std::endl;
+    return;
+  }
   m_metaChain = std::make_unique<TChain>(root_utils::metaTreeName);
   // NOTE: We simply assume that the meta data doesn't change throughout the
   // chain! This essentially boils down to the assumption that all files that
--- a/src/ROOTFrameReader.cc
+++ b/src/ROOTFrameReader.cc
@@ -206,9 +206,6 @@ void ROOTFrameReader::openFiles(const std::vector<std::string>& filenames) {
     if(!m_metaChain->Add(filename.c_str(), -1)) {
       throw std::runtime_error("File " + filename + " couldn't be found");
     }
-    else {
-      break;
-    }
   }
 
   podio::version::Version* versionPtr{nullptr};
--- a/src/ROOTFrameReader.cc
+++ b/src/ROOTFrameReader.cc
@@ -12,8 +12,8 @@
 #include "TTree.h"
 #include "TTreeCache.h"
 
-#include <unordered_map>
 #include <stdexcept>
+#include <unordered_map>
 
 namespace podio {
 
@@ -204,7 +204,7 @@ void ROOTFrameReader::openFiles(const std::vector<std::string>& filenames) {
   // are read this way were written with the same settings.
   // Reading all files is done to check that all file exists
   for (const auto& filename : filenames) {
-    if(!m_metaChain->Add(filename.c_str(), -1)) {
+    if (!m_metaChain->Add(filename.c_str(), -1)) {
       throw std::runtime_error("File " + filename + " couldn't be found");
     }
   }
--- a/src/ROOTFrameReader.cc
+++ b/src/ROOTFrameReader.cc
@@ -202,6 +202,7 @@ void ROOTFrameReader::openFiles(const std::vector<std::string>& filenames) {
   // NOTE: We simply assume that the meta data doesn't change throughout the
   // chain! This essentially boils down to the assumption that all files that
   // are read this way were written with the same settings.
+  // Reading all files is done to check that all file exists
   for (const auto& filename : filenames) {
     if(!m_metaChain->Add(filename.c_str(), -1)) {
       throw std::runtime_error("File " + filename + " couldn't be found");
--- a/src/ROOTFrameReader.cc
+++ b/src/ROOTFrameReader.cc
@@ -13,7 +13,6 @@
 #include "TTreeCache.h"
 
 #include <unordered_map>
-#include <filesystem>
 
 namespace podio {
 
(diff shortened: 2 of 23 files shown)


tmadlener:
Very nice approach with using `filesystem`. I think in the case of root we could definitely get away without it, as `TChain::Add` can already tell us whether a file is usable in this context [as per the documentation](https://root.cern.ch/doc/master/classTChain.html#a78a896924ac6c7d3691b7e013bcbfb1c)
```cpp
auto result = m_chain->Add(filename, -1);
if (!result) {
  // file not present (or at least not with the expected TTree)
}
```
Potentially also for SIO we don't need to get `filesystem` involved here, because, technically the `m_stream` can be queried to check whether it is actually open and in a valid state.

------------ 

On a bigger picture, there will need to be a few more (small) changes here, most importantly the `openFile{s}` methods need to at least have a `bool` return type to signal to the outside that something went wrong. Simply returning early gives us a slightly better error message, but no real way to prevent crashes yet. With that it should also become much easier to add simple unittests for these failure cases.

In slightly more detail:
- For the `SIOReader` and `ROOTReader` you will also have to adapt the `IReader` interface to make `openFile` return a `bool`. For all the others there is no common interface.
- For the tests you should be able to add them here: https://github.com/AIDASoft/podio/blob/master/tests/unittest.cpp . You can probably use a [`TEMPLATE_LIST_TEST_CASE`](https://github.com/catchorg/Catch2/blob/devel/docs/test-cases-and-sections.md#type-parametrised-test-cases), similar to what is used, e.g. [here](https://github.com/key4hep/EDM4hep/blob/master/test/utils/test_kinematics.cpp#L15-L17) to cover all of the different readers without having to duplicate the test for each of them.

gaede:
I think it would be safer if `openFiles(...)` would fail if any one of the files specified does not exist. A warning is  just too easily missed in a larger production environment...

tmadlener:
In principle we could mark the `openFile{s}` return type as `[[nodiscard]]`, that would at least give users a warning (unfortunately no hard error) during compilation that they should check the return type in this case.

Overall, I have no strong feelings about throwing on error, or returning something indicating an error either. The main concern here is that we do it consistently for all of podio, at least for the public interface. Unfortunately, we are a bit all over the place there at the moment. (As also exhibited by the fact that some of the SIO readers throw, while others do not, ...)

I think large production use cases can be handled in either case, and either case actually requires the implementor of whatever podio is used for to either check a return value and handle that accordingly, or to handle the exception appropriately. The latter here has the advantage that it will automatically terminate if it is not handled. However, also that is simply a crashing application, where users will have to check the outputs to see where the exception comes from.

Maybe for a quick fix, simply throwing in all cases is the best solution for the moment, until we have a decision on how we want to do error handling in podio in general. For the moment `std::runtime_error` is probably enough here.

@hegner do you have any strong opinions on this? Short term for now, longer term might require a dedicated discussion / issue.

tmadlener:
One disadvantage of throwing exceptions is that we cannot easily handle them from the python side without crashing, I think. So that would potentially require another layer for python interoperability.

hegner:
The question is whether that is a public public interface or an internal interface for us. If we want users to use this API, things are more complicated. If we do only do a boolean return value nobody guarantees that it is being checked and then we would need to check again at the place where the file content is actually accessed as not to run into other exceptions down the chain.
So exception would be the way to go here IMHO if we assume other people besides us are using this interface

tmadlener:
> then we would need to check again at the place where the file content is actually accessed as not to run into other exceptions down the chain. 

Technically we start reading from the file(s) directly in `openFile{s}`. So in that regard, we are in principle guaranteed to have a valid file if that succeeds.

Given that this is something that is used by users, throwing an exception here sounds like the way to go. In that case `openFiles` should probably through if **any** of the passed files does not exist.

hegner:
agreed

src/ROOTFrameReader.cc
```diff
@@ -201,7 +202,11 @@ void ROOTFrameReader::openFiles(const std::vector<std::string>& filenames) {
   // NOTE: We simply assume that the meta data doesn't change throughout the
   // chain! This essentially boils down to the assumption that all files that
   // are read this way were written with the same settings.
```
tmadlener, line 204: Could you add a sentence here, that adding to the chain is essentially only done for checking the existence of the files and not really for a consistency check.

src/SIOFrameReader.cc
```diff
@@ -17,7 +17,8 @@ SIOFrameReader::SIOFrameReader() {
 void SIOFrameReader::openFile(const std::string& filename) {
   m_stream.open(filename, std::ios::binary);
   if (!m_stream.is_open()) {
-    SIO_THROW(sio::error_code::not_open, "Cannot open input file '" + filename + "' for reading");
+    // SIO_THROW(sio::error_code::not_open, "Cannot open input file '" + filename + "' for reading");
```
tmadlener, line 20:
```suggestion
```

src/SIOLegacyReader.cc
```diff
@@ -16,7 +16,8 @@ SIOLegacyReader::SIOLegacyReader() {
 void SIOLegacyReader::openFile(const std::string& filename) {
   m_stream.open(filename, std::ios::binary);
   if (!m_stream.is_open()) {
-    SIO_THROW(sio::error_code::not_open, "Cannot open input file '" + filename + "' for reading");
+    throw std::runtime_error("File " + filename + " couldn't be opened");
+    // SIO_THROW(sio::error_code::not_open, "Cannot open input file '" + filename + "' for reading");
```
tmadlener, line 20:
```suggestion
```

tmadlener (commented):
This looks good to me. I am not entirely sure whether we need the `try`/ `catch` blocks in our I/O tests, but they also serve as examples there. Just to be on the very safe side could you also add a unittest that verifies that the exception is indeed thrown when we pass in non-existant files?

Our style checks complain about a few formatting issues, and a few superfluous whitespaces:
https://github.com/AIDASoft/podio/actions/runs/3656635159/jobs/6179297741#step:4:330
https://github.com/AIDASoft/podio/actions/runs/3656635159/jobs/6179297741#step:4:431

CMakeLists.txt
```diff
@@ -67,6 +67,9 @@ ADD_CLANG_TIDY()
 #--- Declare options -----------------------------------------------------------
 option(CREATE_DOC "Whether or not to create doxygen doc target." OFF)
 option(ENABLE_SIO "Build SIO I/O support" OFF)
+if (ENABLE_SIO)
+  add_compile_definitions(PODIO_WITH_SIO)
+endif()
```
tmadlener, line 72: Can we move this to the `tests/CMakeLists.txt` or even better as a simple `target_compile_definitions` to the unittests? Otherwise we "pollute" all of our targets with this definition, and I would like to avoid that (at least for the moment).

tests/CMakeLists.txt
```diff
@@ -186,7 +186,10 @@ configure_file(CTestCustom.cmake ${CMAKE_BINARY_DIR}/CTestCustom.cmake)
 
 find_package(Threads REQUIRED)
 add_executable(unittest unittest.cpp frame.cpp)
-target_link_libraries(unittest PUBLIC TestDataModel PRIVATE Catch2::Catch2WithMain Threads::Threads)
+target_link_libraries(unittest PUBLIC TestDataModel PRIVATE Catch2::Catch2WithMain Threads::Threads podio::podioRootIO)
+if (ENABLE_SIO)
+  target_link_libraries(unittest PUBLIC TestDataModel PRIVATE Catch2::Catch2WithMain Threads::Threads podio::podioRootIO SIO::sio podio::podioSioIO)
```
tmadlener, line 191:
```suggestion
  target_link_libraries(unittest PRIVATE podio::podioSioIO)
```
I think this should pull in all the necessary dependencies, without repeating things that are already there from the call above

tests/unittest.cpp
```diff
@@ -979,6 +987,66 @@ TEST_CASE("GenericParameters return types", "[generic-parameters][static-checks]
...
 
+TEST_CASE("Missing files (ROOT readers)", "[basics]") {
+  bool exc = false;
+  auto root_reader = podio::ROOTReader();
+  try {
+    root_reader.openFile("NonExistentFile.root");
+  } catch (const std::runtime_error& e) {
+    exc = true;
+  }
+  REQUIRE(exc);
```
tmadlener, line 998:
```suggestion
  auto root_reader = podio::ROOTReader();
  REQUIRE_THROWS(root_reader.openFile("NonExistantFile.root"));
```
This should work (if I read the [documentation](https://github.com/catchorg/Catch2/blob/devel/docs/assertions.md#exceptions) correctly) and is much cleaner, IMHO. In principle one could even make sure that the exception message indeed contains something pointing to a non-existant file, but I don't think that is necessary.

tmadlener (commented):
Thanks for adding the unittests. I only have a minor issue with adding the `PODIO_USE_SIO` definition globally, other than that it is mainly simplifying the tests by using a corresponding catch2 macro.

tests/CMakeLists.txt
```diff
@@ -186,7 +186,10 @@ configure_file(CTestCustom.cmake ${CMAKE_BINARY_DIR}/CTestCustom.cmake)
 
 find_package(Threads REQUIRED)
 add_executable(unittest unittest.cpp frame.cpp)
-target_link_libraries(unittest PUBLIC TestDataModel PRIVATE Catch2::Catch2WithMain Threads::Threads)
+target_link_libraries(unittest PUBLIC TestDataModel PRIVATE Catch2::Catch2WithMain Threads::Threads podio::podioRootIO)
+if (ENABLE_SIO)
+  target_link_libraries(unittest PRIVATE podio::podioSioIO)
+endif()
```
tmadlener, line 192:
```suggestion
  target_compile_definitions(unittest PRIVATE PODIO_WITH_SIO)
endif()
```
To also enable the test for SIO.

tmadlener (approved):
Thanks again for this.

PR #356, opened 2022-12-04
Fix path in the README and simplify option parsing

jmcarcell:
BEGINRELEASENOTES
- Fix path in the README
- Use the functionality in argparse to choose between options

ENDRELEASENOTES

Now using `--help` will tell us about the possible choices


README.md
```diff
@@ -154,7 +154,7 @@ To run the code generation script, do
     mkdir ../Tmp
     mkdir ../Tmp/data
     mkdir ../Tmp/src
```
tmadlener, line 156:
```suggestion
```

This is also handled by the generator script:
https://github.com/AIDASoft/podio/blob/34ca51e1fc0bcbbe9d594fa66edb8acf8484454f/python/podio_class_generator.py#L533-L536

tmadlener (commented):
Oh that is very nice. I wasn't aware of `choices` in `argparse`.

I really just have a very minor comment for the README, where we technically do not have to create all the subdirectories.

tmadlener (approved):

PR #358, opened 2022-12-09
Store the model definition into files that are written

tmadlener:
BEGINRELEASENOTES
- Embed the EDM definition in JSON format into the shared core datamodel libraries
  -  Generate an additional `DatamodelDefinition.h` header file containing the string literal json encoded definition
  - Statically register this to the newly introduced `EDMDefinitionRegistry` and make collections aware of which datamodel they belong to
- Collect all EDM definitions from all collections that are written with a writer and write all these definitions to the resulting file
  - Currently only done for the `FrameWriter`s
- Give `podio-dump` the necessary functionality to retrieve the stored models and dump them in YAML format again
  - Add roundtrip tests that compare the generated code from the original model and the one that has been dumped from a data file to ensure that all components work as intended.
- See the [advanced topics documentation](https://github.com/tmadlener/podio/blob/store-model-def/doc/advanced_topics.md#retrieving-the-edm-definition-from-a-data-file) for more details.  

ENDRELEASENOTES

This is a first, incomplete, draft of storing the model definition into the written files, such that they can be retrieved again later such that the datamodel(s) of the types stored in the file can be regenerated. This is mainly here to gather some early feedback before investing too much time into the whole thing. @hegner, @gaede, @andresailer please let me know if there are any serious concerns with the overall approach, details can then be hashed out later, I suppose.

The general idea is:
- Store the model definition as a JSON string somewhere in the metadata of the file.
- Provide some standalone tools (or make the class generator or `podio-dump` capable of) either generating the code directly from a file, or simply dumping the yaml definition again.

The JSON string is a raw string literal that we simply generate via the code generator. In order to get this to the files, we also introduce a `EDMDefinitionRegistry` singleton, where we make each datamodel statically register itself. While writing we simply query collections about which definition we need to write in the end, and we collect a list of all defintions that need writing, and write them once everything is done.

The reason for JSON as storage "format" is because it yields relatively compact strings  that need to be stored (especially compared to trying to store the raw YAML files). Additionally, converting between JSON and YAML in python is pretty trivial in the end (and only necessary if we want to dump a YAML file again).

TODO list:
- [x] Make it possible to dump (parsed) datamodel definitions to JSON
- [x] code generation of compile time raw string literal
- [x] `EDMDefinitionRegistry` singleton
- [x] static registration of the JSON definition strings in the registry
- [x] Make collections query-able about which definition they need 
- [x] Instrument writers to collect and write the necessary definitions
- [x] Figure out the best way to handle upstream EDMs in this scenario
- [x] Tooling to retrieve definitions again from written files
- [x] (roundtrip) tests
- [x] (developers) documentation
- [x] release notes
- [x] Needs #361 
- [x] Needs #363


tmadlener:
Moved some of the general refactoring / code cleanup into #363 to make this diff slightly less cluttered.

include/podio/SIOBlock.h
```diff
@@ -141,6 +171,32 @@ class SIOEventMetaDataBlock : public sio::block {
...
+  SIOMapBlock(const SIOMapBlock&) = delete;
+  SIOMapBlock& operator=(const SIOMapBlock&) = delete;
+
+  void read(sio::read_device& device, sio::version_type) override {
+    readMapLike(device, mapData);
+  }
+  void write(sio::write_device& device) override {
+    writeMapLike(device, mapData);
+  }
+
```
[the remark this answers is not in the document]
tmadlener: As it is currently used it needs to be public, because the `SIOFrameReader` steals this directly via public access.

doc/advanced_topics.md
```diff
@@ -110,3 +110,118 @@ To implement your own transient event store, the only requirement is to set the
...
+the same as the original EDM definition.** E.g. all the datatypes will have all
+their fields (`Members`, `OneToOneRelations`, `OneToManyRelations`,
+`VectorMembers`) defined, and defaulted to empty lists in case they were not
+present in the original EDM definition. The reason for this is that the embedded
+EDM definition is the pre-processed and validated one [as described
+below](#technical-details-on-edm-definition-embedding)
+
+### Accessing the EDM definition programmatically
+The EDM definition can also be accessed programmatically via the
+`[ROOT|SIO]FrameReader::getEDMDefinition` method. It takes an EDM name as its
```
hegner, line 154: We have to be a bit more careful in how we name things. There is the EDM (the definition of all data you ever want to store) and the definition of the individual data types. I think we should clean up the interfaces to talk about datatype definitions and not EDM definitions if we are talking about single types / collections. That need for check holds for all what contains `EDM` at the moment.

include/podio/EDMDefinitionRegistry.h
```diff
@@ -0,0 +1,62 @@
+#ifndef PODIO_EDMDEFINITIONREGISTRY_H
+#define PODIO_EDMDEFINITIONREGISTRY_H
+
+#include <string>
+#include <string_view>
+#include <utility>
+#include <vector>
+
+namespace podio {
+class EDMDefinitionRegistry {
```
hegner, line 10: The registry should be a "DatatypeRegistry" which deals with knowing about definitions and how to instantiate classes after just passing a name to it (needed for all non-ROOT cases). I would not like having multiple registries around.

include/podio/ROOTFrameReader.h
```diff
@@ -40,7 +41,7 @@ struct CollectionReadBuffers;
  * This class has the function to read available data from disk
  * and to prepare collections and buffers.
  **/
-class ROOTFrameReader {
+class ROOTFrameReader : public EDMDefinitionHolder {
```
hegner, line 44: ) I would not use mixins in this case. I do not see a reason where and why the `EDMDefinitionCollector` should be exposed as an interface. It can stay internal. For the other one. I wouldn't do it either.

python/templates/DatamodelDefinition.h.jinja2
```diff
@@ -0,0 +1,30 @@
+// AUTOMATICALLY GENERATED FILE - DO NOT EDIT
+
+#include "podio/EDMDefinitionRegistry.h"
+
+namespace {{ package_name }}::meta {
+/**
+ * The complete definition of the datamodel at generation time in JSON format.
+ */
+static constexpr auto {{ package_name }}__JSONDefinition = R"EDMDEFINITION({{ edm_definition }})EDMDEFINITION";
```
hegner, line 9: The handling of referenced-to types not present in the file is incomplete. As of now we would end up with not consistent yaml files if type B (pointing to A) is in the file, while type A is missing.
We need to find a way for how to insert it nicely.

hegner (changes requested):
@tmadlener - thanks for this great implementation. 
There are a few things which need changing here. See my comments.

include/podio/EDMDefinitionRegistry.h
tmadlener: Yes, I agree the name is not the best here. Having one single registry would be nice, but it potentially becomes rather large, depending on what we want to do with it in the end:
- Getting the JSON encoded yaml file
- Creating collection buffers from a type name (and potentially a version)
- Getting the names of the relations of a given data type (would be nice to address #169)
- Do we also want schema evolution hooked into this? I.e. get an evolution function given a type name, and a buffer version (and potentially a backend)?

I think the first three can easily go into this one, I am not entirely sure about the schema evolution yet, but maybe we can see how that works out when we get there.

hegner: I think we should go for the first 3 here you list.
For the schema evolution let's discuss offline.

include/podio/ROOTFrameReader.h
tmadlener: This is changed to internal members now.

tmadlener:
Brief intermediate update: I have renamed the registry to `DatamodelRegistry` since it holds information about datamodels in general. I have also tried to be more clear about what is a *datamodel* and what is an *EDM* and renamed methods and classes accordingly. Essentially now an EDM refers to the compiled library of a datamodel only, whereas the datamodel is everything related to one model, i.e. it's definition in YAML format as well as other information. In the API now everything refers to datamodels unless it is specifically about something more concrete. I still have to update the markdown documentation to include these changes and definitions, and I also still have to to tackle the issue of dependent datamodels.

include/podio/EDMDefinitionRegistry.h
tmadlener: I think that makes sense. I would still defer the other points to separate PRs to not let this grow too large. The main thing to address in this PR is the naming of the Registry, which should now be general enough to allow for adding the other two functionalities as well without causing confusion.

hegner:
@tmadlener - thanks for the updates - in particular as some of them were quite tedious...
I would open a ticket for the dependent datamodels and just fix the documentation for now.

tmadlener:
Alright, documentation is updated, and a final leftover header has also been renamed to match the new terminology.

tmadlener:
I think now I really have all the things that needed to be renamed ;)

The two workflows that are currently failing, are due to failing upstream nightly builds (the `setup.sh` script is missing in both cases, causing the workflow to fail)

tmadlener:
all workflows are working again and passing

hegner:
@tmadlener - GREAT! :-)

tmadlener:
Is there anything still left to do on this one? I would open the issue for the missing correct handling of dependent datamodels once this is merged. I think I addressed the rest here.

hegner (approved):

PR #362, opened 2022-12-23
Rename match -> result for python > 3.10

jmcarcell:
BEGINRELEASENOTES
- Rename the variable `match` to avoid collisions with a python keyword from Python 3.10 onwards

ENDRELEASENOTES

[PEP 634](https://peps.python.org/pep-0634/) introduces match statements so `match` shouldn't be used as a variable name in a similar way to how `klass` is being used to avoid using they keyword `class`. Another reason to change them is that some editors may highlight these variables because it's a keyword.


tmadlener (approved):

tmadlener:
Thanks.

PR #364, opened 2023-01-03
Make workflows not trigger twice on pushes to PRs

jmcarcell:
BEGINRELEASENOTES
- Make workflows not trigger twice on pushes to PRs

ENDRELEASENOTES
by adding a rule to only trigger on pushes to the master branch. With the current behaviour, workflows trigger twice when someone pushes to their branch and they have a PR open, now the workflows will only trigger once in that case, saving some minutes. Workflows won't trigger on pushes to branches in the main repo but opening a PR will solve that and I think there aren't that many branches where that would be useful (?)

### DIFF
--- a/.github/workflows/edm4hep.yaml
+++ b/.github/workflows/edm4hep.yaml
@@ -1,6 +1,11 @@
 name: edm4hep
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+    - master
+  pull_request:
+  workflow_dispatch:
 
 jobs:
   build-and-test:
--- a/.github/workflows/key4hep.yml
+++ b/.github/workflows/key4hep.yml
@@ -1,6 +1,12 @@
 name: key4hep
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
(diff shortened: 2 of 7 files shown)


tmadlener:
I think this would no longer run CI if a user pushes to a branch without a PR on their own repository, right? Is there a way to at least have the possibility to run CI on my own repository if I have no PR associated to a branch? That would make it a lot easier to also run the sanitizer builds, which can be a bit of work to setup locally.

tmadlener:
Thinking about it maybe the `workflow_dispatch` would be nice to not trigger twice for open PRs. Could you add that, and also add a sentence or two to the README on how one can trigger these manually?

tmadlener (approved):
Thanks a lot.

I think running on the last commit of a PR (i.e. as part of the PR) and then running again once everything is merged on `master` is actually a feature, rather than an issue. We try to keep a somewhat linear history for podio, but there are still cases where changes from another PR have not yet run with the ones from the current PR, so running again on `master` / `main` is really what should happen.

PR #368, opened 2023-01-25
Add option to allow relaxing the required match of python versions

### LINKED ISSUE #366: Required match of Python versions?
https://github.com/AIDASoft/podio/blob/52064ba2217f45c27eb992f2b295d4be1508fa52/CMakeLists.txt#L93
This line requires the Python version used when compiling ROOT and when building podio to be the same but this is not always needed. For example, in my linux distribution I get ROOT compiled with Python 3.10.8 but I get Python 3.10.9 in the system so running `cmake` fails, although if I relax the rule it compiles and runs fine. What about relaxing the rule so that the patch number (major.minor.patch) is not required to match exactly and a warning is displayed in the case that they don't match? I volunteer for a PR

jmcarcell:
BEGINRELEASENOTES
- CMAKE: Add option PODIO_RELAX_PYVER to allow relaxing the required match of python version with the one that ROOT has been built with to only check major and minor versions

ENDRELEASENOTES

`find_package` will let everything go through but a later check will only allow mismatches in the patch version and throw and error if there is a mismatch in the major or minor versions. Does it make sense to have it ON by default since it's only the patch version? The error message looks like this:
```
CMake Error at CMakeLists.txt:102 (message):
  There is a mismatch between the major and minor versions in Python (found
  3.10) and ROOT, compiled with Python 3.9
```
in red, and stops `cmake`.

<strike>Also a few changes to lower case to be more consistent across the file</strike>

Closes #366

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,8 +65,9 @@ option(CLANG_TIDY "Run clang-tidy after compilation." OFF)
 ADD_CLANG_TIDY()
 
 #--- Declare options -----------------------------------------------------------
-option(CREATE_DOC "Whether or not to create doxygen doc target." OFF)
-option(ENABLE_SIO "Build SIO I/O support" OFF)
+option(CREATE_DOC        "Whether or not to create doxygen doc target." OFF)
+option(ENABLE_SIO        "Build SIO I/O support" OFF)
+option(PODIO_RELAX_PYVER "Do not require exact python version match with ROOT" OFF)
 
 
 #--- Declare ROOT dependency ---------------------------------------------------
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -80,20 +81,33 @@ if (NOT "cxx_std_17" IN_LIST ROOT_COMPILE_FEATURES)
   message(FATAL_ERROR "You are trying to build podio against a version of ROOT that has not been built with a sufficient c++ standard. podio requires c++17")
 endif()
 #Check if Python version detected matches the version used to build ROOT
-SET(Python_FIND_FRAMEWORK LAST)
+set(Python_FIND_FRAMEWORK LAST)
```
andresailer, line 84: Please no gratuitous changes like these

CMakeLists.txt
```diff
@@ -65,8 +65,9 @@ option(CLANG_TIDY "Run clang-tidy after compilation." OFF)
 ADD_CLANG_TIDY()
 
 #--- Declare options -----------------------------------------------------------
-option(CREATE_DOC "Whether or not to create doxygen doc target." OFF)
-option(ENABLE_SIO "Build SIO I/O support" OFF)
+option(CREATE_DOC        "Whether or not to create doxygen doc target." OFF)
+option(ENABLE_SIO        "Build SIO I/O support" OFF)
+option(PODIO_RELAX_PYVER "Do not require exact python version match with ROOT" ON)
```
andresailer, line 70:
```suggestion
option(PODIO_RELAX_PYVER "Do not require exact python version match with ROOT" OFF)
```

CMakeLists.txt
tmadlener: Yes, please revert for this PR. We can fix the inconsistencies in a different PR that only fixes that. Otherwise we will have commits that are mixing format / style changes and functional changes, which makes it hard to analyze the history.

tmadlener (approved):

PR #372, opened 2023-02-10
Add `double` to the supported types of `GenericParameters`

tmadlener:
BEGINRELEASENOTES
- Make `double` a supported type of `GenericParameters`. A similar thing has been added to LCIO in [iLCSoft/LCIO#143](https://github.com/iLCSoft/LCIO/pull/143) to support storing event weights that need double precision.
- Add more unittests to the `GenericParameters` covering also the available constructors.

ENDRELEASENOTES

The majority of the work was necessary for the python bindings, since python only has `float` (which is double precision), but we need to be able to disambiguate between `float` and `double` in order to call the correct overload for retrieving the values.

### DIFF
--- a/include/podio/GenericParameters.h
+++ b/include/podio/GenericParameters.h
@@ -17,7 +17,7 @@
 namespace podio {
 
 /// The types which are supported in the GenericParameters
-using SupportedGenericDataTypes = std::tuple<int, float, std::string>;
+using SupportedGenericDataTypes = std::tuple<int, float, std::string, double>;
 
 /// Static bool for determining if a type T is a supported GenericParamter type
 template <typename T>
@@ -79,6 +79,7 @@ public:
 private:
   using IntMap = MapType<int>;
   using FloatMap = MapType<float>;
+  using DoubleMap = MapType<double>;
   using StringMap = MapType<std::string>;
   // need mutex pointers for having the possibility to copy/move GenericParameters
   using MutexPtr = std::unique_ptr<std::mutex>;
@@ -110,6 +111,11 @@ public:
   /// Overload for catching const char* setting for string values
   void setValue(const std::string& key, std::string value) {
     setValue<std::string>(key, std::move(value));
+  }
+
+  /// Overlaod for catching initializer list setting of string vector values
+  void setValue(const std::string& key, std::vector<std::string> values) {
+    setValue<std::vector<std::string>>(key, std::move(values));
   }
 
   /// Overload for catching initializer list setting for vector values
@@ -224,6 +230,16 @@ public:
   }
 
   /**
+   * Get the internal double map (necessary for serialization with SIO)
+   */
+  const DoubleMap& getDoubleMap() const {
+    return getMap<double>();
+  }
+  DoubleMap& getDoubleMap() {
+    return getMap<double>();
+  }
+
+  /**
    * Get the internal string map (necessary for serialization with SIO)
    */
   const StringMap& getStringMap() const {
@@ -241,6 +257,8 @@ private:
       return _intMap;
     } else if constexpr (std::is_same_v<detail::GetVectorType<T>, float>) {
       return _floatMap;
+    } else if constexpr (std::is_same_v<detail::GetVectorType<T>, double>) {
+      return _doubleMap;
     } else {
       return _stringMap;
     }
@@ -253,6 +271,8 @@ private:
       return _intMap;
     } else if constexpr (std::is_same_v<detail::GetVectorType<T>, float>) {
       return _floatMap;
+    } else if constexpr (std::is_same_v<detail::GetVectorType<T>, double>) {
+      return _doubleMap;
     } else {
       return _stringMap;
     }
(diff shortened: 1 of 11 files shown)


src/SIOBlock.cc
```diff
@@ -75,12 +75,14 @@ void writeGenericParameters(sio::write_device& device, const GenericParameters&
...
   writeParamMap(device, params.getFloatMap());
   writeParamMap(device, params.getStringMap());
+  writeParamMap(device, params.getDoubleMap());
 }
 
 void readGenericParameters(sio::read_device& device, GenericParameters& params) {
   readParamMap(device, params.getIntMap());
   readParamMap(device, params.getFloatMap());
   readParamMap(device, params.getStringMap());
+  readParamMap(device, params.getDoubleMap());
```
tmadlener, line 85: This should be guarded by some version check to not break backwards compatibility on accident

python/podio/test_Frame.py
```diff
@@ -93,7 +93,11 @@ def test_frame_parameters(self):
...
       # Parameter not available as float (only int and string)
-      _ = self.event.get_parameter('SomeVectorData', as_type='float')
+      _ = self.event.get_parameter('SomeValue', as_type='float')
 
     self.assertEqual(self.event.get_parameter('SomeVectorData', as_type='int'), [1, 2, 3, 4])
     self.assertEqual(self.event.get_parameter('SomeVectorData', as_type='str'), ["just", "some", "strings"])
+    # as_type='float' will also retrieve double values (if the name is unambiguous)
+    self.assertEqual(self.event.get_parameter('SomeVectorData', as_type='float'), [0.0, 0.0])
+    # Avoid float comparison here and only check the size
+    self.assertEqual(len(self.event.get_parameter('SomeVectorData', as_type='float')), 2)
```
tmadlener, line 103:
```suggestion
```
The line above does the full comparison already?

jmcarcell:
One first comment: can you rebase master? Some tests without SIO are not working but this should have been fixed in https://github.com/AIDASoft/podio/pull/376 so just to make sure those work with these changes

include/podio/GenericParameters.h
```diff
@@ -112,6 +113,11 @@ class GenericParameters {
     setValue<std::string>(key, std::move(value));
   }
 
+  /// Overlaod for catching initializer list setting of string vector values
```
jmcarcell, line 116:
```suggestion
  /// Overload for catching initializer list setting of string vector values
```

python/podio/frame.py
```diff
@@ -10,49 +10,52 @@
...
-    f'{cpptype}': f'std::vector<{cpptype}>' for cpptype in SUPPORTED_PARAMETER_TYPES
-    })
+def _get_cpp_vector_types(type_str):
+  """Get the possible std::vector<cpp_type> from the passed py_type string."""
+  # Gather a list of all types that match the type_str (c++ or python)
+  types = list(filter(lambda t: type_str in t, SUPPORTED_PARAMETER_TYPES))
+  if not types:
+    raise ValueError(f'{type_str} cannot be mapped to a valid parameter type')
+
+  return [f'std::vector<{t}>' for t in map(lambda x: x[0], types)]
```
jmcarcell, line 58:
```suggestion
  return [f'std::vector<{t[0]}>' for t in types]
```

python/podio/frame.py
```diff
@@ -10,49 +10,52 @@
...
-    pytype: f'std::vector<{cpptype}>' for (pytype, cpptype) in zip(SUPPORTED_PARAMETER_PY_TYPES,
-                                                                     SUPPORTED_PARAMETER_TYPES)
-    }
-_PY_TO_CPP_TYPE_MAP.update({
-    f'{cpptype}': f'std::vector<{cpptype}>' for cpptype in SUPPORTED_PARAMETER_TYPES
-    })
+def _get_cpp_vector_types(type_str):
+  """Get the possible std::vector<cpp_type> from the passed py_type string."""
+  # Gather a list of all types that match the type_str (c++ or python)
+  types = list(filter(lambda t: type_str in t, SUPPORTED_PARAMETER_TYPES))
```
jmcarcell, line 54:
```suggestion
  types = [t for t in SUPPORTED_PARAMETER_TYPES if type_str in t]
```
More pythonic and I find list comprehensions more readable than compositions of `list(map(filter`...

python/podio/frame.py
```diff
@@ -10,49 +10,52 @@
...
   """Determine the supported types for the parameters.
 
-  Args:
-      lang (str): Language for which the type names should be returned. Either
-          'c++' or 'py'.
-
   Returns:
-      tuple (str): the tuple with the string representation of all **c++**
-          classes that are supported
+      tuple(tuple(str, str)): the tuple with the string representation of all
```
jmcarcell, line 17: This reminded me that maybe we could include at some point type hints? They are already there in the comments. And then use linters that make use of this information

jmcarcell (approved):
A few minor comments. Tests with and without SIO pass. 
If we read a file produced before this is merged and try to read it with podio after this is merged it should be fine right? As long as there is a check for the version every time one wants to read the double parameters.
Is it a good time for a new release? It has been a few months and there are a few changes, in particular this one that checks the version.

Issue #352, opened 2022-11-28
`podio-dump` output of overview table can be improved


tmadlener:
Currently `podo-dump` dumps the collection in a somewhat random order. For better readability it should do so in alphabetical order. Additionally, the column width is currently too narrow for many collection names, e.g.:
```
Collections:
Name                           Type                                     Size      
----------------------------------------------------------------------------------
VertexJets                     edm4hep::ReconstructedParticle           2         
PrimaryVertices                edm4hep::Vertex                          1         
BuildUpVertices_V0             edm4hep::Vertex                          0         
PFOsFromJets                   edm4hep::ReconstructedParticle           89        
LE_TightSelectedPandoraPFOs    edm4hep::ReconstructedParticle           82        
LE_LooseSelectedPandoraPFOs    edm4hep::ReconstructedParticle           88        
BuildUpVertices                edm4hep::Vertex                          4         
LE_SelectedPandoraPFOs         edm4hep::ReconstructedParticle           85        
TightSelectedPandoraPFOs       edm4hep::ReconstructedParticle           67        
SiTracksMCTruthLink            edm4hep::MCRecoTrackParticleAssociation  39        
PandoraClusters                edm4hep::Cluster                         86        
YokeEndcapCollectionContributions edm4hep::CaloHitContribution             33        
MCPhysicsParticles             edm4hep::MCParticle                      406       
RefinedVertexJets              edm4hep::ReconstructedParticle           2         
DebugHits                      edm4hep::TrackerHitPlane                 0         
HCalEndcapCollectionContributions edm4hep::CaloHitContribution             8848      
```
Ideally, the column widths adjust automatically to the contents.

PR #373, opened 2023-02-10
Adding ID to the short podio-dump output

kjvbrt:
BEGINRELEASENOTES
- Adding ID to the short podio-dump output

ENDRELEASENOTES

Also, adjusting the column widths, for example `edm4hep::ReconstructedParticle` has 31 characters.

### DIFF
--- a/tools/podio-dump
+++ b/tools/podio-dump
@@ -38,7 +38,7 @@ def print_frame(frame, cat_name, ientry, detailed):
   print('Collections:')
 
   if not detailed:
-    print(f'{"Name":<30} {"Type":<40} {"Size":<10}')
+    print(f'{"Name":<38} {"ID":<4} {"Type":<32} {"Size":<10}')
     print('-' * 82)
 
   # Print collections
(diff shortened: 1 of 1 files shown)


tools/podio-dump
```diff
@@ -49,7 +49,7 @@ def print_frame(frame, cat_name, ientry, detailed):
       coll.print()
       print(flush=True)
     else:
-      print(f'{name:<30} {coll.getValueTypeName():<40} {len(coll):<10}')
+        print(f'{name:<38} {coll.getID():<4} {coll.getValueTypeName():<32} {len(coll):<10}')
```
tmadlener, line 52:
```suggestion
      print(f'{name:<38} {coll.getID():<4} {coll.getValueTypeName():<32} {len(coll):<10}')
```

tools/podio-dump
```diff
@@ -38,7 +38,7 @@ def print_frame(frame, cat_name, ientry, detailed):
   print('Collections:')
 
   if not detailed:
-    print(f'{"Name":<30} {"Type":<40} {"Size":<10}')
+    print(f'{"Name":<38} {"ID":<4} {"Type":<32} {"Size":<10}')
```
tmadlener, line 41: 32 chars for the type might still not be enough for e.g. `edm4hep::MCRecoParticleAssociation`. Ideally this would be dynamically sized according to the longest type. However, we already have #352 for that, so we can keep this as is for now.

tmadlener (approved):
Thanks

PR #376, opened 2023-02-15
Fix tests without SIO

jmcarcell:
When running tests without SIO some python tests fail because they are trying to access `podio.SIOFrameWriter` in an import which doesn't exist without SIO. With this change I don't get any failures in the tests without and with SIO enabled.

BEGINRELEASENOTES
- Fix tests without SIO

ENDRELEASENOTES

### DIFF
--- a/python/podio/sio_io.py
+++ b/python/podio/sio_io.py
@@ -1,18 +1,14 @@
 #!/usr/bin/env python3
 """Python module for reading sio files containing podio Frames"""
 
-from podio.base_reader import BaseReaderMixin
 
 from ROOT import gSystem
 ret = gSystem.Load('libpodioSioIO')  # noqa: 402
 # Return values: -1 when it doesn't exist and -2 when there is a version mismatch
-if ret == -1:
-  raise ImportError('Unable to find libpodioSioIO')
-if ret == -2:
-  raise ImportError('Version mismatch for libpodioSioIO')
+if ret < 0:
+  raise ImportError('Error when importing libpodioSioIO')
 from ROOT import podio  # noqa: 402 # pylint: disable=wrong-import-position
-
-Writer = podio.SIOFrameWriter
+from podio.base_reader import BaseReaderMixin  # pylint: disable=wrong-import-position
 
 
 class Reader(BaseReaderMixin):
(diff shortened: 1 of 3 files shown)


python/podio/test_ReaderSio.py
```diff
@@ -3,7 +3,10 @@
 
 import unittest
 
-from podio.sio_io import Reader, LegacyReader
+try:
+  from podio.sio_io import Reader, LegacyReader
+except ImportError:
+  print('Unable to load SIO for this unit test, aborting the test...')
```
tmadlener, line 9: There is already a facility in place to skip sio based tests, when podio is not built with sio support. So here, rather than catching the `ImportError` I would just `import` the respective reader in the respective test case.

tmadlener (commented):
Thanks for catching this.

I thought I had handled this, but apparently I haven't ;) Everything should work once things are installed, because the sio specific python files will not be part of the installation.

My main comment below is to not duplicate the facilties for skipping SIO tests in the `test_ReaderSio.py`.

python/podio/sio_io.py
```diff
@@ -1,13 +1,14 @@
...
 from ROOT import gSystem
-gSystem.Load('libpodioSioIO')  # noqa: 402
+ret = gSystem.Load('libpodioSioIO')  # noqa: 402
+# Return values: -1 when it doesn't exist and -2 when there is a version mismatch
+if ret < 0:
+  raise ImportError('Error when importing libpodioSioIO')
 from ROOT import podio  # noqa: 402 # pylint: disable=wrong-import-position
-
-Writer = podio.SIOFrameWriter
+from podio.base_reader import BaseReaderMixin  # pylint: disable=wrong-import-position
```
tmadlener, line 11: Is there a specific reason why this import has been moved here?

Also I don't think the `Writer` should be removed (definitely not as part of this PR).

tmadlener: I mean I don't have anything against moving the `import` per se. It is just that the combination of moving it and then marking it with `disable=wrong-import-position` seems like a "double kill" to me. 

Maybe we should provide a virtualenv or something that has the same versions of `flake8` and `pylint` (and maybe even `pre-commit`) to make sure things can be run consistently in CI and locally.

tmadlener (approved):

PR #178, opened 2021-02-17
[WIP ]add graphviz tool to visualize datamodels

hegner:
BEGINRELEASENOTES
- Adding command line tool podio_vis to visualise data models via GraphViz.

ENDRELEASENOTES

PR #377, opened 2023-02-15
Add a visualization tool

jmcarcell:
Piggybacking on #178 (since it has been stale for a long time), I have modified
it and made the resulting graph look nicer.

The example data model in podio looks like this:

[podio.pdf](https://github.com/AIDASoft/podio/files/10747942/podio.pdf)

and EDM4hep like this:

[EDM4hep.pdf](https://github.com/AIDASoft/podio/files/10747946/gv.pdf)

BEGINRELEASENOTES
- Add a visualization tool that converts a YAML description to a graph

ENDRELEASENOTES

### DIFF
--- a/tools/podio-vis
+++ b/tools/podio-vis
@@ -1,45 +1,50 @@
 #!/usr/bin/env python3
+import argparse
 import os
 from podio.podio_config_reader import PodioConfigReader
 from graphviz import Digraph
 
-THIS_DIR = os.path.dirname(os.path.abspath(__file__))
-
-class Podio2Graphviz(object):
+class ModelToGraphviz(object):
   """Class to transform a data model description into a graphical representation"""
 
-  def __init__(self, yamlfile, target, dot):
+  def __init__(self, yamlfile, target, dot, format, filename):
     self.yamlfile = yamlfile
     self.target = target
     self.use_dot = dot
     self.datamodel = PodioConfigReader.read(yamlfile, 'podio')
-    self.graph = Digraph()
-    self.graph.format = "png"
+    self.graph = Digraph(node_attr={'shape': 'box'})
+    self.graph.attr(rankdir='LR', size='8,5')
+    self.format = format
+    self.filename = filename
 
-  def process(self):
-      # first create all nodes upfront
+  def make_viz(self):
       for name, datatype in self.datamodel.datatypes.items():
-        compatible_name = name.replace("::","_")  # graphviz gets confused with C++ '::' and formatting strings
-        self.graph.node(compatible_name, shape="box", label=name)
-      # then go through the relations
-      for name, datatype in self.datamodel.datatypes.items():
+        compatible_name = name.replace('::', '_')  # graphviz gets confused with C++ '::' and formatting strings
+        self.graph.node(compatible_name, label=name)
+        self.graph.attr('edge', color='red')
         for rel in datatype["OneToOneRelations"]:
-          compatible_name = name.replace("::","_")
-          compatible_type = rel.full_type.replace("::","_")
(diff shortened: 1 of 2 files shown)


vvolkl:
This looks great! I wonder if the graph could be made even more informative by grouping types in clusters (https://graphviz.org/Gallery/directed/cluster.html). I'm thinking of the (Monte Carlo|Raw Data| Digitization | Reco & Analysis)  categories in https://github.com/key4hep/EDM4hep/raw/master/doc/edm4hep_diagram.svg 

The info which type belongs to which cluster would have to be hardcoded in the script or added to the yaml though.

hegner:
@jmcarcell - great that you are picking that up :-)

tmadlener:
Thanks for reviving this.

My 2 cents:
- I think the grouping could be left as an (optional) input to the visualization tool. That way the model definition would remain unchanged, and it would even be possible to generate diagrams with different groupings. This could just be another yaml file along the lines of
```yaml
group:
  - datatype
  - otherDatatype
anotherGroup:
  - anotherType
```
- I would propose svg as output instead of pdf, because it is much easier to embed into a webpage (and one could do some interactive fancyness if desired). This should be fairly trivial to change with graphviz, I think.
- Is it possible to make the `Associations` have two arrow ends? This would make it clearer that these are not directed.
- I am not sure inverting the direction of the edges just to get the right ordering is the way to go. IMHO the diagram gets confusing in this way, because the arrows point in the opposite direction of which information is actually retrievable.
- Maybe also filtering a few datatypes from the graph would be nice to have (e.g. some of the Associations in EDM4hep)

The last two points might be addressable by a more complex grammar for the optional input yaml file. Filtering should be rather easy, by simply only using the types that are stated in the file. Defining a "layout" for the groups potentially becomes much harder. Here, I am really not sure how complex we want this tool to be, and I would rather start with a simple one that gives correct, but potentially not yet visually perfect results.

hegner:
Thanks a lot. Looks great

hegner:
@jmcarcell - can you fix the flake8 error of the pre-commit test?

tools/podio-vis
```diff
@@ -0,0 +1,104 @@
...
+if __name__ == "__main__":
+  parser = argparse.ArgumentParser(description='Given a description yaml file this script generates '
+                                   'a visualization in the target directory')
+
+  parser.add_argument('description', help='yaml file describing the datamodel')
+  parser.add_argument('-d', '--dot', action='store_true', default=False,
+                      help='just write the dot file')
+  parser.add_argument('--fmt', default='svg', help='Which format to use for saving the file')
+  parser.add_argument('--filename', default='gv', help='Which filename to use for the output')
+  parser.add_argument('--extra-input', help='Extra input file for defining groups')
```
tmadlener, line 98:
```suggestion
  parser.add_argument('--graph-config', help='Optional input file for configuring, e.g. the layout of the generated graph')
```
Just to be slightly more specific here on the purpose of this input

tmadlener (commented):
Thanks a lot. Looks good to me as well. Just one minor comment about the naming of the `extra-input` argument.

hegner (changes requested):
You would need to list graphviz as an *optional* dependency in the PODIO description
  https://github.com/AIDASoft/podio/blob/master/README.md
We want to keep the necessary dependencies of PODIO as small as possible.

Since it is optional you could even put a try/catch in the graphviz import of the new tool printing a little comment.
The really clean solution would be actually non-installing the script if graphviz isn't available. But that may be an overkill

hegner:
@jmcarcell - thanks! The `requirements.txt` you can ignore for now.

If you make the code checker (see sys.exit) happy, I can merge :-)

hegner (approved):

PR #378, opened 2023-02-17
Mark EventStore based I/O as deprecated

tmadlener:
BEGINRELEASENOTES
- Introduce deprecation warnings for the **EventStore based I/O model** as it **will be removed in favor of the `Frame` based one**

ENDRELEASENOTES

In order to pass CI deprecation warnings are temporarily accepted and this exception has to be taken back again when the EventStore code is removed.

### DIFF
--- a/.github/workflows/edm4hep.yaml
+++ b/.github/workflows/edm4hep.yaml
@@ -46,7 +46,7 @@ jobs:
             cmake -DENABLE_SIO=ON \
               -DCMAKE_INSTALL_PREFIX=../install \
               -DCMAKE_CXX_STANDARD=17 \
-              -DCMAKE_CXX_FLAGS=" -fdiagnostics-color=always -Werror " \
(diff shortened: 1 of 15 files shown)


tmadlener:
The edm4hep workflow currently fails because it still uses the `EventStore` until key4hep/EDM4hep#184 is merged. I suggest we wait for that to go in first, and the merge this.

hegner:
@tmadlener - before merging we have to get the Gaudi part fixed as well.

tmadlener:
In principle this can be merged even without those fixes since we are only introducing a few deprecation warnings and not yet removing anything. But since this is not extremely urgent we can also wait.

hegner:
@tmadlener - actually it is errors in the current setup. Otherwise the PR checks would succeed ;-)

tmadlener:
The errors are deprecation warnings that are triggered in the EDM4hep build. They should be fixed by key4hep/EDM4hep#194 and key4hep/EDM4hep#182

tmadlener:
All downstream (EDM4hep) fixes have been merged, so I think this could be merged.

hegner (approved):

PR #383, opened 2023-02-28
Introduce the infrastructure for auto-generating API documentation via sphinx and doxygen

### LINKED ISSUE #350: API docs for podio, including python bindings
podio in principle comes with a doxygen configuration, but it seems to not be working properly at the moment. This should be fixed to actually have an API documentation for podio.

Additionally, the python bindings should be documented in a similar way. Doxygen in principle has support for python, but the docstrings of the python bindings would need to be adapted in that case.

tmadlener:
BEGINRELEASENOTES
- Switch to `sphinx` for generating documentation
  - Keep `doxygen` for generating the c++ files and then use `breathe` to integrate that into the `sphinx` generated documentation
- Add python bindings auto documentation generation via `sphinx`
- Fixes #350 

ENDRELEASENOTES

This is currently using the *read the docs* sphinx theme, but we can in principle switch to anything else. The `Sphinx` target is not built automatically even if `CREATE_DOC` is enabled.

This introduces the basic infrastructure to run doxygen as well as autodoc within sphinx to generate API documentation from docstrings automatically. It furthermore also produces the other documentation from the markdown files in `doc`. I think actually fixing the content of the documentation can go to subsequent PRs, where we can also address smaller style changes. 

Additionally, this PR also does not solve the question on where to host the **versioned** documentation. For now this will just push the things to the `gh-pages` branch, which will automatically publish it on github pages.


tmadlener:
This currently includes configuration for publishing via `readthedocs`. Since, readthedocs builds in a slightly different environment then we usually do this will currently not work for the generation of the python API docs. There are a few things that can be done to work around that though.

tmadlener:
Removed the readthedocs configuration again and instead went for an attempt that uses `sphinx-multiversion` + some functionality in the github actions workflow.

The current result can be seen here at https://tmadlener.github.io/podio

What this does not yet do and where I think we might actually need some dedicated scripting:
- [ ] Also automatically generate the c++ and python api documentation for all tags. This seems to currently only work for the current tag / branch

Another slight disadvantage of `sphinx-multiversion` is that it always regenerates all versions, whereas we would probably be fine with having things only regenerate the current version.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,81 @@
+name: publish-doc
+
+on:
+  push:
+    branches:
+       - 'master'
+       - 'docs-on-gh-pages'
```
tmadlener, line 7:
```suggestion
```
Just before merging

tmadlener:
I have removed the `sphinx-multiversion` support again as it will not work as intended. With this, this PR actually removes all "fancy" functionality like versioned documentation. For now it just allows to generate documentation locally, and also publishing the latest master version to github pages.

doc/conf.py
```diff
@@ -0,0 +1,160 @@
...
+    [
+        sys.executable,
+        "-m",
+        "sphinx.ext.apidoc",
+        "--force",
+        "-o",
+        "py_api",
+        "../python",
+        "../*/*test_*.py",  # exclude tests
+        "../python/podio_version.py",
```
tmadlener, line 150:
```suggestion
        "../python/podio_version.py",  # exclude convenience module
```

doc/conf.py
```diff
@@ -0,0 +1,160 @@
...
+        ]
+    )
+
+print("Done with python API doc generation")
+
+# -- multiversion setup -------------------------------------------------------
+mv_tag_whitelist = r'^v[0-9]{2}-[0-9]{2}(-[0-9]{2})?$'
+smv_remote_whitelist = r'^(origin)$'
+smv_branch_whitelist = r'^(master|docs-on-gh-pages)$'
+smv_released_pattern = r'^tags/.*$'
```
tmadlener, line 160:
```suggestion
```
Remove `sphinx-multiversion` leftovers

PR #384, opened 2023-03-02
Add a conversion operator to std::string for podio::version::Version

veprbl:
BEGINRELEASENOTES
- Added an operator for conversion to std::string for podio::version::Version

ENDRELEASENOTES

### DIFF
--- a/podioVersion.in.h
+++ b/podioVersion.in.h
@@ -62,7 +62,7 @@ struct Version {
   #undef DEFINE_COMP_OPERATOR
 #endif
 
-  operator std::string() const {
+  explicit operator std::string() const {
     std::stringstream ss;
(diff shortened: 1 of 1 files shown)


tmadlener:
Can we make this an `explicit operator`? I am not the biggest fan of implicit conversions, and in this case the difference for users would be acceptable IMHO. 
```cpp
const auto versionStr = std::string(version);
```
rather then
```cpp
const std::string versionStr = version;
```

Unless you have a specific use case this would fix where the explicit cast would be too cumbersome?

veprbl:
I also prefer the explicit conversion, but it's not so popular in C++ world.

tmadlener (approved):

PR #387, opened 2023-03-10
Add an argument to exclude SIO files from some tests

jmcarcell:
BEGINRELEASENOTES
- Make sure that the dump model round trip tests work without `ENABLE_SIO`
- Actually test the extension model dumping

ENDRELEASENOTES

This allows not to have to cleanup a few folders in `test` in case the SIO files are present, which would break the tests. Useful in case one is switching between SIO and non-SIO for example

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -270,8 +270,8 @@ set_tests_properties(
 
 set(sio_roundtrip_tests "")
 if (TARGET read_sio)
-  add_test(datamodel_def_store_roundtrip_sio ${CMAKE_CURRENT_LIST_DIR}/scripts/dumpModelRoundTrip.sh false ${CMAKE_CURRENT_BINARY_DIR}/example_frame.sio datamodel)
-  add_test(datamodel_def_store_roundtrip_sio_extension ${CMAKE_CURRENT_LIST_DIR}/scripts/dumpModelRoundTrip.sh false ${CMAKE_CURRENT_BINARY_DIR}/example_frame.sio datamodel extension_datamodel)
+  add_test(datamodel_def_store_roundtrip_sio ${CMAKE_CURRENT_LIST_DIR}/scripts/dumpModelRoundTrip.sh true ${CMAKE_CURRENT_BINARY_DIR}/example_frame.sio datamodel)
+  add_test(datamodel_def_store_roundtrip_sio_extension ${CMAKE_CURRENT_LIST_DIR}/scripts/dumpModelRoundTrip.sh true ${CMAKE_CURRENT_BINARY_DIR}/example_frame.sio datamodel extension_datamodel)
 
   set(sio_roundtrip_tests
     datamodel_def_store_roundtrip_sio
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -252,11 +252,11 @@ endif()
 # Add tests for storing and retrieving the EDM definitions into the produced
 # files
 if(ENABLE_SIO)
-add_test(datamodel_def_store_roundtrip_root ${CMAKE_CURRENT_LIST_DIR}/scripts/dumpModelRoundTrip.sh true ${CMAKE_CURRENT_BINARY_DIR}/example_frame.root datamodel)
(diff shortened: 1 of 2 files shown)


tmadlener:
Thanks for this. Since I always build everything nowadays, it seems I no longer realize the small issues in switching between with and without SIO. I would do this slightly differently from a conceptual point of view:

- Make the `COMP_BASE_FOLDER` a required argument of the `dumpModelRoundTrip.sh` script, and adapt the `CMakeLists.txt` to pass in `.` for the `datamodel` case (or the full path)
- Make the `dumpModelRoundTrip.sh` script simply take additional arguments, that we simply pass on directly to the `diff` commands (using `shift` and `$@`)
- In `CMakeLists.txt` make a variable for which the value depends on `ENABLE_SIO` and simply pass that on to the script which should then do the right thing automatically

This approach would only have one `if` block instead of a few coupled ones, and it wouldn't need the boolean flag.

tmadlener:
rebased this onto the latest master and reworked the overall logic slightly to remove a few of the if/else branches

jmcarcell:
I'll have a look at this; I found out that in the original version the extension test was not being run correctly (it was the same as the non-extension one) so I need to have a look in detail

jmcarcell:
I see now that you fixed that on your last commit. I started working but never pushed it...
I have made one extra change and that is checking for whether `clang-format` is being used right now or not. Currently the tests will fail if the cmake option is disabled so the last commit checks for that and the tests now pass with and without `clang-format`

tests/CMakeLists.txt
```diff
@@ -290,5 +314,5 @@ set_tests_properties(
   PROPERTIES
     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
     ENVIRONMENT
-  "PODIO_BASE=${CMAKE_SOURCE_DIR};IO_HANDLERS=${IO_HANDLERS};LD_LIBRARY_PATH=${CMAKE_CURRENT_BINARY_DIR}:${CMAKE_BINARY_DIR}/src:$ENV{LD_LIBRARY_PATH};PYTHONPATH=${CMAKE_SOURCE_DIR}/python:$ENV{PYTHONPATH};ROOT_INCLUDE_PATH=${CMAKE_SOURCE_DIR}/tests/datamodel:${CMAKE_SOURCE_DIR}/include:$ENV{ROOT_INCLUDE_PATH}"
+  "PODIO_BASE=${CMAKE_SOURCE_DIR};IO_HANDLERS=${IO_HANDLERS};ENABLE_SIO=${ENABLE_SIO};PODIO_USE_CLANG_FORMAT=${PODIO_USE_CLANG_FORMAT};LD_LIBRARY_PATH=${CMAKE_CURRENT_BINARY_DIR}:${CMAKE_BINARY_DIR}/src:$ENV{LD_LIBRARY_PATH};PYTHONPATH=${CMAKE_SOURCE_DIR}/python:$ENV{PYTHONPATH};ROOT_INCLUDE_PATH=${CMAKE_SOURCE_DIR}/tests/datamodel:${CMAKE_SOURCE_DIR}/include:$ENV{ROOT_INCLUDE_PATH}"
```
tmadlener, line 317: If we pass `PODIO_USE_CLANG_FORMAT` as additional argument in the `add_test`s above, we don't need to pass it here into the test environment, right?

tests/scripts/dumpModelRoundTrip.sh
```diff
@@ -3,13 +3,17 @@
...
 EDM_NAME=${2}  # the name of the EDM
-COMP_BASE_FOLDER=""  # where the source to compare against is
-if [ -$# -gt 2 ]; then
-    COMP_BASE_FOLDER=${3}
+COMP_BASE_FOLDER=${3}  # where the source to compare against is
+CLANG_FORMAT=${4}  # whether to use clang-format
+shift 4
+EXTRA_GEN_ARGS=${@}
+
+if [ ${CLANG_FORMAT} = "ON"] || [ ${CLANG_FORMAT} = "AUTO" ]; then
```
tmadlener, line 15:
```suggestion
if [ ${CLANG_FORMAT} = "ON" ] || [ ${CLANG_FORMAT} = "AUTO" ]; then
```

tests/CMakeLists.txt
jmcarcell: I removed it from the script

jmcarcell:
Now it works for non SIO, SIO builds with and without clang-format and also going from non-SIO to SIO and viceversa

tmadlener:
Thanks again for checking. Waiting for the CI before merging.

PR #389, opened 2023-03-13
Add .cache to the gitignore

jmcarcell:
BEGINRELEASENOTES
- Add .cache to the gitignore

ENDRELEASENOTES

from `clangd` indexing.


tmadlener (approved):
Ah someone else who is using a newer version of LLVM ;) Hadn't realized they moved this into a different folder. Thanks.

PR #390, opened 2023-03-13
Check for all the arguments that are going to be used for clang-format

### LINKED ISSUE #388: Issue with clang format in the tests
https://github.com/AIDASoft/podio/pull/358 introduced a sort of hidden dependency on `clang-format` as now there are tests which will generate a model with clang format and the `-fallback-style=llvm` argument is not supported by older versions of `clang-format`. This breaks the nightly build for key4hep, because `llvm` is not a dependency. I think we can agree that the tests should run without `llvm` since it's not a dependency of podio. These are some solutions I can think about:
- In the CMake step, when generating the original model save some variable to be checked later at the tests to make sure that we run the tests with the same arguments as now
- Alternatively we could just parse the dumped model and the original one even in a python one liner and compare the results. It's a slightly weaker test but there are tests already for the generation from the model

jmcarcell:
BEGINRELEASENOTES
- Modify the initial `clang-format` check to try to run with all the arguments that will be used later

ENDRELEASENOTES

Fixes #388. Testing works fine with a recent version of `clang-format`. Should be fine with older versions (we can let the nightly test this).

### DIFF
--- a/python/podio_class_generator.py
+++ b/python/podio_class_generator.py
@@ -40,10 +40,14 @@ REPORT_TEXT = """
 def get_clang_format():
   """Check if clang format is available and if so get the list of arguments to
   invoke it via subprocess.Popen"""
-  try:
-    cformat_exe = subprocess.check_output(['which', 'clang-format']).strip()
-    return [cformat_exe, "-style=file", "-fallback-style=llvm"]
-  except subprocess.CalledProcessError:
+  try :
+    out = subprocess.check_output(["clang-format", "-style=file", "-fallback-style=llvm", "--help"], stderr=subprocess.STDOUT)
+    if b'Unknown' in out:
+      print('ERROR: At least one argument was not recognized by clang-format')
+      print('       Most likely the version you are using is old')
+      return []
+    return ["clang-format", "-style=file", "-fallback-style=llvm"]
+  except FileNotFoundError:
     print("ERROR: Cannot find clang-format executable")
     print("       Please make sure it is in the PATH.")
     return []
--- a/python/podio_class_generator.py
+++ b/python/podio_class_generator.py
@@ -46,6 +46,9 @@ def get_clang_format():
       print('ERROR: At least one argument was not recognized by clang-format')
       print('       Most likely the version you are using is old')
       return []
+    out = subprocess.check_output('echo | clang-format -style=file ', stderr=subprocess.STDOUT, shell=True)
+    if b'.clang-format' in out:
+      return []
     return ["clang-format", "-style=file", "-fallback-style=llvm"]
   except FileNotFoundError:
     print("ERROR: Cannot find clang-format executable")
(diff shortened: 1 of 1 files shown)


tmadlener:
Does this recognize cases where no `.clang-format` file is present, respectively, where it would require a newer version of `clang-format` as well? Or does it currently mainly cover the existence of certain command line flags, but not more?

tmadlener:
Unfortunately clang-format wants a file and doesn't read anything from stdin. We would need a file that is definitely present also in downstream projects, which makes this test slightly non-trivial.

jmcarcell:
It's not super easy because also it doesn't check at the same time the flags and the `.clang-format` file. 
But it does take stdin, from `--help`:
```
If no arguments are specified, it formats the code from standard input
and writes the result to the standard output.
```

python/podio_class_generator.py
```diff
@@ -41,9 +41,17 @@ def get_clang_format():
...
+                                  stderr=subprocess.STDOUT)
+    if b'Unknown' in out:
+      print('ERROR: At least one argument was not recognized by clang-format')
+      print('       Most likely the version you are using is old')
+      return []
+    out = subprocess.check_output('echo | clang-format -style=file ', stderr=subprocess.STDOUT, shell=True)
+    if b'.clang-format' in out:
+      return []
+    return ["clang-format", "-style=file", "-fallback-style=llvm"]
+  except FileNotFoundError:
```
tmadlener, line 54: Stupid question: Why is this a `FileNotfoundError` here instead of a `CalledProcessError`?

jmcarcell:
Current logic: check if `-fallback-style` can be used, if not then don't run `clang-format`. Then check if there is a `.clang-format` file, if not then don't run `clang-format`. After that, if we are here that means we can run `clang-format` with all the flags. The behaviour with the first check is needed for situations like the nightly, for the second check we could run it anyway? Checking now I realize the LLVM style is a default:
```
$ echo | clang-format --style=file
Can't find usable .clang-format, using LLVM style
```
with LLVM 3.4.2
and more recent ones say (`--help`):
```
  --fallback-style=<string>      - The name of the predefined style used as a
                                   fallback in case clang-format is invoked with
                                   -style=file, but can not find the .clang-format
                                   file to use. Defaults to 'LLVM'.
                                   Use -fallback-style=none to skip formatting.
```
with `clang-format` 15.0.7
So alternatively we could just remove the flag and the behaviour would be the same, or am I missing something?

tmadlener:
What happens if there is a `.clang-format` file, but it is malformed (or simply doesn't work with the current version of `clang-format`)? Is there something more robust that we could check (e.g. non-zero exit) to also get such cases?

python/podio_class_generator.py
tmadlener: Ah OK. I wasn't aware of that. TIL. Thanks :)

tmadlener (approved):

PR #394, opened 2023-03-21
Decouple the collection buffer creation from the collection

tmadlener:
BEGINRELEASENOTES
- Introduce a `CollectionBufferFactory` that can create the necessary buffers from a collection type, a schema version and a subset collection flag.
  - Use this factory throughout all existing Readers
  - Remove `createBuffers` and `createSchemaEvolvableBuffers` from `podio::CollectionBase` interface
- Make the minimum allowed `schema_version` 1 in the yaml definition files. Default to 1 if no `schema_version` is provided
- Add a `schemaVersion` to the `DatamodelDefinition.h` header that is generated and that can be accessed via `{{ package_name }}::meta::schemaVersion`. Use this to propagate schema information to the necessary places.
- Make `SIOBlocks` write the current schema version, such that on reading they can generate the appropriate buffers for the version on file.

ENDRELEASENOTES

This PR continues the cleanup of internals and the introduction of full schema evolution capabilities. The most important bit is that it completely disentangles the buffer creation from the collection interface, so that it is now possible to create buffers without having to create a collection first. This is achieved via a `CollectionBufferFactory` (singleton) that is populated during datamodel library loading. The rest of this PR is mainly shuffling some parts of existing code around and removing some other bits that were previously necessary to "steal" the buffers from the collection they were created from.

This does not yet address any schema evolution concerns other than providing the possibility to generate different buffers for different schema versions and the basics for populating the factory with versioned creation function. For actual schema evolution functionality this has to be populated accordingly (either by users or via automatic code generation).

@hegner this introduces an other de-facto registry, but I don't see an easy way around that. We could hide the fact that it is a singleton registry behind an interface of global functions if we wanted to.

This should solve the issue that is currently blocking #257. 

TODO:
- [x] Proper writing of schema version for SIO backend (should be possible to use the version mechanism of SIO here)
- [x] Documentation

### DIFF
--- a/src/UserDataCollection.cc
+++ b/src/UserDataCollection.cc
@@ -17,16 +17,17 @@ namespace {
   template <typename T>
   int registerUserDataCollection(T) {
     // Register with schema version 1 to allow for potential changes
-    CollectionBufferFactory::mutInstance().registerCreationFunc(userDataCollTypeName<T>(), 1, [](bool) {
-      return podio::CollectionReadBuffers{new std::vector<T>(), nullptr, nullptr,
-                                          [](podio::CollectionReadBuffers buffers, bool) {
-                                            return std::make_unique<UserDataCollection<T>>(
-                                                std::move(*buffers.dataAsVector<T>()));
-                                          },
-                                          [](podio::CollectionReadBuffers& buffers) {
-                                            buffers.data = podio::CollectionWriteBuffers::asVector<T>(buffers.data);
-                                          }};
-    });
+    CollectionBufferFactory::mutInstance().registerCreationFunc(
+        userDataCollTypeName<T>(), UserDataCollection<T>::schemaVersion, [](bool) {
+          return podio::CollectionReadBuffers{new std::vector<T>(), nullptr, nullptr,
+                                              [](podio::CollectionReadBuffers buffers, bool) {
+                                                return std::make_unique<UserDataCollection<T>>(
+                                                    std::move(*buffers.dataAsVector<T>()));
+                                              },
+                                              [](podio::CollectionReadBuffers& buffers) {
+                                                buffers.data = podio::CollectionWriteBuffers::asVector<T>(buffers.data);
+                                              }};
+        });
 
     return 1;
   }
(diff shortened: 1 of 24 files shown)


tmadlener:
Updated the docstring for the `CollectionBufferFactory` to include the basics and some of the assumptions. Since this is a class that is mostly targeted at internal use, I have not created additional documentation in the user facing markdown files.

hegner:
@tmadlener - did you see the cling error of the IBs before?

hegner (approved):
Thanks. To me it looks fine. 
There is only a nitpicking one about underscore vs camel case for variable names. But that we should discuss for coding conventions anyhow.

src/UserDataCollection.cc
```diff
@@ -0,0 +1,54 @@
...
+        userDataCollTypeName<T>(), UserDataCollection<T>::schemaVersion, [](bool) {
+          return podio::CollectionReadBuffers{new std::vector<T>(), nullptr, nullptr,
+                                              [](podio::CollectionReadBuffers buffers, bool) {
+                                                return std::make_unique<UserDataCollection<T>>(
+                                                    std::move(*buffers.dataAsVector<T>()));
+                                              },
+                                              [](podio::CollectionReadBuffers& buffers) {
+                                                buffers.data = podio::CollectionWriteBuffers::asVector<T>(buffers.data);
+                                              }};
+        });
```
tmadlener, line 30: This does not put the correct `schemaVersion` into the created buffers.

tmadlener: This doesn't apply yet and needs to be part of #413

hegner: sorry, your comments confuse me a bit

tmadlener: Sorry. I thought I forgot to do this here while working on the other end of schema evolution in #413. But I then realized the necessary `schemaVersion` field is only introduced there, so I can do everything I need in that. Is there anything missing for this or should we merge?

hegner: Sounds fine to me.

hegner (approved):

Issue #393, opened 2023-03-21
Problems in the IO code


jmcarcell:
When implementing support for RNTuple I found a few issues with some parts of the IO code that will have to be modified or worked around for RNTuple to work.

- At least in one place a pointer that is supposed to point to the data points to an `unique_ptr`. This place is https://github.com/AIDASoft/podio/blob/c7328d62f17061c8360eb2ac0a3fe0ce931fd2ae/python/templates/CollectionData.cc.jinja2#LL94-L94C45. 
- When converting buffers for writing, there is a three star cast: What this cast does is basically to return a pointer to `vector.data()` instead of a pointer to the vector itself. This also relies on the fact that implementations of `std::vector` save the pointer to the actual array in the first `size_ptr` bytes (if this wasn't the case the cast wouldn't even point to the data array): https://github.com/AIDASoft/podio/blob/c7328d62f17061c8360eb2ac0a3fe0ce931fd2ae/include/podio/CollectionBuffers.h#L40

How all of this works together is a mystery to me, as the TTree functions seem to be happy but the RNTuple ones don't. For the first point, I think we also probably rely on the fact that `unique_ptr` can have size zero. The first point can be easily worked around by saving another pointer to the actual data and using that one. Fixing any of those points will cause all the writers and readers to stop writing where they either crash or (the writers) don't crash but they don't write the data correctly.

tmadlener:
Let me try to bring some information on the original intent into this. I have to point out, that not all of this is ideally implemented yet, and there is still some historical baggage around that we didn't come around to fixing yet, because schema evolution would have touched similar code. I am working on getting closer to the ideas described below.

 There are two types of collection buffers, which mainly differ by the ownership of the data they transport:
- `CollectionWriteBuffers` only hold pointers to data that is owned by some collection.
- `CollectionReadBuffers` own their data (although that is really not visible at all from the "interface").

The main idea is that since collections own their data, we simply need a convenient way to handle all necessary data for writing through one access point (instead of three like it was in the past). On the other hand when reading that, there is no collection yet to own the data, so we read all data into buffers and then make the collections constructible form these buffers (effectively "stealing" the data in the process).

Another important point is that while relation (`references`) and vector member information is somewhat easy to "standardize" across different data types, the `data` is really different for all data types, but I wanted only one class (instead of some inheritance structure), that is able to transport this, so `data` is a `void*` and we need to inject necessary type information via some other means and cast whenever necessary.
[...]

PR #395, opened 2023-03-22
Add support for the new RNTuple format

jmcarcell:
Add a writer and a reader using the podio Frame format. All tests pass and it doesn't break the other writes but the cost for that is a couple of ugly workarounds (see https://github.com/AIDASoft/podio/issues/393 for some of the issues that were found while doing this). My idea is that this is merged first and then we worry about the workarounds if fixes have not been found in time, otherwise this PR will get quite complicated.

For running with this a recent version of ROOT is needed, 6.28.00 or newer (I'm not sure 100% if it works with 6.28.00 since I have been running off master). There is a new option in the CMakeLists to enable support for this by asking cmake to find the relevant ROOT library. When this option is enabled (`-DENABLE_RNTUPLE`), the writer reader and tests will be compiled, otherwise nothing should have changed. 

The writer and reader have a very similar interface to the current TTree-based frame ones. The tests are the same for both. How the writer and reader work is a bit different, of course, from the TTree-based. For example, RNTuple doesn't support mapped types right now so for the generic parameters there is a 'disentangling' from the map types to simple vector types that are saved in the rntuple and then reconstructed in the reader (from vector to map types). I tried to make it as simple as possible by not using extra types and trying to pull few headers from podio.

It's still not fully ready for merging as cleanups need to be done and more comments are needed in several places. Also the support for reading from multiple files is not complete right now. It is a bit more tricky than the TTree case since there is no equivalent to a TChain as far as I know.

BEGINRELEASENOTES
- Add support for the new RNTuple format by adding a writer, reader and tests.

ENDRELEASENOTES

Supersedes https://github.com/AIDASoft/podio/pull/247

### DIFF
--- a/include/podio/CollectionBuffers.h
+++ b/include/podio/CollectionBuffers.h
@@ -27,6 +27,7 @@ using VectorMembersInfo = std::vector<std::pair<std::string, void*>>;
  */
 struct CollectionWriteBuffers {
   void* data{nullptr};
+  void* vecPtr{nullptr};
   CollRefCollection* references{nullptr};
   VectorMembersInfo* vectorMembers{nullptr};
 
--- a/include/podio/ROOTNTupleReader.h
+++ b/include/podio/ROOTNTupleReader.h
@@ -93,11 +93,10 @@ private:
    * collection infos and all necessary meta data to be able to read entries
    * with this name
    */
-  void initCategory(const std::string& category);
+  bool initCategory(const std::string& category);
 
   GenericParameters readEventMetaData(const std::string& name);
 
-  std::vector<std::string> m_availCategories{};                 ///< All available categories from this file
   
   podio::version::Version m_fileVersion{0, 0, 0};
   DatamodelDefinitionHolder m_datamodelHolder{};
@@ -112,6 +111,8 @@ private:
   std::map<std::string, std::vector<std::string>> m_collectionType;
   std::map<std::string, std::vector<bool>> m_isSubsetCollection;
 
+  std::map<std::string, int> m_totalEntries;
+  std::vector<std::string> m_availableCategories;
 
 };
 
--- a/include/podio/ROOTNTupleReader.h
+++ b/include/podio/ROOTNTupleReader.h
@@ -1,0 +1,120 @@
+#ifndef PODIO_ROOTNTUPLEREADER_H
+#define PODIO_ROOTNTUPLEREADER_H
+
+#include "podio/ROOTFrameData.h"
+#include "podio/CollectionBranches.h"
+#include "podio/ICollectionProvider.h"
+#include "podio/podioVersion.h"
+#include "podio/utilities/DatamodelRegistryIOHelpers.h"
+
+#include <algorithm>
+#include <iostream>
+#include <map>
+#include <string>
+#include <tuple>
+#include <utility>
+#include <vector>
+
+#include <ROOT/RNTuple.hxx>
+#include <ROOT/RNTupleModel.hxx>
+
+
+namespace rnt = ROOT::Experimental;
+// forward declarations
+class TClass;
+class TFile;
+
+namespace podio {
+
+namespace detail {
+  // Information about the data vector as wall as the collection class type
+  // and the index in the collection branches cache vector
+  using CollectionInfo = std::tuple<const TClass*, const TClass*, size_t>;
+
+} // namespace detail
+
+class EventStore;
+class CollectionBase;
+class Registry;
+class CollectionIDTable;
+class GenericParameters;
+/**
+This class has the function to read available data from disk
+and to prepare collections and buffers.
+**/
+class ROOTNTupleReader{
+  friend EventStore;
+
+public:
+  ROOTNTupleReader() = default;
+  ~ROOTNTupleReader() = default;
+
+  // non-copyable
+  ROOTNTupleReader(const ROOTNTupleReader&) = delete;
+  ROOTNTupleReader& operator=(const ROOTNTupleReader&) = delete;
+
+  void openFile(const std::string& filename);
+  void openFiles(const std::vector<std::string>& filename);
+
+  /**
+   * Read the next data entry from which a Frame can be constructed for the
+   * given name. In case there are no more entries left for this name or in
+   * case there is no data for this name, this returns a nullptr.
+   */
+  std::unique_ptr<podio::ROOTFrameData> readNextEntry(const std::string& name);
+
+  /**
+   * Read the specified data entry from which a Frame can be constructed for
+   * the given name. In case the entry does not exist for this name or in case
+   * there is no data for this name, this returns a nullptr.
+   */
+  std::unique_ptr<podio::ROOTFrameData> readEntry(const std::string& name, const unsigned entry);
+
+  /// Returns number of entries for the given name
+  unsigned getEntries(const std::string& name);
+
+  /// Get the build version of podio that has been used to write the current file
+  podio::version::Version currentFileVersion() const {
+    return m_fileVersion;
+  }
+
+  void closeFile();
+
+private:
+  std::vector<std::unique_ptr<ROOTNTupleReader>> m_files;
+  std::unique_ptr<ROOT::Experimental::RNTupleReader> m_file;
+  std::unique_ptr<ROOT::Experimental::RNTupleReader> m_metadata;
+
+  
+private:
+
+  /**
+   * Initialze the passed CategoryInfo by setting up the necessary branches,
+   * collection infos and all necessary meta data to be able to read entries
+   * with this name
+   */
+  void initCategory(const std::string& category);
+
+  GenericParameters readEventMetaData(const std::string& name);
+
+  std::vector<std::string> m_availCategories{};                 ///< All available categories from this file
+  
+  podio::version::Version m_fileVersion{0, 0, 0};
+  DatamodelDefinitionHolder m_datamodelHolder{};
+
+  std::map<std::string, std::vector<std::unique_ptr<ROOT::Experimental::RNTupleReader>>> m_readers;
+  std::map<std::string, std::unique_ptr<ROOT::Experimental::RNTupleReader>> m_metadata_readers;
+  std::vector<std::string> m_filenames;
+  std::map<std::string, int> m_entries;
+
+  std::map<std::string, std::vector<int>> m_collectionId;
+  std::map<std::string, std::vector<std::string>> m_collectionName;
+  std::map<std::string, std::vector<std::string>> m_collectionType;
+  std::map<std::string, std::vector<bool>> m_isSubsetCollection;
+
+
+};
+
+} // namespace podio
+
+#endif
--- a/include/podio/ROOTNTupleReader.h
+++ b/include/podio/ROOTNTupleReader.h
@@ -7,11 +7,9 @@
 #include "podio/podioVersion.h"
 #include "podio/utilities/DatamodelRegistryIOHelpers.h"
 
-#include <algorithm>
 #include <iostream>
 #include <map>
 #include <string>
-#include <utility>
 #include <vector>
 
 #include <ROOT/RNTuple.hxx>
@@ -20,10 +18,6 @@
 
 namespace podio {
 
-class CollectionBase;
-class Registry;
-class CollectionIDTable;
-class GenericParameters;
 /**
 This class has the function to read available data from disk
 and to prepare collections and buffers.
@@ -65,36 +59,34 @@ public:
   void closeFile();
 
 private:
-  std::vector<std::unique_ptr<ROOTNTupleReader>> m_files;
-  std::unique_ptr<ROOT::Experimental::RNTupleReader> m_file;
-  std::unique_ptr<ROOT::Experimental::RNTupleReader> m_metadata;
-
-private:
-
   /**
-   * Initialze the passed CategoryInfo by setting up the necessary branches,
-   * collection infos and all necessary meta data to be able to read entries
-   * with this name
+   * Initialize the given category by filling the maps with metadata information
+   * that will be used later
    */
   bool initCategory(const std::string& category);
 
+  /**
+   * Read and reconstruct the generic parameters of the Frame
+   */
   GenericParameters readEventMetaData(const std::string& name, unsigned entNum);
 
-  podio::version::Version m_fileVersion{0, 0, 0};
+  std::unique_ptr<ROOT::Experimental::RNTupleReader> m_metadata{};
+
+  podio::version::Version m_fileVersion{};
   DatamodelDefinitionHolder m_datamodelHolder{};
 
-  std::map<std::string, std::vector<std::unique_ptr<ROOT::Experimental::RNTupleReader>>> m_readers;
-  std::map<std::string, std::unique_ptr<ROOT::Experimental::RNTupleReader>> m_metadata_readers;
-  std::vector<std::string> m_filenames;
-  std::map<std::string, int> m_entries;
+  std::map<std::string, std::vector<std::unique_ptr<ROOT::Experimental::RNTupleReader>>> m_readers{};
+  std::map<std::string, std::unique_ptr<ROOT::Experimental::RNTupleReader>> m_metadata_readers{};
+  std::vector<std::string> m_filenames{};
 
-  std::map<std::string, std::vector<int>> m_collectionId;
-  std::map<std::string, std::vector<std::string>> m_collectionName;
-  std::map<std::string, std::vector<std::string>> m_collectionType;
-  std::map<std::string, std::vector<bool>> m_isSubsetCollection;
+  std::map<std::string, int> m_entries{};
+  std::map<std::string, unsigned> m_totalEntries{};
 
-  std::map<std::string, unsigned> m_totalEntries;
-  std::vector<std::string> m_availableCategories;
+  std::map<std::string, std::vector<int>> m_collectionId{};
+  std::map<std::string, std::vector<std::string>> m_collectionName{};
+  std::map<std::string, std::vector<std::string>> m_collectionType{};
+  std::map<std::string, std::vector<bool>> m_isSubsetCollection{};
+  std::vector<std::string> m_availableCategories{};
 
 };
 
--- a/include/podio/ROOTNTupleReader.h
+++ b/include/podio/ROOTNTupleReader.h
@@ -4,6 +4,7 @@
 #include "podio/CollectionBranches.h"
 #include "podio/ICollectionProvider.h"
 #include "podio/ROOTFrameData.h"
+#include "podio/SchemaEvolution.h"
 #include "podio/podioVersion.h"
 #include "podio/utilities/DatamodelRegistryIOHelpers.h"
 
@@ -88,6 +89,7 @@ private:
     std::vector<std::string> name{};
     std::vector<std::string> type{};
     std::vector<short> isSubsetCollection{};
+    std::vector<SchemaVersionT> schemaVersion{};
   };
 
   std::unordered_map<std::string, CollectionInfo> m_collectionInfo{};
--- a/include/podio/ROOTNTupleReader.h
+++ b/include/podio/ROOTNTupleReader.h
@@ -85,7 +85,7 @@ private:
   std::unordered_map<std::string, unsigned> m_totalEntries{};
 
   struct CollectionInfo {
-    std::vector<int> id{};
+    std::vector<unsigned int> id{};
     std::vector<std::string> name{};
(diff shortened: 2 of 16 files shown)


tmadlener:
`clang-tidy` in pre-commit workflow is failing because the environment doesn't come with a new enough ROOT version. We will have to think about something there.

include/podio/ROOTNTupleReader.h
```diff
@@ -0,0 +1,95 @@
...
+   */
+  GenericParameters readEventMetaData(const std::string& name, unsigned entNum);
+
+  std::unique_ptr<ROOT::Experimental::RNTupleReader> m_metadata{};
+
+  podio::version::Version m_fileVersion{};
+  DatamodelDefinitionHolder m_datamodelHolder{};
+
+  std::map<std::string, std::vector<std::unique_ptr<ROOT::Experimental::RNTupleReader>>> m_readers{};
+  std::map<std::string, std::unique_ptr<ROOT::Experimental::RNTupleReader>> m_metadata_readers{};
```
tmadlener, line 79:
```suggestion
  std::unordered_map<std::string, std::vector<std::unique_ptr<ROOT::Experimental::RNTupleReader>>> m_readers{};
  std::unordered_map<std::string, std::unique_ptr<ROOT::Experimental::RNTupleReader>> m_metadata_readers{};
```
Mainly for consistency with the rest of podio, where we usually have `unordered_map` instead of `map`. Also might be slightly faster depending on the number of elements.

There are also a few more below, I just saw.

include/podio/CollectionBuffers.h
```diff
@@ -25,6 +25,7 @@ using VectorMembersInfo = std::vector<std::pair<std::string, void*>>;
  */
 struct CollectionWriteBuffers {
   void* data{nullptr};
+  void* vecPtr{nullptr};
```
tmadlener, line 28: Not entirely happy with this one. But at the moment we seem to not have any other option(?).

include/podio/ROOTNTupleReader.h
```diff
@@ -0,0 +1,95 @@
...
+  std::vector<std::string> m_filenames{};
+
+  std::map<std::string, int> m_entries{};
+  std::map<std::string, unsigned> m_totalEntries{};
+
+  std::map<std::string, std::vector<int>> m_collectionId{};
+  std::map<std::string, std::vector<std::string>> m_collectionName{};
+  std::map<std::string, std::vector<std::string>> m_collectionType{};
+  std::map<std::string, std::vector<bool>> m_isSubsetCollection{};
```
tmadlener, line 88:
```suggestion
  std::map<std::string, std::vector<short>> m_isSubsetCollection{};
```
`vector<bool>` has a few quirks that are not worth the benefits usually, IMHO.

src/CMakeLists.txt
```diff
@@ -79,15 +82,32 @@ SET(root_sources
   ROOTFrameReader.cc
   ROOTLegacyReader.cc
 )
+if(ENABLE_RNTUPLE)
+  set(root_sources
+      ${root_sources}
```
tmadlener, line 87:
```suggestion
  list(APPEND root_sources
```
Is the more idiomatic approach in CMake, I think.

src/ROOTNTupleReader.cc
```diff
@@ -0,0 +1,245 @@
...
+#include "podio/GenericParameters.h"
+#include "rootUtils.h"
+
+#include "TClass.h"
+#include <ROOT/RError.hxx>
+#include <memory>
+
+namespace podio {
+
+GenericParameters ROOTNTupleReader::readEventMetaData(const std::string& name, unsigned entNum) {
```
tmadlener, line 14: I think this could be made less repetitive, via a templated function that handles one type, the help of `std::apply` and the `SupportedGenericDataTypes`. That would require to make the `getMap` methods public in `GenericParameters`, but they effectively are already.

include/podio/ROOTNTupleWriter.h
```diff
@@ -0,0 +1,57 @@
...
+
+  std::unique_ptr<TFile> m_file{};
+
+  DatamodelDefinitionCollector m_datamodelCollector{};
+
+  std::map<std::string, std::vector<int>> m_collectionId{};
+  std::map<std::string, std::vector<std::string>> m_collectionName{};
+  std::map<std::string, std::vector<std::string>> m_collectionType{};
+  std::map<std::string, std::vector<bool>> m_isSubsetCollection{};
+  std::set<std::string> m_categories{};
```
tmadlener, line 49: I suppose many of these maps (and a few of the above ones) are here to support writing multiple categories? Wouldn't it be simpler to have a small(-ish) helper struct that bundles all the information for one category and one map for name -> struct? Similar to what is done in the ROOTFrameWriter/Reader?

src/ROOTNTupleReader.cc
```diff
@@ -0,0 +1,245 @@
...
+  // the names following the order of the IDs
+  std::vector<std::pair<int, std::string>> v;
+  v.reserve(names.size());
+  for (size_t i = 0; i < names.size(); ++i) {
+    v.emplace_back(std::make_pair<int, std::string>(int(ids[i]), std::string(names[i])));
+  }
+  std::sort(v.begin(), v.end());
+  for (auto& [id, name] : v) {
+    table->add(name);
+  }
```
tmadlener, line 240: This should be unnecessary, you can construct the `CollectionIDTable` directly from these vectors:
https://github.com/AIDASoft/podio/blob/c7328d62f17061c8360eb2ac0a3fe0ce931fd2ae/include/podio/CollectionIDTable.h#L22-L25

Additionally, for a given category this information will never change, so the `shared_ptr` could be read only once and then kept in the reader itself, rather than building it for every read.

src/rootUtils.h
```diff
@@ -30,6 +30,41 @@ constexpr static auto metaTreeName = "podio_metadata";
...
+
+constexpr static auto intValue = "GPIntValues";
+constexpr static auto floatValue = "GPFloatValues";
+constexpr static auto doubleValue = "GPDoubleValues";
+constexpr static auto stringValue = "GPStringValues";
+
+/**
+ * Name of the field with the list of categories for RNTuples
+ */
+constexpr static auto availableCategories = "availableCategories";
```
tmadlener, line 50: I suppose this needs to be stored with RNTuples, because there is no easy way (yet) to get the "branches" from the file?

tmadlener (commented):
Very nice work, thanks for this. My questions/comments are either minor things, or some design questions. I guess the most important one is, whether it would make sense to introduce similar helper structs as for the TTree versions to keep category information together and remove a few of the maps that all have the same keys at the moment (as far as I understood).

Another one, would be the repetition around the GenericParameter reading/writing, where we should be able to play some template games to drastically reduce it (and additionally make it almost automatic if there should ever by a new type added to them).

src/rootUtils.h
jmcarcell: Exactly, at least I couldn't find an easy one

include/podio/ROOTNTupleReader.h
jmcarcell: Fixed

include/podio/CollectionBuffers.h
jmcarcell: Not yet, it's in my todo list

include/podio/ROOTNTupleReader.h
tmadlener: Fair point. We will have to do some profiling and benchmarking at some point. Since these are internals in any case switching at a later point should be straight forward.

src/ROOTNTupleReader.cc
jmcarcell: Fixed

include/podio/ROOTNTupleWriter.h
jmcarcell: Now they are grouped, with the possibility of grouping a few more things that are used in the reader too

src/ROOTNTupleReader.cc
jmcarcell: Implemented

jmcarcell:
I have implemented almost all the comments. The simplification for writing and reading the generic parameters adds a couple of `constexpr` to `rootUtils` (to get the "branch" name for each parameter) and also makes one `getMap` function public instead of private

src/ROOTNTupleReader.cc
```diff
@@ -0,0 +1,212 @@
...
+
+namespace podio {
+
+template<typename T>
+void ROOTNTupleReader::readParams(const std::string& name, unsigned entNum, GenericParameters& params) {
+  auto keyView   = m_readers[name][0]->GetView<std::vector<std::string>>(root_utils::getGPKey<T>());
+  auto valueView = m_readers[name][0]->GetView<std::vector<std::vector<T>>>(root_utils::getGPValue<T>());
+
+  auto keys = keyView(entNum);
+  auto values = valueView(entNum);
```
tmadlener, line 20: Just for my understanding. These are really just views to the data, right? There is no way we can "steal" them without destroying things. I.e. we cannot simply read this things once and then move them into the `GenericParameter` maps?

src/ROOTNTupleReader.cc
```diff
@@ -0,0 +1,212 @@
...
+  auto edmView = m_metadata->GetView<std::vector<std::tuple<std::string, std::string>>>(root_utils::edmDefBranchName);
+  auto edm = edmView(0);
+
+  auto availableCategoriesField = m_metadata->GetView<std::vector<std::string>>(root_utils::availableCategories);
+  m_availableCategories = availableCategoriesField(0);
+
+}
+
+unsigned ROOTNTupleReader::getEntries(const std::string& name) {
+  if (m_readers.find(name) == m_readers.end()) {
```
tmadlener, line 89: Can we cache this after the first calculation for a category? Otherwise constructs like the following could become costly
```cpp
for (auto i = 0u; i < reader.getEntries("events"); ++i) {
  // ... do something
}
```
Especially also because this is not even marked `const` so the compiler will most likely not even be able to hoist this to the outside of the loop.

src/ROOTNTupleReader.cc
```diff
@@ -0,0 +1,212 @@
...
+      return nullptr;
+    }
+  }
+
+  m_entries[category] = entNum+1;
+
+  ROOTFrameData::BufferMap buffers;
+  auto dentry = m_readers[category][0]->GetModel()->GetDefaultEntry();
+
+  std::map<std::pair<std::string, int>, std::vector<podio::ObjectID>*> tmp;
```
tmadlener, line 126: If I understand correctly here: This map of `vector<ObjectID>*` is necessary, because we cannot directly "connect" the `buffers.references[i].get()` to the `vector*` that is expected in `CaptureValueUnsafe`?

If that is the case, could you put a small comment here stating why we need this map?

src/ROOTNTupleWriter.cc
```diff
@@ -0,0 +1,209 @@
...
+    finish();
+  }
+}
+
+template<typename T>
+void ROOTNTupleWriter::fillParams(const std::string& category, GenericParameters& params) {
+  auto gpKeys = m_writers[category]->GetModel()->Get<std::vector<std::string>>(root_utils::getGPKey<T>());
+  auto gpValues = m_writers[category]->GetModel()->Get<std::vector<std::vector<T>>>(root_utils::getGPValue<T>());
+  gpKeys->clear();
+  gpValues->clear();
```
tmadlener, line 32: Maybe also add a `reserve(params.getMap<T>().size()` here? (Although it probably doesn't matter too much apart from the first event).

src/rootUtils.h
```diff
@@ -30,6 +30,78 @@ constexpr static auto metaTreeName = "podio_metadata";
...
+constexpr static auto intValue = "GPIntValues";
+constexpr static auto floatValue = "GPFloatValues";
+constexpr static auto doubleValue = "GPDoubleValues";
+constexpr static auto stringValue = "GPStringValues";
+
+/**
+ * Get the name of the key depending on the type
+ */
+template <typename T>
+constexpr auto getGPKey() {
```
tmadlener, line 51:
```suggestion
constexpr auto getGPKeyName() {
```
I was slightly confused at first how this would get keys without a connection to the map while reading things at the calling site.

src/rootUtils.h
```diff
@@ -30,6 +30,78 @@ constexpr static auto metaTreeName = "podio_metadata";
...
+  } else {
+    static_assert(sizeof(T) == 0, "Unsupported type for generic parameters");
+  }
+}
+
+/**
+ * Get the name of the value depending on the type
+ */
+template <typename T>
+constexpr auto getGPValue() {
```
tmadlener, line 69:
```suggestion
constexpr auto getGPValueName() {
```

src/rootUtils.h
```diff
@@ -30,6 +30,78 @@ constexpr static auto metaTreeName = "podio_metadata";
  */
 constexpr static auto paramBranchName = "PARAMETERS";
 
+/**
+ * Names of the fields with the keys and values of the generic parameters for
+ * the RNTuples until map types are supported
+ */
+constexpr static auto intKey = "GPIntKeys";
```
tmadlener, line 37:
```suggestion
constexpr static auto intKeyName = "GPIntKeys";
```
mainly consistency since I think all previously existing branch/writer name strings have `Name` as their ending.

tmadlener (commented):
Thanks. Mainly a few questions for my understanding and minor variable/function name nitpicking for better readability.

Making the second `getMap<T>` public for the GenericParameters is not really an issue, since they have effectively been public before as well.

src/ROOTNTupleReader.cc
jmcarcell: Yep. Do you mean reading them at once for all `entNum` and then moving? I don't think that would change anything `values` here is already a vector and we are writing in a `map<string, vector<T>>`. But one thing that can be done is to cast to` an `rvalue` so that `values` can be moved in the next line:
```
  for (size_t i = 0; i < keys.size(); ++i) {
    params.getMap<T>()[keys[i]] = std::move(values[i]);
  }
```

src/ROOTNTupleReader.cc
jmcarcell: Done

src/ROOTNTupleWriter.cc
jmcarcell: Done, yeah I think it would not be measurable unless you have lots and lots (4 or 5 digits?) of collections

src/rootUtils.h
jmcarcell: Done

src/ROOTNTupleReader.cc
tmadlener: Yeah, my thinking was that we could provide some functionality for `GenericParameters` that simply takes the keys `vector<string>&&` and the values `vector<vector<T>>&&` and then we move them into that function and inside that function we would move the individual elements into the internal map. But maybe something like:

```cpp
params.getMap<T>().emplace(std::move(keys[i]), std::move(values[i]));
```

would be good enough. In this way we would also move the keys instead of copying them. Anyhow, all of this only works if we don't accidentally steal the things from the RNTuple view. I am not sure about that, maybe we have to make copies in any case.

src/ROOTNTupleReader.cc
jmcarcell: This is related to `buffers.references[i].get()` being a `unique_ptr` I think. I'll double check but I think passing either the `unique_ptr` or the `unique_ptr.get()` both crash and that's why I had to choose that not very elegant solution since I don't know better :shrug:

tmadlener: Hmm, pity. Would have been nice if the `get()` version would have worked. But if both crash, then this is the way it has to be for the moment.

src/ROOTNTupleReader.cc
jmcarcell: I checked and the values are being moved inside the function because the vectors are left empty but if you call `GetView` again then you get the values again which means that at least one copy is being made. So I'm actually not sure the moves will do anything. But I guess there is no downside to `std::move`. I guess we are only changing when the copies are made

tmadlener: Yeah, I suppose the `GetView` will have to do some reading and potentially uncompressing. Maybe it skips this if the thing it views is non-empty. In any case, since we call `GetView` only once per event, and we don't break things by `move`-ing, I agree that we keep `std::move` here and at least avoid unnecessary copies.

jmcarcell: Ok this is done now moving both keys and values, I also removed the intermediate variables since they are not really used more times than that one

jmcarcell:
Ok now this should be properly formatted

jmcarcell:
Updated after #394

jmcarcell:
Writer updated to use a bare model from the comments from Jakob. It isn't that elegant since not everything was being written through `void*`, although when support for `std::map` is there it will be simplified. I have tried to do a similar thing for the reader but it doesn't seem to be trivial since the tests end up failing later on

src/ROOTNTupleWriter.cc
```diff
@@ -0,0 +1,233 @@
...
+        entry->CaptureValueUnsafe(brName, c.get());
+      }
+    }
+
+    if (auto vmInfo = collBuffers.vectorMembers) {
+      int i = 0;
+      for (auto& [type, vec] : (*vmInfo)) {
+        const auto typeName = "vector<" + type + ">";
+        const auto brName = root_utils::vecBranch(name, i++);
+        auto ptr = *(std::vector<int>**)vec;
```
tmadlener, line 107:
```suggestion
        const auto brName = root_utils::vecBranch(name, i++);
        auto ptr = *(std::vector<int>**)vec;
```
The `typeName` seems to be unused, so I suppose this works because the necessary type information has been passed to the whole thing when we create the fields? And the `vector<int>` cast works for all types that potentially come along here and is simply necessary for getting to the correct `void*`?

tmadlener (commented):
Thanks for looking into this. Does this mean we can drop one of the `void*` in the CollectionBuffers? Or do we have to keep them around for now? If there is no easy way around it, then I would probably be OK with merging this for now and properly documenting that the buffers are meant for internal use only and that they are subject to changes.

I once more also have a question about some RNTuple details that I missed before.

src/ROOTNTupleWriter.cc
jmcarcell: Yeah I don't think there is a typename for the other members, must have forgotten to remove it when copying and pasting from the TTree reader. I should double check that it works for every type, but in principle it's the same three star cast used in other places: https://github.com/AIDASoft/podio/blob/930e6001e893bef9415639710e12b67b5deba5d6/include/podio/CollectionBuffers.h#L41

jmcarcell:
No way of getting rid of the extra pointer that I know of, the last changes were to change how the RNTuple model was created, now it is a bare one which doesn't do instantiation (before it was a default one which does). This is unrellated to the `void*` pointers.

hegner:
@jmcarcell - can you update your PR ? would like to see whether we can merge today

tmadlener:
clang-tidy seems to break, because the root version in that LCG release is not yet recent enough: https://github.com/AIDASoft/podio/actions/runs/5386875055/jobs/9777453794?pr=395#step:4:416

I think now that we have LLVM in the key4hep stack we could also switch to that for the CI. However, we should do that in a separate PR, after we have merged this. 

There are also some clang-format complaints. I suppose these are due to different versions running on your machine and the CI(?). Can you fix those? That would make it easier to have one global commit for when we switch to a newer version of clang-format.

jmcarcell:
Now `clang-format` is happy

hegner:
@jmcarcell - can you spell out the branch name issues we discussed yesterday? Thanks!

jmcarcell:
Updated now to conform to #405, which was the last thing pending, at least that I remember. The reader needs some cleanup of these changes but I think that will have to wait for tomorrow :sleepy:
Some test results on my PC since I have them handy:
``` 
    Start 15: write_rntuple
1/2 Test #15: write_rntuple ....................   Passed    0.17 sec
    Start 16: read_rntuple
2/2 Test #16: read_rntuple .....................   Passed    0.63 sec
```
writing and reading the same frame:
```
    Start 12: write_frame_root
1/2 Test #12: write_frame_root .................   Passed    1.25 sec
    Start 11: read_frame_root
2/2 Test #11: read_frame_root ..................   Passed    1.26 sec

```

hegner:
@jmcarcell - thanks a lot! Do you think that's it for now? If yes, I will redo the review

hegner:
@jmcarcell - thanks a lot. I think it is fine as it is now. We will do the cleanups in another PR then :-)

PR #397, opened 2023-03-23
Do not mark generated files as GENERATED

### LINKED ISSUE #396: Generated cc files are deleted on cmake --target clear
- OS version: e.g. Debian Mint (Ubuntu) 22.04
 - Compiler version: gcc 11.3, cmake version 3.26.0
 - PODIO version: master at c7328d62f17061c8360eb2ac0a3fe0ce931fd2ae
 - Reproduced by: 
 
 ```bash
 # on project that generates podio model: 
 
cmake -B build ./                   # configure
cmake --build build --target all    # build 
cmake --build build --target clean  # clean 
cmake --build build --target all    # <= will never build again
```
 
 - Goal: 

On configuration stage cmake generates files for data model:

* datamodel/(headers)
* src/(.cc)
* podio_generated_files.cmake

When one runs 

```bash
cmake --target clean
```

Clean removes all .cc files but leaves everything else. 

Since file generation is done only on the first configuration stage I would prefer that `clean` would not touch any of generated files. In my development workflow which I believe to be common, in terms of build steps I do something like this: 

```bash
# 
configure
build
build
clean   # <= the code will not compile after this, as files are removed and new `configure` is needed 
build
clean
build
...
```

Also IDE-s usually have an option like "Rebuild project" which is "clean+build" for projects with podio data model that will always fail.

tmadlener:
BEGINRELEASENOTES
- Remove the `GENERATED` property from generated files in CMake to avoid inconsistent removal of headers and source files with the `clean` target. Fixes #396 

ENDRELEASENOTES

@andresailer do you have any better idea for this? There was quite some discussion already in #143 but nothing too specific about the `GENERATED` property.

@DraTeots could you give this a go and check whether this indeed solves your issue?

### DIFF
--- a/python/podio_class_generator.py
+++ b/python/podio_class_generator.py
@@ -464,7 +464,6 @@ have resolvable schema evolution incompatibilities:")
         list_cont.append(f'  {os.path.join(target_folder, fname)}')
 
       list_cont.append(')')
-      list_cont.append(f'SET_PROPERTY(SOURCE ${{{name}}} PROPERTY GENERATED TRUE)\n')
 
       return '\n'.join(list_cont)
 


DraTeots:
I fully fixes my issue. Thank you!!!

tmadlener:
@jmcarcell just to make sure, this also doesn't break any of your workflows?

jmcarcell:
It doesn't break anything for me because I don't use `clean`. But with cmake 3.25.2 I can't reproduce the original issue, in my case neither the headers nor the .cc files are deleted when running `clean`. Also using the version in this PR nothing is deleted. There is something weird about this because the documentation says that:
```
The [Makefile Generators](https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html#makefile-generators) will remove GENERATED files during make clean.
```
so it should delete both headers and .cc files (master)?. I think that anyway it's better to remove this option since it doesn't say anything for other generators so maybe there is different behaviour for `make` and `ninja` for example

PR #398, opened 2023-03-24
Do not reject build if ROOT is built with C++20

paulgessinger:
BEGINRELEASENOTES
- Do not reject building if ROOT was built with C++20 (instead of C++17).

ENDRELEASENOTES

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,8 +77,8 @@ find_package(ROOT REQUIRED COMPONENTS RIO Tree)
 
 # Check that root is compiled with a modern enough c++ standard
 get_target_property(ROOT_COMPILE_FEATURES ROOT::Core INTERFACE_COMPILE_FEATURES)
-if (NOT "cxx_std_17" IN_LIST ROOT_COMPILE_FEATURES)
-  message(FATAL_ERROR "You are trying to build podio against a version of ROOT that has not been built with a sufficient c++ standard. podio requires c++17")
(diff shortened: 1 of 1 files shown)


jmcarcell:
Can ROOT be built with C++20? And in case it can, does podio build and work fine?

paulgessinger:
I believe ROOT v6.28.02 has at least experimental support. The podio build itself runs through without problems it seems, but I haven't done a lot of downstream testing yet.

jmcarcell:
> EDIT: No official support yet as per https://root-forum.cern.ch/t/root-v6-28-02-is-out/54200. But should podio really reject the build altogether until ROOT has full C++20 support?
I didn't say that, I just wanted to know the current status. I can test if podio works fine after ROOT finishes compiling

jmcarcell:
So it seems with ROOT with c++17 and podio c++20 it builds fine and tests work. But this is already allowed, what this PR would change is when ROOT has C++20 support trying to build podio with c++20 (but it isn't the case yet or at least I can confirm ROOT with c++20 doesn't build). I guess it can go in and we are prepared for c++20 when ROOT has it, what do you think @tmadlener ?

tmadlener:
Thanks for checking @jmcarcell. Good to know that podio with c++20 works against ROOT c++17. Given that it apparently also builds against ROOT with (experimental) c++20, I think this can go in. We will probably have to fine-tune this a bit to cover all possibilities once ROOT has proper c++20 support, as I am not sure ROOT with c++20 and podio with c++17 is a working combination.

PR #399, opened 2023-03-24
Define PODIO_ENABLE_SIO for targets depending on podioSioIO

tmadlener:
BEGINRELEASENOTES
- Add `PODIO_ENABLE_SIO=1` to the public `target_compile_definitions` for `podioSioIO` so that all dependent targets automatically get it as well. This should make it easier to use SIO dependent features in dependencies. 
- Consistently use a scope for `target_link_libraries` in tests.

ENDRELEASENOTES

 Since `podioSioIO` is the main target from podio that will always be necessary to link against when using SIO features this should cover all possible cases (as long as CMake is used to build dependencies).

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -125,10 +125,10 @@ if(ENABLE_SIO)
 
   PODIO_ADD_LIB_AND_DICT(podioSioIO "${sio_headers}" "${sio_sources}" sio_selection.xml)
   target_link_libraries(podioSioIO PUBLIC podio::podio SIO::sio ${CMAKE_DL_LIBS} ${PODIO_FS_LIBS})
+  target_compile_definitions(podioSioIO PUBLIC PODIO_ENABLE_SIO=1)
 
   # Make sure the legacy python bindings know about the SIO backend
   target_link_libraries(podioPythonStore PRIVATE podioSioIO)
-  target_compile_definitions(podioPythonStore PRIVATE PODIO_ENABLE_SIO=1)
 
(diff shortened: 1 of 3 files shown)


src/CMakeLists.txt
```diff
@@ -125,10 +125,10 @@ if(ENABLE_SIO)
 
   PODIO_ADD_LIB_AND_DICT(podioSioIO "${sio_headers}" "${sio_sources}" sio_selection.xml)
   target_link_libraries(podioSioIO PUBLIC podio::podio SIO::sio ${CMAKE_DL_LIBS} ${PODIO_FS_LIBS})
+  target_compile_definitions(podioSioIO PUBLIC PODIO_ENABLE_SIO=1)
```
jmcarcell, line 128: Does it need to have a value?

tmadlener: Yes, there are places where we depend on the value, e.g.
https://github.com/AIDASoft/podio/blob/9cd03731dfe4b4c37eac34e6364411bb55be940a/src/IOHelpers.cc#L26-L28

I think it is the cleaner solution to define this to a non-false value rather then depending on whether it is simply defined. I could also change the usage in the unittests, to reflect that.

Additionally, from the [cmake documentation](https://cmake.org/cmake/help/latest/command/target_compile_definitions.html) it seems we can be explicit here rather than implicit without changing things:
> Definitions may optionally have values:
> 
> ```cmake
> target_compile_definitions(foo PUBLIC FOO=1)
> ```
> Note that many compilers treat -DFOO as equivalent to -DFOO=1, but other tools may not recognize this in all circumstances

Issue #313, opened 2022-07-23
static version of getValueTypeName() and friends


faustus123:
It would be convenient in certain cases to have the following members of the type-specific collection classes be static:

getTypeName()
getValueTypeName()
getDataTypeName()

context:
In an application I'm writing I have a templated function that gets passed in a collection class as one of the template parameters. It is creating a collection of that type and partially deriving the name of the collection from what is returned from getValueTypeName(). A natural way to write this for a template parameter C where e.g. C=ClusterCollection would be:

auto name = C::getValueTypeName() + suffix;

As it currently is, I have to do this:

C tmp;
auto name = tmp.getValueTypeName() + suffix;

If you really want those methods to be overridden by subclasses, then maybe you could consider adding static versions to supplement the class for these types of scenarios.

hegner:
Hi @faustus123 - I think it makes perfect sense to make the information statically available.

PR #402, opened 2023-04-04
Add static constexpr char* names and hand out views for them

tmadlener:
BEGINRELEASENOTES
- Add public `static constexpr char*` type names to the collections and make the `getXXXName()` methods return `string_view`s to these strings. **This is a breaking change to the interface of the collections if you explicitly rely on them being `std::string`**
  - `typeName`: the full type name of the collection (returned also by `getTypeName`)
  - `valueTypeName`: the (immutable) type name of the objects of the collection (returned by `getValueTypeName`)
  - `dataTypeName`: the type name of the data PODs (returned by `getDataTypeName`)
- Make unittest environment properly use `PODIO_SIOBLOCK_PATH`
- `USE_EXTERNAL_CATCH2` now can also be set to `AUTO` to look for a suitable version of Catch2 before falling back and fetching and building it's own version instead of a hard fail.

ENDRELEASENOTES

Addresses some of the things raised in #313 and is also in preparation for other work of making run-time type information available without too many `std::string` copies

### DIFF
--- a/.github/workflows/key4hep.yml
+++ b/.github/workflows/key4hep.yml
@@ -30,7 +30,7 @@ jobs:
             -DCMAKE_INSTALL_PREFIX=../install \
             -DCMAKE_CXX_STANDARD=17 \
             -DCMAKE_CXX_FLAGS=" -fdiagnostics-color=always -Werror -Wno-error=deprecated-declarations " \
-            -DUSE_EXTERNAL_CATCH2=ON \
+            -DUSE_EXTERNAL_CATCH2=AUTO \
             -G Ninja ..
           echo "::endgroup::"
           echo "::group::Build"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -142,7 +142,8 @@ endif()
 
 #--- enable unit testing capabilities ------------------------------------------
 include(CTest)
-option(USE_EXTERNAL_CATCH2 "Link against an external Catch2 v3 static library, otherwise build it locally" ON)
+set(USE_EXTERNAL_CATCH2 AUTO CACHE STRING "Link against an external Catch2 v3 static library, otherwise build it locally")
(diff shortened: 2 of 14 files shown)


tmadlener:
This will break DD4hep for example.

tmadlener:
@faustus123 @nathanwbrei @wdconinc this still needs a bit of work, but potentially as a heads up that this might break things on your end. Would you prefer to have deprecation warnings (as implemented in #404) to flush out uses without breaking things first? Otherwise we might just go ahead and check and fix the packages that we build regularly and introduce everything with a bunch of "synchronized" PRs.

wdconinc:
I'll give it a shot with #404 and see how much we should change beyond our `datamodel_glue.h` header. I'm expecting it to be minimal, though.

nathanwbrei:
Thanks for the heads-up, @tmadlener! It looks like this affects us pretty minimally.

tmadlener:
@nathanwbrei @wdconinc would you like to have proper deprecation warnigs and then a switch afterwards, or could we switch directly as done in this PR?

wdconinc:
After https://github.com/eic/EICrecon/pull/618 we shouldn't be affected by this change.

tmadlener:
CI currently fails on anything that tries to load an SIO block library in the environment that has not yet gone through this switch (i.e. EDM4hep from the underlying environment). Apart from a proper plugin mechanism, I am not sure we can easily do anything against this.

tmadlener:
With the latest changes to the cmake config, `PODIO_SIOBLOCK_PATH` should now be properly passed to the unittest environments. This requires cmake >= 3.22 to work with catch test discovery. If that is not met we fall back to not run test discovery, and all unittest will be reported collectively as `unittest`

hegner:
much better!

PR #404, opened 2023-04-12
Introduce warnings about upcoming signature change

tmadlener:
BEGINRELEASENOTES
- Add a deprecation warning for podio::CollectionBase::getTypeName` and friends to warn about the upcoming (#402) change in return type. 

ENDRELEASENOTES

As discussed during the meeting on Apr 11 in order to give a heads up to users.


hegner (approved):
I would have used DEPR_STRING_SIGNATURE, since the file is very widely visible. But fine as it is for a temporary warning.

tmadlener:
It looks like #402 can be merged without deprecating first. Closing this.

PR #405, opened 2023-04-13
Make branch names for relations more legible for ROOT based I/O

### LINKED ISSUE #169: Possibility of more intelligible branch names for relations

tmadlener:
BEGINRELEASENOTES
- Make the branch names for relations and vector members more legible and valid c++ variable names to improve interoperability with RDataFrame. Fixes #169 
  - The branch names will have the following structure: `_<collection-name>_<relation-name>`, resp. `_<collection-name>_<vectormember-name>`, where `relation-name`, resp.`vectormember-name` are taken from the YAML definitions.
  - Subset collections will have a single branch with `<collection-name>_objIdx`. This makes it easier to disambiguate them from normal collections.

ENDRELEASENOTES

- [ ] Make this backward compatible on the reading side
  - [x] Make reader capable
  - [x] Add test. This needs an input file -> Moved to #423

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,7 @@ project(podio)
 #--- Version -------------------------------------------------------------------
 SET( ${PROJECT_NAME}_VERSION_MAJOR 0 )
 SET( ${PROJECT_NAME}_VERSION_MINOR 16 )
-SET( ${PROJECT_NAME}_VERSION_PATCH 5 )
+SET( ${PROJECT_NAME}_VERSION_PATCH 6 )
 
 SET( ${PROJECT_NAME}_VERSION  "${${PROJECT_NAME}_VERSION_MAJOR}.${${PROJECT_NAME}_VERSION_MINOR}.${${PROJECT_NAME}_VERSION_PATCH}" )
 
--- a/include/podio/CollectionBase.h
+++ b/include/podio/CollectionBase.h
@@ -53,6 +53,9 @@ public:
   /// number of elements in the collection
   virtual size_t size() const = 0;
 
+  /// Get the names of the relations and vector members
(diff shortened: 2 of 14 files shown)


hegner:
@tmadlener - what is the status about the testing and the test file?

tmadlener:
I have to figure out who has the rights to upload things to where we already host a legacy test file:
https://github.com/AIDASoft/podio/blob/76c98a6e76ec25d37e1e726fb5afa432d84e8420/tests/scripts/get_test_inputs.sh#L11

After that it is hopefully just placing a file produced with a previous version of podio there and integrating things into the tests.

andresailer:
> I have to figure out who has the rights to upload things to where we already host a legacy test file:
> 
Soon you should be able to...

PR #412, opened 2023-05-04
Make CollectionIDs a 32bit hash value of the collection name

### LINKED ISSUE #381: CollectionIDs should not depend on the insertion order
Currently the order of insertion into a `Frame` determines the `collectionID`. This is extremely unstable, especially in multi threaded contexts, but also in cases where collections might be missing. The `collectionID` should be something stable, even across different processes and should ideally only depend on the collection name. In principle hashing the name should be good enough for this, however with only 32 bits available at the moment collisions are not unlikely, and even with 64 bits there should probably still be a way to handle collisions.

hegner:
BEGINRELEASENOTES
- Using string hashes as CollectionID based on MurmurHash

ENDRELEASENOTES

This PR is work in progress. The frame interface works. However the legacy interface used the previous structure of IDs for some optimizations, which I have to remove.


src/MurmurHash2.cpp
```diff
@@ -0,0 +1,523 @@
...
+
+    data += 4;
+    len -= 4;
+  }
+
+  // Handle the last few bytes of the input array
+
+  switch(len)
+  {
+  case 3: h ^= data[2] << 16;
```
tmadlener, line 72: It looks like our warnings don't like the implicit fall throughs (which are quite intended in this case), which makes CI fail at the moment:
https://github.com/AIDASoft/podio/actions/runs/4884987514/jobs/8718460817?pr=412#step:4:338

Probably need to add a few [`[[fallthrough]]`s ](https://en.cppreference.com/w/cpp/language/attributes/fallthrough) in a few places.

hegner: in fact none of the cases was meant as fallthrough, so a few breaks were needed

tmadlener: I think they are meant to fall through to handle the last few bytes of the input that were not handled in the loops that take care of 8 (or 4) bytes above these statements? If they do not fall through, the last (up-to) 7 (or 3) bytes will be ignored in the hashing, I think?

hegner: You are correct. I made a mistake there which I found in the night. Will revert

tmadlener:
How does this fare with reading files prior to this? It should work because we just replace the calculation of the ID, right? Can we easily verify that (in CI)?

hegner:
@tmadlener - on fixing the undetected narrowing from 64 to 32 bit I found quite a few signed/unsigned inconsistencies which haven't been caught before. the collectionID is now uint64_t throughout

jmcarcell:
What's the final take on the hash size, 64 bits? With 64 bits there is some small collision probability with many collections, right?

hegner:
Yes. 64bit hash size. 32 bit was too little to be safe from collisions for a huge number of collections. With 64 bits we are in very safe territory

hegner:
See e.g. here https://preshing.com/20110504/hash-collision-probabilities/

include/podio/ObjectID.h
```diff
@@ -9,7 +11,7 @@ class ObjectID {
   /// index of object in collection
   int index;
   /// ID of the collection
-  int collectionID;
+  uint64_t collectionID;
```
tmadlener, line 14: Do these still work nicely with the `untracked` and `invalid` constants below?

src/selection.xml
```diff
@@ -18,6 +18,9 @@
     <class name="std::vector<std::tuple<int, std::string, bool, unsigned>>"/>
     <class name="std::vector<std::tuple<int, std::string, bool>>"/>
     <class name="std::vector<std::tuple<std::string, std::string>>"/>
+    <class name="std::vector<std::tuple<unsigned long, std::string, bool, unsigned int>>"/>
+    <class name="std::vector<std::tuple<unsigned long, std::string, bool, unsigned>>"/>
+    <class name="std::vector<std::tuple<unsigned long, std::string, bool>>"/>
```
tmadlener, line 23:
```suggestion
    <class name="std::vector<std::tuple<uint64_t, std::string, bool, unsigned int>>"/>
    <class name="std::vector<std::tuple<uint64_t, std::string, bool, unsigned>>"/>
    <class name="std::vector<std::tuple<uint64_t, std::string, bool>>"/>
```
Just to be more precise here.

tmadlener (commented):
Making the collectionID have 64 bits (and planning to use all of them) has potentially a few more implications, e.g. the `id` method will no longer be very useful:
https://github.com/AIDASoft/podio/blob/08117a1b5f8655aba4a576250db22dba1414c3e3/python/templates/macros/declarations.jinja2#L96

tmadlener:
Regarding backwards compatibility, for SIO we would need to increase the version number for the `SIOCollectionIDTableBlock` (and fix the current inconsistency):
https://github.com/AIDASoft/podio/blob/705721d53bcdd68c265351b86d4973c5bca3c0f9/include/podio/SIOBlock.h#L101-L108
https://github.com/AIDASoft/podio/blob/705721d53bcdd68c265351b86d4973c5bca3c0f9/src/SIOBlock.cc#L16

This could then be handled accordingly on the reading side:
https://github.com/AIDASoft/podio/blob/705721d53bcdd68c265351b86d4973c5bca3c0f9/src/SIOBlock.cc#L35-L42


There is also a missing conversion to `uint64_t` here. https://github.com/AIDASoft/podio/blob/705721d53bcdd68c265351b86d4973c5bca3c0f9/src/sioUtils.h#L49

hegner:
Thanks. The SIO I really didn't test properly yet.

include/podio/ObjectID.h
hegner: Yes. Because we check only on the ObjectID and there the index is more than enough

tmadlener:
This looks like it needs a rebase (and potentially some conflict resolution).

For me this looks good, the main points that we need discussion IMHO are:
- Are we OK with effectively increasing the ObjectID size by 50%
- An alternative (potentially additional) approach would be to effectively externalize the collection ID generation to the outside and make it possible to supply the Frame with a pre-populated CollectionIDTable and then either falling back to hashing or failing on collection names that are not present in the supplied external table.

hegner:
Yes, that would be something for the next EDM4hep meeting to discuss.

tmadlener:
There are some conflicts. I think they should resolve themselves with a rebase on to master

hegner:
rebase didn't do it

hegner:
@tmadlener thanks

tmadlener:
One of the sanitizer workflows seems to be picking up on #174 and `clang-tidy` is complaining about a few things in murmurhash3. They are in principle easily fixable, and since we are already changing it to make `clang-format` happy, I don't see a reason not to also fix these issues here: https://github.com/AIDASoft/podio/actions/runs/5093184726/jobs/9155485339?pr=412#step:4:659

tmadlener:
I added another tag to the failing test to ignore it in UndefinedBehavior sanitizer runs since it is picking up on a known issue and the test will be obsolete once the EventStore is removed in any case.

andresailer:
It is 64 bits now, or was I mistaken in changing the title of the PR?

jmcarcell:
During the last call one week ago it was agreed that it would be 32 bit and collisions would be dealt with due to concerns about increasing file sizes, personally I would like to see some numbers first in case dealing with collisions turns out to be some work...

tmadlener:
I have added a small utility tool that takes a list of collection names and checks for collisions using the chosen hash function. It should be fairly straight forward to come up with a list of all currently used collection names and see if we have collisions in there already.

Speaking of collisions @hegner, I think we could also use this PR to make collisions more visible in the Frame, currently this is effectively silent (and potentially incomplete):
https://github.com/AIDASoft/podio/blob/76c98a6e76ec25d37e1e726fb5afa432d84e8420/include/podio/Frame.h#L318-L327
which calls
https://github.com/AIDASoft/podio/blob/76c98a6e76ec25d37e1e726fb5afa432d84e8420/include/podio/Frame.h#L410-L427

hegner:
Do you want to push your little addition into the branch?

hegner:
@tmadlener - should we merge?

tmadlener:
To avoid the really stupid mistake of introducing collisions in already existing workflows / files, I would like to run the test tool on a list of "all" collection names that are currently in use. I have concatenated REC outputs of ILD and CLIC, k4SimDelphes outputs from FCC, and will also add the EIC names to the list. I will report back here as soon as I have the complete list.

tmadlener:
Alright, I collected this list of collection names: [unique_coll_names.txt](https://github.com/AIDASoft/podio/files/11654282/unique_coll_names.txt)
from the following sources:
- a REC file from the ILD standard reconstruction `Output_REC`
- a REC file from the CLIC standard reconstruction `Output_REC`
- the list of files that can be found in the miniDST files that were produced for the snowmass tutorials of the ILC (slide 8 of [this presentation](https://agenda.linearcollider.org/event/9264/contributions/48192/attachments/36863/57667/jlist_miniDST_2010623.pdf)
- The output names of the default configuration for k4SimDelphes
- The list of EIC [reconstruction output](https://github.com/eic/EICrecon/blob/main/src/services/io/podio/JEventProcessorPODIO.cc#L38-L152) as well as the output of `rootls -t` on a ddsim file provided by @wdconinc.

This currently yields 361 unique collection names which have no collisions, and puts us somewhere into the 1:10000 and 1:100000 collision probability range according to this table here:

image

tmadlener (approved):
I think (and hope) we have now considered pretty much all the collection names that are currently in use. I didn't find any collisions among them, so  at least for now 32 bits seem to be enough.

PR #413, opened 2023-05-04
Add schema evolution hooks to Frame

tmadlener:
BEGINRELEASENOTES
- Introduce `podio::SchemaEvolution` that can hold schema evolution functions and that offers an `evolveBuffers` method that does the schema evolution on these buffers before collections are created.
- Add hooks in `podio::Frame` to call this when collections are read from the FrameData.

ENDRELEASENOTES

This introduces a `podio::SchemaEvolution` registry that holds schema evolution functions and that offers an `evolveBuffers` method that takes `podio::CollectionReadBuffers`, a schema version and the type of the collection that will be constructed from these buffers and evolves them to the current schema version (potentially by doing nothing at all). **This will not yet introduce any machinery to actually also populate this registry, that will be another PR.**

TODO:
- [x] Needs #394 
- [x] Needs #402 
- [x] Release notes
- [x] documentation


include/podio/SchemaEvolution.h
```diff
@@ -2,13 +2,147 @@
...
+ * version to the current schema version. From the evolved buffers it is then
+ * possible to create collections.
+ *
+ * It is implemented as a singleton that is populated at the time shared
+ * datamodel libraries (or their schema evolution libraries) are loaded. It is
+ * assumed that this happens early on in the startup of any application, such
+ * that the registration still happens on a single thread. After this
+ * initialization evolutions can be done from multiple threads.
+ */
+class SchemaEvolution {
```
hegner, line 31: not sure why you talk about SchemaEvolution and SchemaEvolutionRegistry interchangeably. Maybe one should just make a global (and stupid) rename?

hegner (approved):
Besides a rename of SchemaEvolution into SchemaEvolutionRegistry I am fine with this. thanks!

include/podio/SchemaEvolution.h
tmadlener: Removed the registry bits again. They were leftovers from when it was not yet entirely clear to me what the name should be.

PR #414, opened 2023-05-07
Fix dumping of UserDataCollection

tmadlener:
BEGINRELEASENOTES
- Fix off-by-one error in `UserDataCollection::print` that caused the first element to be printed twice.

ENDRELEASENOTES


tmadlener:
The failing of the `dev3` based `ubuntu` workflow seems to be due to some upstream problems as far as I can see.

hegner (approved):

PR #416, opened 2023-05-17
Remove selection rules for classes that don't exist anymore

jmcarcell:
BEGINRELEASENOTES
- Remove selection rules for classes that don't exist anymore

ENDRELEASENOTES

Forgotten in #415. There are a few small warnings but it's very hard to see them between all the warnings about deprecated stuff. I think that should be all, I only found a reference in a [comment](https://github.com/AIDASoft/podio/blob/master/lcio/datalayout.yaml#L93)


tmadlener (approved):
Good catch.

Issue #411, opened 2023-05-03
ROOTFrameReader crash when reading multiple edm4hep files


lkosarz:
Podio crashes when reading multiple edm4hep.root files. However, it works fine for a single file.
ROOTFrameReader produces memory leaks when opening multiple files with a separate ROOTFrameReader instance per file - may be related.

 - OS version: Debian Linux 3.10.0-1160.71.1.el7.x86_64
 - Compiler version: gcc (Debian 12.2.0-14) 12.2.0
 - PODIO version: tag or commit ID, or GitHub branch
 - Reproduced by: ROOT 6.26/10, Geant4 11.1.1, cmake version 3.24.3, singularity, eic-shell, dd4hep-1.25.1, podio version = 0.16.3
 - exact steps to reproduce the problem: checkout, setup environment (Geant, ROOT versions, iLCSoft release), cmake, build, run...
 - Input: open multiple edm4hep.root files either by ROOTFrameReader::openFiles() or ROOTFrameReader::openFile()
 - Output: crash at the beginning, but works fine with single file and reads the data
 - Goal: Read multiple edm4hep.root files

tmadlener:
Thanks for reporting this. We will try to have a look as soon as possible, but it will potentially take until after next week since many of us will be at CHEP.

wdconinc:
@lkosarz @c-dilks Can you add a more reproducible example, maybe connected with a fake PR that modifies `tests/read-multiple.cpp` to use `ROOTFrameReader`, if that reproduces the issue?

tmadlener:
Thanks for providing a reproducer. We will have a look ASAP.

lkosarz:
Thank you! My initial test was with openFiles() and passing of a pre-defined vector, so not on the fly. I did see the same crash happening in both openFiles() and multiple openFile() cases. I did also see memory leaks, but haven't checked it for this standalone code, I sent you. Maybe we should double-check for these too.

PR #417, opened 2023-05-19
Fix reading of multiple files with ROOTFrameReader

tmadlener:
BEGINRELEASENOTES
- Fix an issue with reading multiple files via the `ROOTFrameReader` (#411)
  - Add documentation for API of opening file(s)
  - Add tests for reading multiple files

ENDRELEASENOTES

- [x] Add tests covering this.

### DIFF
--- a/include/podio/ROOTFrameReader.h
+++ b/include/podio/ROOTFrameReader.h
@@ -147,7 +147,7 @@ private:
   /**
    * Read the parameters for the entry specified in the passed CategoryInfo
    */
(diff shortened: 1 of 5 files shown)


tmadlener:
The check of whether we switched TTrees inside the TChain had to be hoisted a level higher to make it possible to jump back again. The previous implementation implicitly relied on only moving forward. In that case checking that `localEntry == 0` was enough to detect TTree changes.

hegner (approved):

PR #418, opened 2023-05-19
Revive public `getMap` functionality for `GenericParameters`

tmadlener:
BEGINRELEASENOTES
- Bring back the public templated `getMap` functionality for `podio::GenericParameters` as they are already used in DD4hep (see AIDASoft/DD4hep#1112). 
  - Mark the existing `getXYZMap` as deprecated but keep them for a brief transition period.
  - These have been removed in #415.

ENDRELEASENOTES

### DIFF
--- a/include/podio/GenericParameters.h
+++ b/include/podio/GenericParameters.h
@@ -18,6 +18,9 @@ class write_device;
 using version_type = uint32_t; // from sio/definitions
 } // namespace sio
 
+#define DEPR_NON_TEMPLATE                                                                                              \
+  [[deprecated("Non-templated access will be removed. Switch to templated access functionality")]]
+
 namespace podio {
 
 /// The types which are supported in the GenericParameters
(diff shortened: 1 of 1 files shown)


include/podio/GenericParameters.h
```diff
@@ -162,6 +160,19 @@ class GenericParameters {
...
+  DEPR_NON_TEMPLATE const auto& getFloatMap() const {
+    return getMap<float>();
+  }
+
+  DEPR_NON_TEMPLATE const auto& getIntMap() const {
+    return getMap<int>();
+  }
+
+private:
   /// Get a reference to the internal map for a given type (necessary for SIO)
```
jmcarcell, line 176: Why is the non const version needed for SIO?

jmcarcell (approved):

jmcarcell:
I probably never checked for those in every repo as they were not originally part of #415 :sweat_smile:

include/podio/GenericParameters.h
tmadlener: It currently populates the maps in place.

PR #422, opened 2023-05-24
Make retrieval of parameters more robust in Frame python bindings

tmadlener:
BEGINRELEASENOTES
- Fix small bug in Frame python bindings where set but empty parameters could crash `podio-dump` when trying to access a non-existent element

ENDRELEASENOTES

### DIFF
--- a/python/podio/frame.py
+++ b/python/podio/frame.py
@@ -136,9 +136,9 @@ class Frame:
     """
     def _get_param_value(par_type, name):
       par_value = self._frame.getParameter[par_type](name)
-      if len(par_value) > 1:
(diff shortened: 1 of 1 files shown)


jmcarcell:
In which cases can you have set but empty parameters?

tmadlener:
That came up during the LCIO -> EDM4hep conversion. Not sure why, but there are cases where there are parameter keys without any non-empty value associated to it.

jmcarcell (approved):
Without being able to reproduce it looks good, the bug is quite obvious when pointed out. As long as the returned parameters are in a list (even an empty list) this will work fine

PR #424, opened 2023-06-04
wip: change std::vector to std::pmr::vector

stch-zhangyzh:
BEGINRELEASENOTES
- change std::vector to std::pmr::vector

ENDRELEASENOTES

### DIFF
--- a/include/podio/CollectionBuffers.h
+++ b/include/podio/CollectionBuffers.h
@@ -16,10 +16,10 @@ namespace podio {
 class CollectionBase;
 
 template <typename T>
-using UVecPtr = std::unique_ptr<std::vector<T>>;
+using UVecPtr = std::unique_ptr<std::pmr::vector<T>>;
 
-using CollRefCollection = std::vector<UVecPtr<podio::ObjectID>>;
-using VectorMembersInfo = std::vector<std::pair<std::string, void*>>;
+using CollRefCollection = std::pmr::vector<UVecPtr<podio::ObjectID>>;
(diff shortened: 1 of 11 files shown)


tmadlener:
I like the general idea of this, and I have also thought about doing this in the past, as it would potentially allow some quite cool features. May I ask what your motivation for changing to `pmr` vectors is?

Overall doing this *right* comes with quite a few considerations:

- If we switch to `pmr` vectors we should also think about which parts we expose to (framework) users. Simply switching to `pmr` doesn't really change anything. Having the possibility to pass in a custom memory resource would potentially allow to use podio more easily within a framework that keeps ownerships of the objects itself (rather than having podio do that).
  - This comes with the question of what the defaults should be.
- How does `pmr` play with our I/O, specifically ROOT. Is it supported directly or do we have to do some work for this?
- Can we use this to make things easily work on heterogeneous resources, e.g. by using something like [vecmem](https://github.com/acts-project/vecmem/tree/main).
  - Potentially unrelated, but could play a role, can we more easily support SoA vs AoS with this?
- Compiler support for polymorphic memory resources has only landed in LLVM 16 ([see here](https://en.cppreference.com/w/cpp/compiler_support/17#C.2B.2B17_library_features)) so anyone would have to switch to that.

From a prioritization point of view, I would really like to finish podio v1.0 first and then see how to best go forward with this. With all that in mind, I still think this is at least something that should be tried to see how far this can be taken and how it works in the end, and I would be happy to help with that.

PR #425, opened 2023-06-07
Make Catch2 setup more robust against podios in environment

tmadlener:
BEGINRELEASENOTES
- Add a `SKIP_CATCH_DISCOVERY` cmake option to skip the unittest discovery of Catch2 to avoid running the catch discovery in an unsuitable environment.
- Make environment for unittests more specific to avoid catching too much of the underlying environment.

ENDRELEASENOTES

Currently this can be quite annoying if there is another podio installation in the environment because the test discovery is then run in an unsuitable environment which can lead to errors during dynamic library loading.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -251,7 +251,7 @@ if (USE_SANITIZER MATCHES "Memory(WithOrigin)?")
     add_test(NAME unittest COMMAND unittest ${filter_tests})
     set_property(TEST unittest
       PROPERTY ENVIRONMENT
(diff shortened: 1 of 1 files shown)


hegner:
Thanks @tmadlener . Looks fine to me. Should I merge after tests finish/succeed?

tmadlener:
Sounds good to me. The key4hep nightlies are broken at the moment, `python` is not picking up the key4hep python . I have adapted the python unittest to use `python3` which should work.

(but now github seems to have degraded service, so it might take a while until this appears here)

PR #428, opened 2023-06-14
Reorganize the CMake configuration for the tests

tmadlener:
BEGINRELEASENOTES
- Split the tests directory into several (more or less) topical sub-directories to declutter the main test `CMakeLists.txt` a bit
- Move commonly used functionality into `cmake/podioTests.cmake` (e.g. setting up a test environment)
- Move python unittests config to the `python` directory

ENDRELEASENOTES

This is really (almost entirely) just a reshuffling of existing CMake configuration, mainly to declutter the `tests/CMakeLists.txt` that has grown to  > 300 lines and which was becoming hard to navigate. I put this cleanup into a separate PR, to have a better commit history in the end as this is also preparatory work for the testing of schema evolution functionality.


hegner (approved):
If tests run through, we can merge. Thanks for the cleanup

Issue #436, opened 2023-06-29
Misleading Error: Invalid Member Definition


Ananya2003Gupta:
- OS version: Ubuntu 22.04.2
 
 - Reproduced by:
 `python ../python/podio_class_generator.py ../tests/datalayout.yaml ../Tmp data ROOT`
 
- Input:
`datalayout.yaml`
```
options:
  getSyntax: False
  exposePODMembers: True

components:
  # Component representing position
  PositionComponent:
    Members:
      - float x
      - float y
      - float z

datatypes:
  # Data type for calorimeter hits
  CalorimeterHit:
    Description: "Calorimeter Hits"
    Author: "Me"
    Members:
     - float energy
     - PositionComponent position
    OneToOneRelations:
     - Particle particle // One-to-one relation with Particle

  # Data type for particles
  Particle:
    Description: "Particle"
    Author: "Me"
    Members:
     - int pdgId
     - float energy
     - PositionComponent momentum
    OneToOneRelations:
     - CalorimeterHit calorimeterHit // One-to-one relation with CalorimeterHit
     
```
- Error Message:
` Error while generating the datamodel: 'float energy' is not a valid member definition `
 
- Issue: 
The error message is misleading and does not accurately describe the root cause of the problem.

- Analysis: 
After thorough debugging and documentation review, it was discovered that the error was due to the absence of comments after each member definition.

- Solution: 
Adding comments to the member definitions resolved the error.

**Correction:**
```
datatypes:
  # Data type for calorimeter hits
  CalorimeterHit:
    Description: "Calorimeter Hits"
    Author: "Me"
    Members:
     - float energy //energy
     - PositionComponent position //position
    OneToOneRelations:
     - Particle particle // One-to-one relation with Particle

  # Data type for particles
  Particle:
    Description: "Particle"
    Author: "Me"
    Members:
     - int pdgId //pdgId
     - float energy //energy
     - PositionComponent momentum //momentum
    OneToOneRelations:
     - CalorimeterHit calorimeterHit // One-to-one relation with CalorimeterHit
     
```
- Expected Behavior:
The error message should accurately reflect the issue, indicating that comments are required for each member definition.

- Actual Behavior:
The error message is misleading and does not provide clear guidance on the root cause, leading to confusion during debugging.

- Possible Solution:
Updating the error message to include information about the requirement of comments in each member definition to prevent confusion and provide more accurate troubleshooting guidance.

tmadlener:
[...]

PR #437, opened 2023-06-29
Solves Issue #436 : Misleading Error: Invalid Member Definition

Ananya2003Gupta:
BEGINRELEASENOTES
Modified parse function definition to incorporate `missing description in member definition` error message.

ENDRELEASENOTES
Solves Issue #436 : Misleading Error: Invalid Member Definition


Ananya2003Gupta:
> Thanks for taking care of this. I triggered continuous integration (CI), and I have some minor suggestions below.

I didn't do anything. The additions were suggested by you only : )

python/podio/podio_config_reader.py
```diff
@@ -57,7 +57,7 @@ def _parse_with_regexps(string, regexps_callbacks):
       if result:
         return callback(result)
 
-    raise DefinitionError(f"'{string}' is not a valid member definition")
+    raise DefinitionError(f"'{string}' is not a valid member definition. Check syntax of the member definition.")
```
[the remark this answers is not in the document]
Ananya2003Gupta: Thank you : )

tmadlener:
It looks like our format and lint checks are not entirely happy yet:

https://github.com/AIDASoft/podio/actions/runs/5415352399/jobs/9843528142?pr=437#step:4:499

The rest looks fine. There is one workflow that is known to fail and which you can ignore in this case (`key4hep`).

Ananya2003Gupta:
> It looks like our format and lint checks are not entirely happy yet:
> 
> https://github.com/AIDASoft/podio/actions/runs/5415352399/jobs/9843528142?pr=437#step:4:499
> 
> The rest looks fine. There is one workflow that is known to fail and which you can ignore in this case (`key4hep`).

Following errors are generated.
```
python/podio/podio_config_reader.py:89:13: E126 continuation line over-indented for hanging indent
  python/podio/podio_config_reader.py:91:9: E121 continuation line under-indented for hanging indent
  python/podio/podio_config_reader.py:94:21: E126 continuation line over-indented for hanging indent
  python/podio/podio_config_reader.py:96:7: E121 continuation line under-indented for hanging indent
  python/podio/podio_config_reader.py:103:121: E501 line too long (261 > 120 characters)
  python/podio/podio_config_reader.py:103:182: E262 inline comment should start with '# '
```

tmadlener (commented):
You will have to fix the format issues to make that check happy. I have marked most of the occurrences below. If you have `flake8` and `pre-commit` installed locally you can [run pre-commit locally](https://github.com/AIDASoft/podio/blob/master/doc/advanced_topics.md#running-pre-commit). You can also just do `pre-commit run -a flake8` to just run the flake8 checks (which is much quicker than the full suite).

python/podio/podio_config_reader.py
```diff
@@ -85,18 +85,25 @@ def _bare_member_conv(result):
...
+    no_desc_matchers_cbs = [
+        (self.bare_array_re, self._bare_array_conv),
+        (self.bare_member_re, self._bare_member_conv)]
+
+    if require_description:
+      try:
+        return self._parse_with_regexps(string, default_matchers_cbs)
+      except DefinitionError:
+        """check whether we could parse this if we don't require a description and
+        provide more details in the error if we can"""
```
tmadlener, line 101:
```suggestion
        # check whether we could parse this if we don't require a description and
        # provide more details in the error if we can
```
Otherwise pylint complains about a string statement without effect: [log](https://github.com/AIDASoft/podio/actions/runs/5416170153/jobs/9845547895?pr=437#step:4:393)

python/podio/podio_config_reader.py
```diff
@@ -85,18 +85,25 @@ def _bare_member_conv(result):
...
+        self._parse_with_regexps(string, no_desc_matchers_cbs)
```
tmadlener, line 102:
```suggestion
        self._parse_with_regexps(string, no_desc_matchers_cbs)
        # pylint: disable-next=raise-missing-from
```
To silence the pylint warning which is more of a false positive in this case

python/podio/podio_config_reader.py
```diff
@@ -85,18 +85,25 @@ def _bare_member_conv(result):
...
+        raise DefinitionError(f"""'{string}' is not a valid member definition. Description comment is missing.
+Correct Syntax: <type> <name> // <comment>""")
```
tmadlener, line 104:
```suggestion
        raise DefinitionError(f"'{string}' is not a valid member definition. Description comment is missing."
                                                " Correct Syntax: <type> <name> // <comment>")
```
You can also split strings this way and they will be concatenated. I find this a bit more readable.

tmadlener (commented):
Thanks for fixing the formatting. I just realized that there are also some `pylint` warnings, which should be fixed by the comments below.

Ananya2003Gupta:
Doing the above changes. 
Really sorry. I just ran the flake8 tests and committed the changes.

Ananya2003Gupta:
Also, should I put `Correct Syntax: <type> <name> // <comment>` in `raise DefinitionError(f"'{string}' is not a valid member definition. Check syntax of the member definition.")`

python/podio/podio_config_reader.py
```diff
@@ -85,18 +85,26 @@ def _bare_member_conv(result):
...
+    if require_description:
+      try:
+        return self._parse_with_regexps(string, default_matchers_cbs)
+      except DefinitionError:
+        # check whether we could parse this if we don't require a description and
+        # provide more details in the error if we can
+        self._parse_with_regexps(string, no_desc_matchers_cbs)
+        # pylint: disable-next=raise-missing-from
+        raise DefinitionError(f"'{string}' is not a valid member definition. Description comment is missing.\n"
+                               + "Correct Syntax: <type> <name> // <comment>")
```
tmadlener, line 105:
```suggestion
        raise DefinitionError(f"'{string}' is not a valid member definition. Description comment is missing."
                               " Correct Syntax: <type> <name> // <comment>")
```
You don't need the `+` in this case, and I would also not put in a newline (`\n`) to the output. Other than that this looks good.

Ananya2003Gupta: I put the newline character so that the error message looks good.
With `newline character`
image
Without `newline character`
image
Without `newline character` the error doesn't look good on screen.

tmadlener: OK. I like the newline. But the `+` is not necessary for string concatenation in this case, because it will be done automatically inside the `()`.

Ananya2003Gupta: Okay, I'll keep the newline character and remove + .

Ananya2003Gupta:
> Just waiting for the last few workflows to finish, but they should be fine. Thanks again for all the work (and the patience in addressing my comments).
 
It was really fun working on this issue and I got to learn a lot about pre-commit through it.

> If I may ask: How did you get into contact with podio?
 
I am selected as HSF India Fellow and will be working on interfacing PODIO with Julia backend.

tmadlener:
Ah very nice. We will be hearing more from you then :)

Ananya2003Gupta:
> Ah very nice. We will be hearing more from you then :)

Yes : )

PR #441, opened 2023-07-10
[WIP] Schema evolution code generation

hegner:
BEGINRELEASENOTES
- ...

ENDRELEASENOTES

Based on #430 
- [ ] Needs #460


tmadlener:
> Based on https://github.com/AIDASoft/podio/pull/430

Just for my clarification from my side: This means we still keep 430 as it is and merge it separately, or we close that and keep this one for the feature?

hegner:
Since I am fixing bugs in #430, I think we better close it. I previously thought functionality in there was stable.

tmadlener:
Rebased this onto #460. For me locally there are a few test issues with the datamodel storage roundtrip tests, because they now trip over the fact that for the schema evolution tests we generate "old" `Data.h` headers into the same directory. This makes the final `diff` step in those tests fail.

I don't have a real fix for that yet, but I would simply wait for now to see how the final solution to schema evolution looks like and then fix those tests afterwards.

hegner:
Since based on #430 , this development will not be integrated.

PR #445, opened 2023-07-12
Allow to specify units

hegner:
BEGINRELEASENOTES
- Allow to specify units as part of the datamodel definition 

ENDRELEASENOTES

This WIP is to allow the explicit inclusion of units into the datamodel definition. 

- [x] Extend the regular expressions/parsing to allow a unit description
- [x] Documentation

### DIFF
--- a/python/podio/test_MemberParser.py
+++ b/python/podio/test_MemberParser.py
@@ -243,6 +243,13 @@ class MemberParserTest(unittest.TestCase):
     self.assertEqual(parsed.description, 'descriptions are not ignored even though they are not required')
     self.assertTrue(not parsed.is_builtin)
 
+  def test_parse_unit(self):
+    """Test that units are properly parsed"""
+    parser = MemberParser()
+
+    parsed = parser.parse('unsigned long long var [GeV] // description')
+    self.assertEqual(parsed.unit, 'GeV')
+
   def test_string_representation(self):
     """Test that the string representation that is used in the jinja2 templates
     includes the default initialization"""
--- a/python/podio/test_MemberParser.py
+++ b/python/podio/test_MemberParser.py
@@ -250,6 +250,11 @@ class MemberParserTest(unittest.TestCase):
     parsed = parser.parse('unsigned long long var [GeV] // description')
     self.assertEqual(parsed.unit, 'GeV')
 
+    parsed = parser.parse('unsigned long long var{42} [GeV] // description')
+    self.assertEqual(parsed.unit, 'GeV')
+    self.assertEqual(parsed.default_val, '42')
+
+
   def test_string_representation(self):
     """Test that the string representation that is used in the jinja2 templates
     includes the default initialization"""
--- a/python/podio/test_MemberParser.py
+++ b/python/podio/test_MemberParser.py
@@ -254,7 +254,6 @@ class MemberParserTest(unittest.TestCase):
     self.assertEqual(parsed.unit, 'GeV')
     self.assertEqual(parsed.default_val, '42')
 
-
   def test_string_representation(self):
     """Test that the string representation that is used in the jinja2 templates
     includes the default initialization"""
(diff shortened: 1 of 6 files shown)


hegner:
@tmadlener - can you let me know whether you'd agree with this? I needed some break from genreflex work and did this long-standing item on my wish list.

python/podio/test_MemberParser.py
```diff
@@ -243,6 +243,13 @@ def test_parse_valid_no_description(self):
     self.assertEqual(parsed.description, 'descriptions are not ignored even though they are not required')
     self.assertTrue(not parsed.is_builtin)
 
+  def test_parse_unit(self):
+    """Test that units are properly parsed"""
+    parser = MemberParser()
+
+    parsed = parser.parse('unsigned long long var [GeV] // description')
+    self.assertEqual(parsed.unit, 'GeV')
+
```
tmadlener, line 252: Could we add another test case here with an additional default value?

tmadlener (commented):
I think this is a good idea. It is not yet clear to me if/how this will affect the generated code. That might have quite an impact on the size of this feature.

Would it mainly be an addition in the docstring, or would it actually enter in the generated c++ code. If the latter, which base units do we use, or do we leave that choice to the users (and if so, how)? Would we try to make this rather strict and introduce a proper units library to do all the conversions for us (there is a propoal for c++29 to have this in the standard library)?

python/podio/test_MemberParser.py
hegner: what would be a default value?

hegner:
For the moment it is for documentation purposes. And I want to have a detection in the schema evolution in case it changes. I do not think we should use a units library unless being asked for explicitly.

python/podio/test_MemberParser.py
tmadlener: Yes, exactly. Having a default value and a unit. Just to cover all possibilities.

tmadlener:
OK. That sounds reasonable to achieve then. Then the main comment/question I have is whether we also need this for the components?

hegner:
Obviously :-)

tmadlener:
Does this already work for components as well?

hegner:
Yes it does. The test case covers `NotSoSimpleStruct` (component) and `ExampleHit` (datatype).

tmadlener:
Ah right, sorry I missed that because it also has a description. Then more specifically does this also work with components that do not have a description, e.g. the `SimpleStruct`?

hegner:
Sure. The description is default initialized as empty string (`""`), thus there is no special case to handle

tmadlener:
I was wondering more about the yaml parsing side here. IIRC I special cased the components there to not require a description, and I am not sure if this already covers that special case

hegner:
The combined regex would give a `None` for that part and we are back to default handling later. That's the nice thing of having unified the syntax of components and datatypes.

tmadlener:
OK. IIUC, everything should work from the parsing side and nothing breaks if I add units, e.g. here:
https://github.com/AIDASoft/podio/blob/dbf9425aa6d2b956c4c6fa49d45abe40cbe9e5af/tests/datalayout.yaml#L13-L17

However, I think in the current implementation these units would also not end up in a docstring, because they are never passed to the `MemberVariable` in this case, right? (The `bare_<xyz>` regexes do not have the `unit_str` yet, and so the unit is also not passed to the `MemberVariable` constructor in the end).

hegner:
@tmadlener - I addressed your comment with a fix. Indeed units were not handled correctly there.

tmadlener:
It looks like the `clang12` based workflows are failing due to not having root in their environment. I suspect that these are transient issues in the LCG nightlies and they should be safe to ignore. The `key4hep` (release) based one is failing due to an unsuitable ROOT version, so failure there is also expected.

andresailer:
The clang12 nightlies no longer exist

tmadlener (approved):

tmadlener:
@hegner could you rebase this onto the latest master (or merge that in), so that we pick up a completely working CI again. After that finishes successfully, I would merge this.

hegner:
@tmadlener - seems all green now :-)

tmadlener (approved):

PR #446, opened 2023-07-13
Make the SIOFrameWriter call to finish non-mandatory

tmadlener:
BEGINRELEASENOTES
- Make calling `finish` for the `SIOFrameWriter` non-mandatory. See #442 for other related changes.

ENDRELEASENOTES

PR #447, opened 2023-07-13
Add python bindings for writing Frames

### LINKED ISSUE #432: Improve python interface when adding collections to a frame
At the moment writing collections and frames is possible from python but a bit cumbersome:
```python
from edm4hep import edm4hep
from ROOT import podio
from podio.root_io import Writer
import cppyy

writer = Writer('frame.root') # This will write the frame to a file called frame.root

frame = podio.Frame() # This is where the collection and other things will be stored
coll = edm4hep.MCParticleCollection() # Create an EDM4hep collection

coll.create() # Create a MutableMCParticle object

frame.put(cppyy.gbl.std.move(coll), "MCParticles") # Add collection to frame 

writer.writeFrame(frame, 'events')
writer.finish()

```
I think the agreement here is that `frame.put` should work without `cppyy.gbl.std.move` (and also with it in case someone is translating literally the C++ code?). I can give it a try this week, it should be easy to do

tmadlener:
BEGINRELEASENOTES
- Add a python wrapper around the different available Frame writers.
- Add a `put` method to the `Frame` wrapper that allows to add collections to the Frame without having to explicitly use `cppyy.gbl.std.move`. Fixes #432 
- Add test cases that write via python bindings and read via c++.

ENDRELEASENOTES

- [x] Needs #446 
- [x] Also add `put_parameters` (+ tests)
- [x] rebase this onto #449
- [x] Needs #457

### DIFF
--- a/python/podio/frame.py
+++ b/python/podio/frame.py
@@ -236,7 +236,6 @@ class Frame:
 
     self._param_key_types = self._get_param_keys_types()  # refresh the cache
 
-
   def get_parameters(self):
     """Get the complete podio::GenericParameters object stored in this Frame.
 
--- a/python/podio/frame.py
+++ b/python/podio/frame.py
@@ -94,7 +94,7 @@ class Frame:
     else:
       self._frame = podio.Frame()
 
-    self._param_key_types = self._init_param_keys()
+    self._param_key_types = self._get_param_keys_types()
 
   @property
   def collections(self):
@@ -196,6 +196,47 @@ class Frame:
 
     return _get_param_value(vec_types[0], name)
 
+  def put_parameter(self, key, value):
+    """Put a parameter into the Frame.
+
+    Puts a parameter into the Frame after doing some (incomplete) type checks.
+    If a list is passed the parameter type is determined from looking at the
+    first element of the list only. Additionally, since python doesn't
+    differentiate between floats and doubles, floats will always be stored as
+    double.
+
+    Args:
+        key (str): The name of the parameter
+        value (int, float, str or list of these): The parameter value
+
+    Raises:
+        ValueError: If a non-supported parameter type is passed
+    """
+    # For lists we determine the c++ vector type and use that to call the
+    # correct template overload explicitly
+    if isinstance(value, (list, tuple)):
+      vec_types = _get_cpp_vector_types(type(value[0]).__name__)
+      if len(vec_types) == 0:
+        raise ValueError(f"Cannot put a parameter of type {type(value[0])} into a Frame")
+
+      par_type = vec_types[0]
+      if isinstance(value[0], float):
+        # Always store floats as doubles from the python side
+        par_type = par_type.replace("float", "double")
+
+      self._frame.putParameter[par_type](key, value)
+    else:
+      # If we have a single integer, a std::string overload kicks in with higher
+      # priority than the template for some reason. So we explicitly select the
+      # correct template here
+      if isinstance(value, int):
+        self._frame.putParameter["int"](key, value)
+      else:
+        self._frame.putParameter(key, value)
+
+    self._param_key_types = self._get_param_keys_types()  # refresh the cache
+
+
   def get_parameters(self):
     """Get the complete podio::GenericParameters object stored in this Frame.
 
@@ -233,7 +274,7 @@ class Frame:
 
     return par_infos
 
-  def _init_param_keys(self):
+  def _get_param_keys_types(self):
     """Initialize the param keys dict for easier lookup of the available parameters.
 
     Returns:
--- a/python/podio/frame.py
+++ b/python/podio/frame.py
@@ -82,7 +82,6 @@ class Frame:
   # with the correct type that we can compare against
   _coll_nullptr = cppyy.bind_object(cppyy.nullptr, 'podio::CollectionBase')
 
-
   def __init__(self, data=None):
     """Create a Frame.
 
--- a/python/podio/frame.py
+++ b/python/podio/frame.py
@@ -131,13 +131,18 @@ class Frame:
   def put(self, collection, name):
     """Put the collection into the frame
 
+    The passed collectoin is "moved" into the Frame, i.e. it cannot be used any
+    longer after a call to this function. This also means that only objects that
+    were in the collection at the time of calling this function will be
+    available afterwards.
+
     Args:
         collection (podio.CollectionBase): The collection to put into the Frame
         name (str): The name of the collection
 
     Returns:
         podio.CollectionBase: The reference to the collection that has been put
-           into the Frame
+            into the Frame. NOTE: That mutating this collection is not allowed.
 
     Raises:
         ValueError: If collection is not actually a podio.CollectionBase
--- a/python/podio/frame.py
+++ b/python/podio/frame.py
@@ -48,13 +48,19 @@ def _determine_supported_parameter_types():
 SUPPORTED_PARAMETER_TYPES = _determine_supported_parameter_types()
 
 
-def _get_cpp_vector_types(type_str):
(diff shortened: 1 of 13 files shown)


python/podio/root_io.py
```diff
@@ -2,12 +2,13 @@
...
 
 from podio.base_reader import BaseReaderMixin
+from podio.base_writer import BaseWriterMixin
 
 from ROOT import gSystem
 gSystem.Load('libpodioRootIO')  # noqa: E402
 from ROOT import podio  # noqa: E402 # pylint: disable=wrong-import-position
 
-Writer = podio.ROOTFrameWriter
+# Writer = podio.ROOTFrameWriter
```
tmadlener, line 11:
```suggestion
```

python/podio/frame.py
```diff
@@ -66,6 +82,7 @@ class Frame:
   # with the correct type that we can compare against
   _coll_nullptr = cppyy.bind_object(cppyy.nullptr, 'podio::CollectionBase')
 
+
```
tmadlener, line 85:
```suggestion
```

tmadlener:
This became quite a big bigger than I initially anticipated, since there seem to be some quirks around how cppyy selects overloads for the parameters part. Given the improvements in the overall "pythonness" that we now have in EDM4hep, I am no longer sure the user visible import structure we have here is the best, but I would put that into another PR.

hegner:
@tmadlener - so this PR is complete as is?

tmadlener:
Technically yes. I have to check why it is failing CI. Also assuming that we address a nicer "import interface" later.

python/podio/frame.py
```diff
@@ -163,6 +196,46 @@ def _get_param_value(par_type, name):
 
     return _get_param_value(vec_types[0], name)
 
+  def put_parameter(self, key, value):
```
jmcarcell, line 199: Since the overload for floats will always be double, is it worth it to add a `float=True` that will select the float overload? Otherwise it's impossible to add a float parameter from python, in the rare case someone needs float and not double...

python/podio/frame.py
```diff
@@ -107,9 +122,27 @@ def get(self, name):
...
+    Returns:
+        podio.CollectionBase: The reference to the collection that has been put
+           into the Frame
+
+    Raises:
+        ValueError: If collection is not actually a podio.CollectionBase
+    """
+    if not _is_collection_base(collection):
+      raise ValueError("Can only put podio collections into a Frame")
+    return self._frame.put(cppyy.gbl.std.move(collection), name)
```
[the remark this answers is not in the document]
jmcarcell: I just tried it and indeed it crashes, for this PR is fine but there should be something to prevent those...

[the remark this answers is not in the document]
tmadlener: I think it depends on how you want to use the collection. Indexing into it will crash, because it is effectively empty. Adding new things to it might work, but I haven't tried it.

tests/sio_io/CMakeLists.txt
```diff
@@ -30,3 +32,9 @@ if (TARGET read_sio)
   add_test(NAME check_benchmark_outputs_sio COMMAND check_benchmark_outputs write_benchmark_sio.root read_benchmark_sio.root)
   set_property(TEST check_benchmark_outputs_sio PROPERTY DEPENDS read_timed_sio write_timed_sio)
 endif()
+
+#--- Write via python and the SIO backend and see if we can read it back in in
+#--- c++
+add_test(NAME write_python_frame_sio COMMAND python3 ${CMAKE_CURRENT_LIST_DIR}/write_frame_sio.py)
+PODIO_SET_TEST_ENV(write_python_frame_sio)
+set_property(TEST read_python_frame_sio PROPERTY DEPENDS write_python_frame_sio)
```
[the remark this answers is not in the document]
tmadlener: Pulled this one out into: https://github.com/AIDASoft/podio/pull/457

python/podio/frame.py
tmadlener: Possible now via an additional `as_type` argument.

python/podio/frame.py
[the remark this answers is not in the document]
tmadlener: [updated docstring](https://github.com/AIDASoft/podio/pull/447/files#diff-eaff5706c27751eee9136bf2594ec21b3112a331d85ea76b5c7afc69de298a9dR134-R137)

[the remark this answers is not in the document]
jmcarcell: I feel like here going out of range may be more of a problem than in C++ (this is not related to this PR but just indexing out of range a `std::deque`):
``` python
event = Frame()
hits = ExampleHitCollection()
hits[0] # seg fault
hits[1] # <cppyy.gbl.MutableExampleHit object at 0x55dcd5d30d50>
hits[1].x() # seg fault
hits.create(0, 1, 2. 3, 4)
hits[0] # <cppyy.gbl.MutableExampleHit object at 0x55dcc4324040>
hits[1] # <cppyy.gbl.MutableExampleHit object at 0x55dcd5d572b0>
hits[2] # seg fault
```
 It's inconsistent since the behavior is undefined, could be a bit hard to debug when you go out of range and it may seem to work but then it doesn't later on. I'm not sure if there are options, not checking in C++ I think is a design choice since there is also `.at`, but in python no one is going to use that probably

jmcarcell (approved):
Looks good. The interface can be improved like in EDM4hep but that can be done later.

python/podio/frame.py
```diff
@@ -163,6 +207,58 @@ def _get_param_value(par_type, name):
...
+        raise ValueError(f"Cannot put a parameter of type {type_name} into a Frame")
+
+      par_type = vec_types[0]
+      if isinstance(value[0], float):
+        # Always store floats as doubles from the python side
+        par_type = par_type.replace("float", "double")
+
+      self._frame.putParameter[par_type](key, value)
+    else:
+      if as_type is not None:
```
jmcarcell, line 246:
```suggestion
      if as_type:
```

python/podio/frame.py
```diff
@@ -48,16 +48,38 @@ def _determine_cpp_type(idx_and_type):
 SUPPORTED_PARAMETER_TYPES = _determine_supported_parameter_types()
 
 
-def _get_cpp_vector_types(type_str):
-  """Get the possible std::vector<cpp_type> from the passed py_type string."""
-  # Gather a list of all types that match the type_str (c++ or python)
+def _get_cpp_types(type_str):
+  """Get all possible c++ types from the passed py_type string."""
   types = list(filter(lambda t: type_str in t, SUPPORTED_PARAMETER_TYPES))
```
jmcarcell, line 53:
```suggestion
  types = [t for t in SUPPORTED_PARAMETER_TYPES if type_str in t]
```
simpler, more pythonic

python/podio/frame.py
```diff
@@ -48,16 +48,38 @@ def _determine_cpp_type(idx_and_type):
...
     raise ValueError(f'{type_str} cannot be mapped to a valid parameter type')
 
+  return types
+
+
+def _get_cpp_vector_types(type_str):
+  """Get the possible std::vector<cpp_type> from the passed py_type string."""
+  # Gather a list of all types that match the type_str (c++ or python)
+  types = _get_cpp_types(type_str)
   return [f'std::vector<{t}>' for t in map(lambda x: x[0], types)]
```
jmcarcell, line 64:
```suggestion
  return [f'std::vector<{t[0]}>' for t in types]
```

python/podio/frame.py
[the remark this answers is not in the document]
tmadlener: Yeah in the current form there are not that many handrails in python. Could probably be fixed by another thin python shim that introduces the range check. That could then also be used to raise an error in the *use-after-move* case.

Created #447 to keep track of this.

PR #449, opened 2023-07-13
Fix the pre-commit workflow

tmadlener:
BEGINRELEASENOTES
- Fix the `pre-commit` workflow by making it run on top of the key4hep nightlies that come with a recent enough root version to be able to work with the RNTuple addition (#395)
- Fix a few minor complaints from newer versions of `clang-format` and `clang-tidy`

ENDRELEASENOTES

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -15,14 +15,17 @@ jobs:
     - uses: cvmfs-contrib/github-action-cvmfs@v3
     - uses: aidasoft/run-lcg-view@v4
       with:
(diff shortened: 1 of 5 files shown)


hegner:
Great. Waiting to see what the test will say :-)

hegner:
@tmadlener - I assume it only takes effect *after* merging :-(

tmadlener:
No, it should take into account the state of this branch. I am not entirely sure why `pylint` starts complaining now, could be a minor version mismatch between before and now. I am having a look at this

tmadlener:
OK. Now everything should work. Let's see if CI agrees.

hegner:
thanks!

Issue #371, opened 2023-02-07
Components should have a Description and Author field


tmadlener:
Currently components cannot have a `Description` and/or an `Author` field. For datatypes these are required to be present. I think it would make sense to have the possibility to also add a brief, optional description and author to components.

Otherwise these will only appear as commens in the yaml definition, but not in the generated documentation, e.g.https://github.com/key4hep/EDM4hep/blob/67ee6735f6aadf3f05d2b310b4b3c92f89d7138e/edm4hep.yaml#L68-L69

PR #450, opened 2023-07-13
add description and author to components

hegner:
BEGINRELEASENOTES
- Add optional description and author fields to component definition
ENDRELEASENOTES

addressing #371

### DIFF
--- a/python/podio/podio_config_reader.py
+++ b/python/podio/podio_config_reader.py
@@ -162,7 +162,7 @@ class ClassDefinitionValidator:
     """Check the components."""
     for name, component in datamodel.components.items():
       for field in component:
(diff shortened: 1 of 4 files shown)


tmadlener (commented):
I would have thought this to be more invasive, but this looks good to me. How does the generated code look like if `Author` and `Description` are not set?  
https://github.com/AIDASoft/podio/blob/c13ada895df12044ec6a07fcc8410142b2b42928/python/templates/macros/declarations.jinja2#L1-L6
Can doxygen work with an unset `@author` field? The rest should be fine as far as I can see

hegner:
my doxygen is happy with it

tmadlener (approved):

PR #452, opened 2023-07-13
Hide the JSON output functionality from CLING

### LINKED ISSUE #435: `PODIO_JSON_OUTPUT` not set downstream, but json support still compiled into libraries
> *Note: this is a lazy copy from mattermost*

This seems to be caused as follows:
- during the build of EDM4hep, `PODIO_JSON_OUTPUT` is defined and libedm4hepDict.so includes the `to_json` function ([Component.h.jinja2](https://github.com/AIDASoft/podio/blob/master/python/templates/Component.h.jinja2#L43) is `inline`, but others aren't e.g. [Collection.cc.jinja2](https://github.com/AIDASoft/podio/blob/master/python/templates/Collection.cc.jinja2#L231)).
- during compiling with ROOT cling, the `#define` isn't set so it gets confused.

Here is a reproducer in our production environment which also works on my local spack with dd4hep, podio, edm4hep installed:
```console
ddsim --compactFile $DD4hepINSTALL/DDDetectors/compact/SiD.xml -G -N10 --outputFile SiD.edm4hep.root
root -l -q 'podio_issue.cxx+g("SiD.edm4hep.root")'
```
where `podio_issue.cxx` is
```c++
#include "ROOT/RDataFrame.hxx"
#include "TCanvas.h"

#include "edm4hep/MCParticleCollection.h"

int podio_issue(const char* fname)
{
  ROOT::RDataFrame df("events", fname);

  auto df0 = df.Define("n", "MCParticles.size()");
  auto h_n = df0.Histo1D({"h_n", "; N ", 10, 0, 10}, "n");

  auto c = new TCanvas();
  h_n->DrawCopy();
  return 0;
}
```

That returns
```console

Processing podio_issue.cxx+g("SiD.edm4hep.root")...
libedm4hepDict dictionary payload:267:7: error: no matching constructor for initialization of 'nlohmann::json' (aka 'basic_json<>')
  j = nlohmann::json{{"location", value.location},
      ^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/opt/local/include/nlohmann/json.hpp:892:5: note: candidate constructor not viable: requires at most 3 arguments, but 9 were provided
    basic_json(initializer_list_t init,
    ^
/opt/local/include/nlohmann/json.hpp:1007:5: note: candidate constructor not viable: requires 2 arguments, but 9 were provided
(etc)
```

tmadlener:
BEGINRELEASENOTES
- Extend the pre-processor condition for the `to_json` functionality to not be visible in `rootcling` or the root interpreter. Fixes #435 

ENDRELEASENOTES

@wdconinc thanks for the hint with `__CLING__`, wouldn't have thought of that. Could you check if this is indeed enough to fix the issue. The example in #435 works for me, but that might not be the end of the story.

### DIFF
--- a/python/templates/Collection.cc.jinja2
+++ b/python/templates/Collection.cc.jinja2
@@ -227,7 +227,7 @@ const auto registeredCollection = registerCollection();
 } // namespace
 
 
-#ifdef PODIO_JSON_OUTPUT
+#if defined(PODIO_JSON_OUTPUT) && !defined(__CLING__)
 void to_json(nlohmann::json& j, const {{ collection_type }}& collection) {
   j = nlohmann::json::array();
   for (auto&& elem : collection) {
--- a/python/templates/Collection.h.jinja2
+++ b/python/templates/Collection.h.jinja2
@@ -18,7 +18,7 @@
 #include "podio/CollectionBase.h"
 #include "podio/CollectionIDTable.h"
 
-#ifdef PODIO_JSON_OUTPUT
+#if defined(PODIO_JSON_OUTPUT) && !defined(__CLING__)
 #include "nlohmann/json.hpp"
 #endif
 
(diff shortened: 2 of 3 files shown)


wdconinc:
Confirming this works fine with EDM4hep 0.7.2 in our stack, no more to_json failures in running ROOT macros.

tmadlener:
Thanks for the quick testing. I am also backporting this to the v00-16-06 patch release.

PR #457, opened 2023-07-20
Make building SIO tests depend on ENABLE_SIO

tmadlener:
BEGINRELEASENOTES
- Simplify the test setup for SIO in CMake and make it explicit on the `ENABLE_SIO` option rather than on the presence of targets.

ENDRELEASENOTES


jmcarcell:
Looks good

jmcarcell (approved):

Issue #431, opened 2023-06-16
Mutability conversion fails on `push_back` to collection via Python bindings


bartosik-hep:
Python bindings have been recently added to `edm4hep`, which don't work as expected in what regards the implicit conversion of mutable objects to immutable ones. In particular this happens in the following case:

```python
from edm4hep import edm4hep

collection = edm4hep.MCParticleCollection()
particle = edm4hep.MutableMCParticle()
particle.setPDG(22)
collection.push_back(particle)
```
producing the following error:
```c++
TypeError: void edm4hep::MCParticleCollection::push_back(edm4hep::MCParticle object) =>
    TypeError: could not convert argument 1
```

This error doesn't happen when an object is created directly in the collection:

```python
from edm4hep import edm4hep

collection = edm4hep.MCParticleCollection()
particle = collection.create()
particle.setPDG(22)
```

Setup description:
```
 - OS version: Alma Linux 9.2
 - Compiler version: GCC 11.3.1
 - PODIO version: 0.16.6
```

tmadlener:
The actual issue is most likely in cppyy, but we should be able to work around this from our side, depending on how long it takes to fix this upstream

PR #460, opened 2023-07-21
[WIP] Moving PODIO to using rootcling instead of reflex

hegner:
BEGINRELEASENOTES
- Migrating to rootcling (replacing reflex)
ENDRELEASENOTES

To facility the usage of schema evolution features of ROOT, move to rootcling for dictionary generation.

This is WIP and for very early sharing and debugging.


cmake/podioMacros.cmake
```diff
@@ -260,58 +260,46 @@ endfunction()
...
-
-  add_library(${dict_name} SHARED)
-  target_link_libraries(${dict_name} PUBLIC
-    ${CORE_LIB}
-    podio::podio
-    ROOT::Core
-    )
-  target_include_directories(${dict_name} PUBLIC
-    $<BUILD_INTERFACE:${ARG_OUTPUT_FOLDER}>
-    $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
```
tmadlener, line 308: All of this is now handled properly by `ROOT_GENERATE_DICTIONARY`?

tmadlener (commented):
This looks very sensible to me so far. I few clarifying questions inline and two bigger points here.

We probably have to add a new exclusion rule to the `clang-format` check in `pre-commit`, as that will otherwise complain about unformated LinkDef files (or we treat them the same way as other headers during generation, such that they fall out of that properly formatted already).

There also seem to be missing include guards that fail the build here: https://github.com/AIDASoft/podio/blob/master/python/templates/DatamodelDefinition.h.jinja2
(Not sure how that worked so far ...)

Finally, once this is finished, we will probably have to bump the minimal ROOT version to one that ships a working (for us) version of `ROOT_GENERATE_DICTIONARY`.

cmake/podioBuild.cmake
```diff
@@ -50,7 +50,7 @@ macro(podio_set_compiler_flags)
   # build Catch2 locally without them
   SET(CXX_FLAGS_CMAKE_DEFAULTS "${CMAKE_CXX_FLAGS}")
 
-  SET(COMPILER_FLAGS -Wshadow -Wformat-security -Wno-long-long -Wdeprecated -fdiagnostics-color=auto -Wall -Wextra -pedantic -Weffc++)
+  SET(COMPILER_FLAGS -Wshadow -Wformat-security -Wno-long-long -Wdeprecated -fdiagnostics-color=auto -Wall -Wextra -pedantic -Wno-error=deprecated-declarations -Wno-error=shadow -Wno-error=deprecated-copy-dtor)
```
tmadlener, line 53: Are these mainly for decluttering the compiler output during development, or are some of these necessary for rootcling?

cmake/podioTest.cmake
```diff
@@ -23,7 +23,12 @@ function(CREATE_PODIO_TEST sourcefile additional_libs)
   add_executable( ${name} ${sourcefile} )
   add_test(NAME ${name} COMMAND ${name})
 
-  target_link_libraries(${name} PRIVATE TestDataModel ExtensionDataModel ${additional_libs})
+if (ENABLE_SIO)
+  target_link_libraries(${name} PRIVATE TestDataModel ExtensionDataModel ${additional_libs} TestDataModelSioBlocks ExtensionDataModelSioBlocks)
```
tmadlener, line 27: I am not sure this is necessary, as the `SioBlocks` libraries are only dynamically loaded during runtime, and I assume that the linker will simply ignore them during linking, because nothing in the actual executable depends on them.

python/templates/DictLinkDef.h.jinja2
```diff
@@ -0,0 +1,27 @@
...
+#pragma link off all classes;
+#pragma link off all functions;
+#pragma link C++ nestedclasses;
+
+{% for class in components %}
+{{ class_selection(class) }}
+{% endfor %}
+
+{% for class in datatypes %}
+{{ class_selection(class) }}
```
tmadlener, line 21: Do we also need the `Mutable` ones here if we want to make them available in the python bindings? Or does this still work the same as with the `selection.xml` where things work (maybe due to the `Collection`)? For practical purposes this probably is of somewhat lower priority (see #431), but I would not like to accidentally break things.

cmake/podioMacros.cmake
```diff
@@ -297,21 +193,10 @@ function(PODIO_ADD_ROOT_IO_DICT dict_name CORE_LIB HEADERS SELECTION_XML)
...
+  ROOT_GENERATE_DICTIONARY(${dict_name} ${HEADERS} LINKDEF ${selectionfile}
```
tmadlener, line 196: Do we need to bump the minimum version of ROOT for this? Practically we have mainly been running with 6.26 and 6.28 in any case for quite some time, so we might want to do that in any case.

src/RootLinkDef.h
```diff
@@ -0,0 +1,12 @@
+#ifdef __CLING__
+
+  #pragma link off all globals;
+  #pragma link off all classes;
+  #pragma link off all functions;
+  #pragma link C++ nestedclasses;
+
+  #pragma link C++ class podio::ROOTFrameReader;
```
tmadlener, line 8: Stupid question: Why does the Reader not need a `+`?

tests/CMakeLists.txt
```diff
@@ -61,8 +66,8 @@ add_subdirectory(unittests)
 add_subdirectory(dumpmodel)
 
 # Tests that don't fit into one of the broad categories above
-CREATE_PODIO_TEST(ostream_operator.cpp "")
-CREATE_PODIO_TEST(write_ascii.cpp "")
+CREATE_PODIO_TEST(ostream_operator.cpp ROOT::Tree)
+CREATE_PODIO_TEST(write_ascii.cpp ROOT::Tree)
```
tmadlener, line 70: Why do these now need to be explicitly linked against `ROOT::Tree`? In principle they are completely standalone and make no use of any ROOT features (at least they shouldn't).

tests/root_io/CMakeLists.txt
```diff
@@ -47,8 +47,10 @@ set_property(TEST check_benchmark_outputs PROPERTY DEPENDS read_timed write_time
 if (DEFINED CACHE{PODIO_TEST_INPUT_DATA_DIR})
   message(STATUS "Using test inputs stored in: "  ${PODIO_TEST_INPUT_DATA_DIR})
   add_executable(read-legacy-files-root read-legacy-files-root.cpp)
-  target_link_libraries(read-legacy-files-root PRIVATE TestDataModel TestDataModelDict podio::podioRootIO)
-
+  target_link_libraries(read-legacy-files-root PRIVATE TestDataModel podio::podioRootIO)
+  if (ENABLE_SIO)
+    target_link_libraries(read-legacy-files-root PRIVATE TestDataModelSioBlocks)
```
tmadlener, line 52: Similar to above, this should not be necessary as this is loaded dynamically at runtime.

cmake/podioMacros.cmake
tmadlener, line 196: This also doesn't seem to create a cmake target(?) At least our EDM4hep tests fail because they can no longer find the target: https://github.com/AIDASoft/podio/actions/runs/5714684545/job/15482494614?pr=460#step:6:979

tmadlener (commented):
This all still looks quite straight forward and I mainly have clarifying questions about some details. One bigger question is why the EventStore python tests have been removed. Is the EventStore no longer working from the python side with this? Then I would remove it all together not just the tests.

CI currently seems to stumble over a polluted environment and potentially the fact that the names of the dictionary libraries change. At least the logs suggest that we now have two dictionaries from which we pick the same symbol names.

cmake/podioBuild.cmake
hegner: That was necessary for rootcling debugging. Not final yet

python/templates/DictLinkDef.h.jinja2
hegner: I think you are right we should make that explicit. Will include in the next update

cmake/podioMacros.cmake
hegner: Good idea

src/RootLinkDef.h
hegner: Because it is a stupid oversight ;-)

cmake/podioMacros.cmake
hegner: It doesn't because it changes the way how the library itself is compiled (linking the dictionary inside for the moment)

tmadlener:
I can build this and run all tests successfully locally (with a key4hep like environment). It looks like CI is mainly broken due to environment issues.

cmake/podioMacros.cmake
```diff
@@ -260,34 +153,37 @@ endfunction()
...
 # Conditionally add the ROOT dictionary to the targets if the corresponding
-# selection xml has been generated by PODIO_GENERATE_DATAMODEL.
+# linkdef file has been generated by PODIO_GENERATE_DATAMODEL.
 #
 # Arguments:
 #    dict_name            Name of the dictionary
 #    CORE_LIB             The core datamodel library (e.g. from PODIO_ADD_DATAMODEL_CORE_LIB)
 #    HEADERS              The list of all header files generated by PODIO_GENERATE_DATAMODEL
-#    SELECTION_XML        The selection.xml file genertaed by PODIO_GENERATE_DATAMODEL (either an absolute path or relative to OUTPUT_FOLDER)
+#    LINKDEF              The LinkDef.h file genertaed by PODIO_GENERATE_DATAMODEL (either an absolute path or relative to OUTPUT_FOLDER)
```
andresailer, line 167:
```suggestion
#    LINKDEF              The LinkDef.h file generated by PODIO_GENERATE_DATAMODEL (either an absolute path or relative to OUTPUT_FOLDER)
```

cmake/podioTest.cmake
andresailer: > will simply ignore them during linking

Only if the linker flag `--as-needed` is set?

hegner:
@tmadlener - that's interference between Reflex and rootcling dictionary generation for the very same classes. So without podio in the original stack all is fine. I think we should give it a try in a separate build.

cmake/podioMacros.cmake
hegner: I have it successfully running on 6.24. So I think we can do without for the moment

tmadlener: I just realize our README still claims 6.08, which has probably not been tested in a year or two, so maybe we should also use this opportunity to bring that up to speed.

hegner: OK. Will put it in there.

hegner:
@tmadlener - can you check this locally?

tmadlener:
When I build this locally, I get an error during installation in edm4hep:

```
CMake Error at edm4hep/cmake_install.cmake:276 (file):
  file INSTALL cannot find
  "/home/tmadlener/work/key4hep/EDM4hep/spack-build-cbvcjvm/edm4hep/edm4hepDictDict.rootmap":
  No such file or directory.
  cmake_install.cmake:84 (include)


make: *** [Makefile:103: install] Error 1
```

I am not yet entirely sure why I get past the `cmake` stage locally, but in CI we get stuck there due to a [missing target](https://github.com/AIDASoft/podio/actions/runs/5939754853/job/16106846136?pr=460#step:6:986):
```
  CMake Error at edm4hep/CMakeLists.txt:16 (add_library):
    add_library cannot create ALIAS target "EDM4HEP::edm4hepDict" because
    target "edm4hepDict" does not already exist.
  
  
  -- Not adding the ROOT dictionaries for edm4hep, because '/home/runner/work/podio/podio/edm4hep/edm4hep/src/selection.xml' does not exist
  CMake Error at edm4hep/CMakeLists.txt:27 (install):
    install TARGETS given target "edm4hepDict" which does not exist.

```

hegner:
the move is not-transparent for the users. adjustments in CMake are needed. If you can confirm that PODIO itself is OK, I can make the necessary changes on edm4hep side

tmadlener:
podio tests run fine for me in a key4hep environment.

python/podio_class_generator.py
```diff
@@ -482,13 +482,13 @@ def _needs_include(self, classname) -> IncludeFrom:
...
 
-  def _create_selection_xml(self):
+  def _create_ROOT_dictionary_selection(self):
     """Create the selection xml that is necessary for ROOT I/O"""
     data = {'components': [DataType(c) for c in self.datamodel.components],
             'datatypes': [DataType(d) for d in self.datamodel.datatypes],
             'old_schema_components': [DataType(d) for d in
                                       self.old_datamodels_datatypes | self.old_datamodels_components]}
-    self._write_file('selection.xml', self._eval_template('selection.xml.jinja2', data))
+    self._write_file('LinkDef.h', self._eval_template('DictLinkDef.h.jinja2', data))
```
tmadlener, line 491: I just realized that there is a check in `_write_file` that puts the `LinkDef.h` file into the include directory with all the other (public) headers. Given that is only necessary for dictionary generation, I would consider it an implementation detail that should rather go into the sources output folder, right?

tmadlener (commented):
Another leftover from the previous approach is the creation of a `SELECTION_XML` list in the generated cmake file:

https://github.com/hegner/podio/blob/312fb6118a8d6b4675d66a68da2726e48b8e6d3b/python/podio_class_generator.py#L467-L468

hegner:
> Another leftover from the previous approach is the creation of a `SELECTION_XML` list in the generated cmake file:
> 
> https://github.com/hegner/podio/blob/312fb6118a8d6b4675d66a68da2726e48b8e6d3b/python/podio_class_generator.py#L467-L468

OK. Will check that.

python/podio_class_generator.py
hegner: Yes, it should at least not appear in the install. Will try to fix that

hegner: Done with the last commits

tests/CMakeLists.txt
tmadlener: If linking to `ROOT::Tree` is necessary, we should make the datamodel library (target) depend on that already, I think.

python/podio_class_generator.py
```diff
@@ -177,7 +177,9 @@ def _eval_template(self, template, data):
   def _write_file(self, name, content):
     """Write the content to file. Dispatch to the correct directory depending on
     whether it is a header or a .cc file."""
-    if name.endswith("h"):
+    if name.endswith("LinkDef.h"):
+      fullname = os.path.join(self.install_dir, "src", name)
+    elif name.endswith("h"):
```
tmadlener, line 182: A very similar check is also need in `_write_cmake_lists_file` to avoid putting the `LinkDef.h` file into the list of generated headers that cmake then consumes.

Here: https://github.com/AIDASoft/podio/blob/60dcd12f062658cbd62442d83514b2eb9f140277/python/podio_class_generator.py#L441

hegner:
Closing this as #472 is now the decided way forward.

Issue #462, opened 2023-07-21
Subset collections of datatypes with VectorMembers do not work


tmadlener:
If a subset collection of a datatype that has any `VectorMembers` is written it cannot be read back with ROOT because the `recast` function tries to access non-existing elements of a vector here:
https://github.com/AIDASoft/podio/blob/6b8eeb1824b0fb19f00b8d581b393282a5542be9/python/templates/Collection.cc.jinja2#L200-L209

The stacktrace looks something like this, when using the `ExampleWithVectorMember` datatype
```
#7  podio::CollectionWriteBuffers::asVector<int> (raw=<error reading variable: Cannot access memory at address 0x20>) at /home/tmadlener/work/AIDASoft/podio/include/podio/CollectionBuffers.h:40
#8  operator() (buffers=..., __closure=0x7fffa1653910) at /home/tmadlener/work/AIDASoft/podio/tests/src/ExampleWithVectorMemberCollection.cc:189
#9  std::__invoke_impl<void, (anonymous namespace)::createBuffers(bool)::<lambda(podio::CollectionReadBuffers&)>&, podio::CollectionReadBuffers&> (__f=...) at /usr/include/c++/12/bits/invoke.h:61
#10 std::__invoke_r<void, (anonymous namespace)::createBuffers(bool)::<lambda(podio::CollectionReadBuffers&)>&, podio::CollectionReadBuffers&> (__fn=...) at /usr/include/c++/12/bits/invoke.h:111
#11 std::_Function_handler<void(podio::CollectionReadBuffers&), (anonymous namespace)::createBuffers(bool)::<lambda(podio::CollectionReadBuffers&)> >::_M_invoke(const std::_Any_data &, podio::CollectionReadBuffers &) (__functor=..., __args#0=...) at /usr/include/c++/12/bits/std_function.h:290
#12 0x00007fb313b7de1f in std::function<void (podio::CollectionReadBuffers&)>::operator()(podio::CollectionReadBuffers&) const (__args#0=..., this=0x7fffa1653910) at /usr/include/c++/12/bits/std_function.h:591
#13 podio::ROOTFrameReader::getCollectionBuffers (this=<optimized out>, catInfo=..., iColl=<optimized out>, reloadBranches=<optimized out>, localEntry=0) at /home/tmadlener/work/AIDASoft/podio/src/ROOTFrameReader.cc:109
[...]

PR #463, opened 2023-07-21
Fix I/O issues of subset collections with vector members

tmadlener:
BEGINRELEASENOTES
- Make sure to only access vector member buffers of collections of datatypes with `VectorMembers` if they actually exist. This is necessary to make subset collections of such datatypes work in I/O. Fixes [#462](https://github.com/AIDASoft/podio/issues/462)
- Add a test that reproduces the original issue and is fixed by this.

ENDRELEASENOTES

### DIFF
--- a/python/podio/test_Frame.py
+++ b/python/podio/test_Frame.py
@@ -18,7 +18,7 @@ EXPECTED_COLL_NAMES = {
     'emptyCollection', 'emptySubsetColl'
     }
 # The expected collections from the extension (only present in the other_events category)
-EXPECTED_EXTENSION_COLL_NAMES = {"extension_Contained", "extension_ExternalComponent", "extension_ExternalRelation"}
+EXPECTED_EXTENSION_COLL_NAMES = {"extension_Contained", "extension_ExternalComponent", "extension_ExternalRelation", "VectorMemberSubsetColl"}
 
 # The expected parameter names in each frame
 EXPECTED_PARAM_NAMES = {'anInt', 'UserEventWeight', 'UserEventName', 'SomeVectorData', 'SomeValue'}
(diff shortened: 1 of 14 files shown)


python/podio/test_Frame.py
```diff
@@ -16,7 +16,10 @@
     'emptyCollection', 'emptySubsetColl'
     }
 # The expected collections from the extension (only present in the other_events category)
-EXPECTED_EXTENSION_COLL_NAMES = {"extension_Contained", "extension_ExternalComponent", "extension_ExternalRelation"}
+EXPECTED_EXTENSION_COLL_NAMES = {
+    "extension_Contained", "extension_ExternalComponent", "extension_ExternalRelation",
```
hegner, line 20: can you fix the indentation / formatting?

tests/read_frame.h
```diff
@@ -60,6 +61,14 @@ void processExtensions(const podio::Frame& event, int iEvent, podio::version::Ve
   ASSERT(structs[2].y == 2 * iEvent, "struct value not as expected");
 }
 
+void checkVecMemSubsetColl(const podio::Frame& event) {
+  const auto& subsetColl = event.get<ExampleWithVectorMemberCollection>("VectorMemberSubsetColl");
+  const auto& origColl = event.get<ExampleWithVectorMemberCollection>("WithVectorMember");
+  ASSERT(subsetColl.isSubsetCollection(), "subset collection not read back as a sbuset collection");
```
hegner, line 67: simple typo in mediate ("sbuset")

hegner (changes requested):
Logic and new tests seem fine to me. Just some nitpicking about formatting and typos. See comments.

tmadlener:
Now also fixing the second issue reported in https://github.com/AIDASoft/podio/issues/462#issuecomment-1649866017

Unless we want to have that in another PR?

python/podio/test_Frame.py
hegner: Sorry, my bad. Seems my phone display was playing me tricks.

PR #465, opened 2023-08-14
python/templates: add references between Objects and Collection types

veprbl:
BEGINRELEASENOTES
- Object's collection type can now be referenced as `Object::collection`. Conversely, the object type is reachable as `ObjectCollection::object`.

ENDRELEASENOTES

This should allow to remove some codegen hack from EICrecon https://github.com/eic/EICrecon/blob/1f6727948696b70fc29b1fe582f23f2e37c476b2/src/services/io/podio/make_datamodel_glue.py#L76-L84

cc @wdconinc @nathanwbrei

### DIFF
--- a/python/templates/Collection.h.jinja2
+++ b/python/templates/Collection.h.jinja2
@@ -48,6 +48,7 @@ A Collection is identified by an ID.
 */
 class {{ class.bare_type }}Collection : public podio::CollectionBase {
 public:
+  using value_type = {{ class.bare_type }};
   using const_iterator = {{ class.bare_type }}CollectionIterator;
   using iterator = {{ class.bare_type }}MutableCollectionIterator;
 
--- a/python/templates/Object.h.jinja2
+++ b/python/templates/Object.h.jinja2
@@ -22,6 +22,7 @@
 
 {{ utils.namespace_open(class.namespace) }}
 class Mutable{{ class.bare_type }};
+class {{ class.bare_type }}Collection;
 
 {{ macros.class_description(class.bare_type, Description, Author) }}
 class {{ class.bare_type }} {
@@ -31,6 +32,9 @@ class {{ class.bare_type }} {
   friend class {{ class.bare_type }}CollectionIterator;
 
 public:
+  using mutable_type = Mutable{{ class.bare_type }};
+  using collection_type = {{ class.bare_type }}Collection;
+
 {{ macros.constructors_destructors(class.bare_type, Members) }}
 
 public:
--- a/python/templates/MutableObject.h.jinja2
+++ b/python/templates/MutableObject.h.jinja2
@@ -32,6 +32,8 @@ class Mutable{{ class.bare_type }} {
   friend class {{ class.bare_type }};
 
 public:
+  using object_type = {{ class.bare_type }};
+  using collection_type = {{ class.bare_type }}Collection;
 
 {{ macros.constructors_destructors(class.bare_type, Members, prefix='Mutable') }}
   /// conversion to const object


python/templates/Collection.h.jinja2
```diff
@@ -48,6 +48,7 @@ A Collection is identified by an ID.
 */
 class {{ class.bare_type }}Collection : public podio::CollectionBase {
 public:
+  using object = {{ class.bare_type }};
```
tmadlener, line 51:
```suggestion
  using value_type = {{ class.bare_type }};
```
Then the renaming `object` -> `object_` should also become obsolete.

Do we need / want the `Mutable` version here as a member type.

python/templates/Object.h.jinja2
```diff
@@ -31,6 +32,9 @@ class {{ class.bare_type }} {
   friend class {{ class.bare_type }}CollectionIterator;
 
 public:
+  using mutable_ = Mutable{{ class.bare_type }};
+  using collection = {{ class.bare_type }}Collection;
```
tmadlener, line 36:
```suggestion
  using mutable_type = Mutable{{ class.bare_type }};
  using collection_type = {{ class.bare_type }}Collection;
```
For overall consistency and ease of use.

tmadlener (commented):
Thanks for this proposal. I think this is very useful and also necessary (e.g. for #257). So my main concern for this PR here is to make sure that we do this comprehensively and consistently (also taking into account other c++ conventions).

Overall podio itself does not (yet) have a convention (or at least not one that is followed consistently) for member types. Since some of the exiting member types try to mimic the ones from [`std::vector`](https://en.cppreference.com/w/cpp/container/vector) for the collections, I think we should try to just continue there and try to follow the same convention of using `<xyz>_type` to define member types.

Then one question is which member types do the different (user level) classes need?

Finally one that mainly concerns Jana2 and how it is used there; Are there any concerns wrt breaking things when we introduce this?

veprbl:
> Thanks for this proposal. I think this is very useful and also necessary (e.g. for #257). So my main concern for this PR here is to make sure that we do this comprehensively and consistently (also taking into account other c++ conventions).
> 
> Overall podio itself does not (yet) have a convention (or at least not one that is followed consistently) for member types. Since some of the exiting member types try to mimic the ones from [`std::vector`](https://en.cppreference.com/w/cpp/container/vector) for the collections, I think we should try to just continue there and try to follow the same convention of using `<xyz>_type` to define member types.

Good suggestion.

> 
> Then one question is which member types do the different (user level) classes need?

The only type reference that we were truly missing was from object to collection, because no object method or member would provide that. I don't see what else we could add, the Obj, Data and CollectionData are not part of the public interface.

> 
> Finally one that mainly concerns Jana2 and how it is used there; Are there any concerns wrt breaking things when we introduce this?

Since this only adds fields, this should be backwards-compatible with any codebase.

python/templates/Collection.h.jinja2
veprbl: My assumption was that coll::value_type::mutable_type would be sufficient.

python/templates/Object.h.jinja2
```diff
@@ -31,6 +32,9 @@ class {{ class.bare_type }} {
   friend class {{ class.bare_type }}CollectionIterator;
 
 public:
+  using mutable_type = Mutable{{ class.bare_type }};
+  using collection_type = {{ class.bare_type }}Collection;
```
tmadlener, line 36: Thinking about this here, I think the `Mutable` objects should also have at least the `collection_type`. Mainly for symmetry reasons and also because it would allow a (cumbersome) way to come back from the `Mutable` types to the default ones.

veprbl: Done!

tmadlener:
Thanks for address the comments. Could you rebase this onto the latest master/merge that in to pick up the updated CI configuration, please?

tmadlener (approved):

PR #469, opened 2023-08-21
CI: use clang16 nightly builds

andresailer:
BEGINRELEASENOTES
- Tests: update required catch2 version to 3.4 for builds with c++20
- CI: use clang16 on el9 (alma9), instead of clang12 on cs7, using c++20
- CI: disable key4hep-release-based tests (`tabulate` available)

ENDRELEASENOTES

Hmm: these nightly builds are with c++20...

### DIFF
--- a/.github/workflows/key4hep.yml
+++ b/.github/workflows/key4hep.yml
@@ -13,8 +13,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        release: ["sw.hsf.org/key4hep",
-                  "sw-nightlies.hsf.org/key4hep"]
+        include:
+          - release: "sw.hsf.org/key4hep"
+            RNTUPLE: OFF
+          - release: "sw-nightlies.hsf.org/key4hep"
+            RNTUPLE: ON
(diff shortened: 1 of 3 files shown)


andresailer:
c++20 needs a newer version of catch2

andresailer:
key4hep stack still fails, now complaining about too old catch2
```
  CMake Error at tests/unittests/CMakeLists.txt:10 (find_package):
    Could not find a configuration file for package "Catch2" that is compatible
    with requested version "3.1".
```

What do you want to do about this: Abort? Retry? Ignore?

tmadlener:
No `tabulate` yet in the key4hep release. In that case I would vote for removing the `RNTUPLE` matrix for the `key4hep` workflows again, and instead wait for a new release where everything should work again.

andresailer:
So just drop the key4hep-release based build you mean?

tmadlener (approved):
Looks good now. Just need to remember to re-enable the key4hep release once it can be used again.

Thanks for taking care of this.

PR #472, opened 2023-09-01
Schema evolution based on ROOT and Reflex dictionaries

hegner:
BEGINRELEASENOTES
- Allow comparison of data schemata across versions
- Provide syntax to clarify user intentions in schema evolution
- Provide schema evolution implementation based on ROOT backend
- Include infrastructure for future support for schema evolution in other backends
- Documentation for schema evolution functionality

ENDRELEASENOTES

Schema evolution based on old Reflex dictionary approach.


hegner:
@tmadlener - this is the current status w/ Reflex based schema evolution. It contains a debug output. The problem is that the value of `y_old` read in is already claimed to be zero. Opening with bare ROOT the value however is correct.
If playing with not-evolved members like `x` we get the proper number. Thus my assumption is that there is no proper read before this custom code kicks in. Still - this code crashes if the original file does not contain an `y_old` leaf. So I would assume another ROOT bug behind this.

hegner:
@tmadlener - seems I never reported back in the tracker. so here for the records. Newer ROOT versions do not suffer from this issue any more.

tmadlener (commented):
This is working for me locally with ROOT v6.28/06 and as far as the schema evolution part is concerned seems to be complete (TBC at tomorrows meeting). I have left a few inline comments for cleanup / nitpicking.

As we have already discussed in private handling multiple old versions will be done in a separate PR. From a technical point of view that should "just be another loop" over all the older versions.

The roundtrip tests that dump the model from a written file are currently failing because the generation of the dumped model does not generate any schema evolution code, but the diffing against the original model has schema evolution code now:
```console
92: + diff -ru /home/tmadlener/work/AIDASoft/podio/build/tests/root_io/example_frame.root.dumped_datamodel/datamodel /home/tmadlener/work/AIDASoft/podio/tests/datamodel
92: Only in /home/tmadlener/work/AIDASoft/podio/tests/datamodel: ExampleWithArrayv2Data.h
92: Only in /home/tmadlener/work/AIDASoft/podio/tests/datamodel: ExampleWithNamespacev2Data.h
92: Only in /home/tmadlener/work/AIDASoft/podio/tests/datamodel: ExampleWithUserInitv2Data.h
92: Only in /home/tmadlener/work/AIDASoft/podio/tests/datamodel: NamespaceStructv2.h
```

This leads me to a point on the naming and file layout of the generated code, currently we will have a
- `FooData` in `FooData.h`
- `FoovXData` in `FoovXData.h`

I think this could be cleaner if we had a `vX` namespace and potentially a `vX` subfolder to put "old" headers into. (The latter could also make the roundtrip tests slightly easier to fix).

Finally, I think we should consider splitting this PR into two:
- One that only has the refactoring of the jinja templates that happens in the early commits here which leads to quite a bit of unrelated noise here. We would then also avoid having these mixed up with the "interesting" changes when we squash in the end.
- The actual changes for schema evolution.

python/templates/Collection.cc.jinja2
```diff
@@ -223,6 +193,36 @@ bool registerCollection() {
...
+        {{ package_name }}::meta::schemaVersion,
+        podio::SchemaEvolution::noOpSchemaEvolution,
+        podio::SchemaEvolution::Priority::AutoGenerated
+      );
+    }
+
+{% if old_schema_version is defined %}
+    // register a buffer creation function for the schema evolution buffer
+   // SCHEMA EVOLUTION: Not yet required with only ROOT backend
+   // factory.registerCreationFunc("{{ class.full_type }}Collection", {{ old_schema_version }}, createBuffersV{{old_schema_version}}); //TODO
```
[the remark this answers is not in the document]
hegner: correct.

python/podio_class_generator.py
```diff
@@ -264,6 +378,28 @@ def _process_datatype(self, name, definition):
...
+            if member.name == schema_change.member_name_new:
+              member_type = member.full_type
+
+          iorule = RootIoRule()
+          iorule.sourceClass = type_name
+          iorule.targetClass = type_name
+          iorule.version = self.old_schema_version.lstrip("v")
+          iorule.source = f'{member_type} {schema_change.member_name_old}'
+          iorule.target = schema_change.member_name_new
+          iorule.code = f'std::cout<< "reading {schema_change.member_name_old} with value " << onfile.{schema_change.member_name_old} << std::endl; {iorule.target} = onfile.{schema_change.member_name_old};' # noqa
```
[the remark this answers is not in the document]
hegner: done

python/podio_class_generator.py
```diff
@@ -248,12 +288,86 @@ def _process_component(self, name, component):
...
+          for member in component['Members']:
+            if member.name == schema_evolution.member_name_new:
+              member.name = schema_evolution.member_name_old
+          component['class'] = DataType(name + self.old_schema_version)
+        else:
+          raise NotImplementedError
+      self._fill_templates('Component', component)
+      self.root_schema_component_names.add(name + self.old_schema_version)
+
+  def _replaceComponentInPaths(self, oldname, newname, paths):
```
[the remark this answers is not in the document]
hegner: done

python/podio_class_generator.py
```diff
@@ -248,12 +288,86 @@ def _process_component(self, name, component):
...
+          dir(member)
+#          iorule = RootIoRule()
+#          iorule.sourceClass = member.full_type
+#          iorule.targetClass = member.full_type
+#          iorule.version = 2 #self.schema_version.lstrip("v")
+#          iorule.target = "y"
+#          iorule.source = "int y_old"
+#          iorule.code = "std::cout<< onfile.y_old << y << std::endl; y = onfile.y_old;"
+#          if len(self.root_schema_iorules) == 0:
+#            self.root_schema_iorules.add(iorule)
```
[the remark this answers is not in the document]
hegner: correct. removed

hegner:
Thanks for checking. I was addressing most of the points now.

Concerning the split into two PRs. The part you want to have in a separate PR is probably commit b65c6eecde5101834497cd71a98881d5fb19bee4

I could cherry pick that one for a new PR if that is what you want.

When it comes to namespaces and directory structure for the old data - most of the work will actually be in CMake, and not in the code generation part.  Do we want to do it in a separate PR?

tmadlener:
> Concerning the split into two PRs. The part you want to have in a separate PR is probably commit https://github.com/AIDASoft/podio/commit/b65c6eecde5101834497cd71a98881d5fb19bee4

Yes exactly. Plus this one (to not break things again) https://github.com/AIDASoft/podio/pull/472/commits/6b8fe156308161289e212b67e1bfc2e2d5dea4b4

python/podio_class_generator.py
```diff
@@ -248,12 +288,86 @@ def _process_component(self, name, component):
...
+            if member.name == schema_evolution.member_name_new:
+              member.name = schema_evolution.member_name_old
+          component['class'] = DataType(name + self.old_schema_version)
+        else:
+          raise NotImplementedError
+      self._fill_templates('Component', component)
+      self.root_schema_component_names.add(name + self.old_schema_version)
+
+  def _replaceComponentInPaths(self, oldname, newname, paths):
+    """Replace component in paths"""
```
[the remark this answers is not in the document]
tmadlener: Ah sorry, I missed the `if` in the loop. I thought all elements were touched, but they aren't. Then we leave it like this.

tmadlener:
One other observation: The key4hep-nightlies seem to work for the schema evolution part, but they only come with root 6.28/04, so maybe we don't even need the latest version?

python/podio_class_generator.py
```diff
@@ -115,23 +135,24 @@ def __init__(self, yamlfile, install_dir, package_name, io_handlers, verbose, dr
...
       self._create_selection_xml()
 
     self._write_cmake_lists_file()
-    self.process_schema_evolution()
 
     self.print_report()
 
-  def process_schema_evolution(self):
+  def preprocess_schema_evolution(self):
     """Process the schema evolution"""
```
andresailer, line 156:
```suggestion
    """Preprocess the schema evolution"""
```

python/podio_class_generator.py
```diff
@@ -156,6 +178,17 @@ def process_schema_evolution(self):
           print(warning)
         sys.exit(-1)
 
+      # now go through all the io_handlers and see what we have to do
+      if 'ROOT' in self.io_handlers:
+        for item in root_filter(comparator.schema_changes):
+          schema_evolutions = self.root_schema_dict.get(item.klassname)
```
andresailer, line 184: This should replace the entire loop body
```python
# add whatever is relevant to our ROOT schema evolution
self.root_schema_dict.setdefault(item.klassname, []).append(item)
```
Or use a `defaultdict`

python/podio_class_generator.py
```diff
@@ -248,12 +288,78 @@ def _process_component(self, name, component):
...
+          self._replace_component_in_paths(member.array_type, member.array_type + self.old_schema_version,
+                                        schema_evolution_datatype['includes_data'])
+          member.full_type = member.full_type.replace(member.array_type, member.array_type + self.old_schema_version)
+          member.array_type = member.array_type + self.old_schema_version
+
+      else:
+        if member.full_type in self.root_schema_dict:
+          needs_schema_evolution = True
+          # prepare the ROOT I/O rule
+          print(member.full_type)
```
andresailer, line 347: remove debugging print?

python/podio_class_generator.py
```diff
@@ -248,12 +288,78 @@ def _process_component(self, name, component):
...
+          dir(member)
```
andresailer, line 348: remove?

python/podio_schema_evolution.py
```diff
@@ -107,6 +107,17 @@ def __init__(self, name, member_name_old, member_name_new):
     super().__init__(f"'{self.name}': member '{self.member_name_old}' renamed to '{self.member_name_new}'.")
 
 
+class RootIoRule:
```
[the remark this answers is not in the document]
hegner: that actually got moved there after the discussion with Thomas.

python/podio_class_generator.py
hegner: yap.

python/podio_class_generator.py
hegner: yap

python/podio_class_generator.py
hegner: good observation. simplified now.

hegner:
> One other observation: The key4hep-nightlies seem to work for the schema evolution part, but they only come with root 6.28/04, so maybe we don't even need the latest version?

Any 6.28 is enough. 6.24 is not. And 6.26 I did not try.

tmadlener:
Ah alright. Then I simply misunderstood "latest release". 6.26 did not work for me.

hegner:
@tmadlener - if you are fine, can you review and approve?

python/podio_class_generator.py
```diff
@@ -247,12 +271,76 @@ def _process_component(self, name, component):
...
+
+    # ROOT schema evolution preparation
+    # Compute and prepare the potential schema evolution parts
+    schema_evolution_datatype = copy.deepcopy(datatype)
+    needs_schema_evolution = False
+    # check whether it has a renamed member
+    # if name in self.root_schema_dict.keys():
+    # for member in schema_evolution_datatype['Members']:
+    # if
+    # then check for components with a renamed member
```
tmadlener, line 316:
```suggestion
```
Comment seems to no longer correspond with the code that follows.

python/podio_class_generator.py
```diff
@@ -263,6 +351,28 @@ def _process_datatype(self, name, definition):
...
+          iorule = RootIoRule()
+          iorule.sourceClass = type_name
+          iorule.targetClass = type_name
+          iorule.version = self.old_schema_version.lstrip("v")
+          iorule.source = f'{member_type} {schema_change.member_name_old}'
+          iorule.target = schema_change.member_name_new
+          iorule.code = f'{iorule.target} = onfile.{schema_change.member_name_old};'
+          self.root_schema_iorules.add(iorule)
+        else:
+          raise NotImplementedError("Schema evolution for this type not yet implemented")
```
tmadlener, line 374: This is potentially confusing, without further information on what "type" is.

tmadlener (commented):
Looks good. Can you confirm whether the comments I marked are really left overs from earlier? The Error message can be addressed in a follow up PR as well, just wanted to have it on a "list" somewhere.

tmadlener (approved):

PR #473, opened 2023-09-01
Juliareset

Ananya2003Gupta:
Completed:
- Fixed the issue with the `‘.jl’ `file, which was caused by an empty namespace. Moving forward, the file will be named as `‘<package_name>.jl’` .
- Resolved the empty namespace issue in the `ParentModule.jl.jinja2` template by utilizing the `package_name` as the module name.
- Adjusted the `Constructor.jl.jinja2` template to include default parameters in the constructors.
- Altered the code generation approach to support multiple namespaces within a single Julia file (using modules).
- Updated the tests to accommodate recent modifications.
- Resolved the main scope issue by following Single Module implementation.


graeme-a-stewart (changes requested):
Hi - looks pretty good, a few comments embedded. Are the tests as extensive as the tests for the C++ rendering of the data model? It's not too exhaustive.

Also, we need to rethink the "forward declaration" solution we use too.

python/podio/generator_utils.py
```diff
@@ -29,6 +29,37 @@ def _prefix_name(name, prefix):
   return name
 
 
+def get_julia_type(cpp_type=None, is_array=False, array_type=None, array_size=None):
+  """Parse the given c++ type to a Julia type"""
+  builtin_types_map = {"int": "Int32", "float": "Float32", "double": "Float64",
+                       "bool": "Bool", "long": "Int32", "unsigned int": "UInt32",
+                       "unsigned long": "UInt32", "char": "Char", "short": "Int16",
+                       "long long": "Int64", "unsigned long long": "UInt64"}
```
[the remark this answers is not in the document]
tmadlener: `long long` is 64 bits in practice at the moment (as far as the table [on the integers](https://en.cppreference.com/w/cpp/language/types) shows). We could put in a header that encodes these "assumptions" into the c++ code, that essentially consists of `static_assert`s of the form
```cpp
static_assert(sizeof(long long int) == 8, "long long int is not 64 bits wide");
```

graeme-a-stewart:
BTW, you also failed some `pylint` tests that should be fixed

Ananya2003Gupta:
I have corrected all the changes that were asked.

jmcarcell:
Is this optional? I would say even if Julia is installed there will be cases when we'll want podio not to have the Julia interface by default. Of course if Julia is not installed then it should work as it works now, tests included

Ananya2003Gupta:
> Is this optional? I would say even if Julia is installed there will be cases when we'll want podio not to have the Julia interface by default. Of course if Julia is not installed then it should work as it works now, tests included

Yes, the code generation will work regardless of whether julia is present in the system or not, as julia code for a particular datamodel is generated using jinja2 templates. And yes the tests for julia will only run if the users have julia in their system.

Ananya2003Gupta:
> Is this optional? I would say even if Julia is installed there will be cases when we'll want podio not to have the Julia interface by default. Of course if Julia is not installed then it should work as it works now, tests included

Also, that's why I have introduced a new parameter --lang (-l). By default the language is cpp, but if the users want to generate julia code, they can by specifying `-l=julia` or `--lang=julia`.

graeme-a-stewart (approved):
Hi - I think this isn't the last word on the EDM in Julia, but it works and allows us to develop from here without very painful rebases, so I am happy to merge this now.

hegner:
@Ananya2003Gupta @graeme-a-stewart - I'd agree once the conflicts are resolved ;-)

Ananya2003Gupta:
> @Ananya2003Gupta @graeme-a-stewart - I'd agree once the conflicts are resolved ;-)

Yes, it's done : )

graeme-a-stewart:
Hi @Ananya2003Gupta 

I got a test failure when I was running in your branch:

```julia
ERROR: LoadError: ArgumentError: Package StaticArrays not found in current path.
- Run `import Pkg; Pkg.add("StaticArrays")` to install the StaticArrays package.
```

Can you fix that one? You shouldn't make assumptions about which packages are installed in the test environment.

tmadlener:
I think this would just need to be put into the README as a dependency and also how to install it, right? CI seems to be configured to do that and the tests are running and succeeding.

Ananya2003Gupta:
> I think this would just need to be put into the README as a dependency and also how to install it, right? CI seems to be configured to do that and the tests are running and succeeding.

Yes. The users need to have StaticArrays package installed in their system. Otherwise it will throw this error and ask them to install it.

graeme-a-stewart:
> Actually you would have to install StaticArrays package in your system, then only it would work. This can be done: https://discourse.julialang.org/t/install-a-packages-from-a-list/30920/6

But you should not fail the *test*, we are testing the code, not the user's environment. Probably the best way is to have a `Project.toml` specifically for the test.

For the user, document that `StaticArrays` is required, but the user should not have to install it in their default Julia environment (which is really bad practice).

Ananya2003Gupta:
> > Actually you would have to install StaticArrays package in your system, then only it would work. This can be done: https://discourse.julialang.org/t/install-a-packages-from-a-list/30920/6
> 
> But you should not fail the _test_, we are testing the code, not the user's environment. Probably the best way is to have a `Project.toml` specifically for the test.
> 
> For the user, document that `StaticArrays` is required, but the user should not have to install it in their default Julia environment (which is really bad practice).

Okay

Ananya2003Gupta:
I have added a try-catch block in `unittest.jl` to check if `StaticArrays` is available, if not, install it locally.

Ananya2003Gupta:
@hegner @tmadlener can you please review the PR? If there are anymore changes required I would do them.

python/podio_gen/generator_utils.py
```diff
@@ -29,6 +29,37 @@ def _prefix_name(name, prefix):
   return name
 
 
+def get_julia_type(cpp_type=None, is_array=False, array_type=None, array_size=None):
```
tmadlener, line 32:
```suggestion
def get_julia_type(cpp_type, is_array=False, array_type=None, array_size=None):
```
Since this doesn't do anything if `cpp_type` is actually `None`. Better make not passing one an error rather than going on and producing something that will most likely crash down the line.

python/podio_gen/generator_utils.py
```diff
@@ -29,6 +29,37 @@ def _prefix_name(name, prefix):
...
+  """Parse the given c++ type to a Julia type"""
+  builtin_types_map = {"int": "Int32", "float": "Float32", "double": "Float64",
+                       "bool": "Bool", "long": "Int64", "unsigned int": "UInt32",
+                       "unsigned long": "UInt64", "char": "Char", "short": "Int16",
+                       "long long": "Int64", "unsigned long long": "UInt64"}
+  # is a global type as described in test_MemberParser.py #L121
+  if cpp_type and cpp_type.startswith("::"):
```
tmadlener, line 39:
```suggestion
  if cpp_type.startswith("::"):
```
No need to check here, since we will always have a value.

python/podio_gen/generator_utils.py
```diff
@@ -29,6 +29,37 @@ def _prefix_name(name, prefix):
...
+  if not is_array:
+    if cpp_type.startswith('std::'):
+      cpp_type = cpp_type[5:]
+    if cpp_type in ALLOWED_FIXED_WIDTH_TYPES:
+      regex_string = re.split("(u|)int(8|16|32|64)_t", cpp_type)
+      cpp_type = regex_string[1].upper() + "Int" + regex_string[2]
+      return cpp_type
+
+  else:
+    array_type = get_julia_type(cpp_type=array_type)
```
tmadlener, line 53:
```suggestion
    array_type = get_julia_type(array_type)
```
We can also skip the keyword here, since now it is clear which is the only required argument.

python/podio_gen/generator_utils.py
```diff
@@ -151,6 +185,9 @@ def __init__(self, name, **kwargs):
     else:
       self.namespace, self.bare_type = _get_namespace_class(self.full_type)
 
+    self.julia_type = get_julia_type(cpp_type=self.bare_type, is_array=self.is_array,
```
tmadlener, line 188:
```suggestion
    self.julia_type = get_julia_type(self.bare_type, is_array=self.is_array,
```
Similarly here. No need to specify the keyword here.

python/podio_class_generator.py
```diff
@@ -305,46 +354,52 @@ def _process_datatype(self, name, definition):
...
-      if member.is_array:
-        if member.array_type in self.root_schema_dict:
-          needs_schema_evolution = True
-          self._replace_component_in_paths(member.array_type, member.array_type + self.old_schema_version,
-                                           schema_evolution_datatype['includes_data'])
-          member.full_type = member.full_type.replace(member.array_type, member.array_type + self.old_schema_version)
-          member.array_type = member.array_type + self.old_schema_version
+    if self.proglang == "julia":
+      self._fill_templates('MutableStruct', datatype)
+
```
tmadlener, line 359:
```suggestion
    return datatype
```
Can we return early here and not have a huge diff below just from re-indenting everything from the additional `if self.proglang == "cpp":`?

python/podio_class_generator.py
```diff
@@ -271,23 +310,33 @@ def _process_component(self, name, component):
...
-        else:
-          raise NotImplementedError
+    component['upstream_edm'] = self.upstream_edm
+    component['upstream_edm_name'] = ''
+    if self.upstream_edm:
+      component['upstream_edm_name'] = self.upstream_edm.options["includeSubfolder"].split("/")[-2].capitalize()
+
+    if self.proglang == "julia":
+      self._fill_templates('MutableStruct', component)
+
```
tmadlener, line 320:
```suggestion
    return component
```
Can we return early here to avoid having a larger than necessary diff (from re-indenting due to the `if self.proglang == "cpp":`)?

tmadlener (commented):
I can't comment too much on the generated Julia code. @graeme-a-stewart and @hegner are better equipped to comment on that. There are a few minor things (see also the inline comments)

- I would merge `params.jinja2` and `abstracttypes.jinja2` into one `julia_helpers.jinja2` file and include that where necessary in the main jinja2 template files
- The `unittest.jl` file should be in `tests/unittests`, rather than in `tests`
- I would like to have an option `ENABLE_JULIA` in the top level CMakeLists.txt to toggle Julia support manually, rather than relying on finding the Julia executable alone.
- Can you bring back the `BEGINRELEASENOTES` and `ENDRELEASENOTES` in the top level comment of the PR? Those are used as delimiter for our tagging script.


More generally, I am not sure whether mixing in the Julia code generator into the existing c++ code generator is the best idea, as that is already complicated enough on its own and sprinkling a few `if lang == "xxx"` in there doesn't help in that regard. Having said that, I think we can also fix this after this PR and ideally split them into to separate generator classes that each handle exactly one language.

python/podio_class_generator.py
Ananya2003Gupta: I indented the code `if self.proglang == "cpp":` so that it runs for c++ only and not for julia.

Ananya2003Gupta:
I'll do the changes that you suggested.

Ananya2003Gupta:
> More generally, I am not sure whether mixing in the Julia code generator into the existing c++ code generator is the best idea, as that is already complicated enough on its own and sprinkling a few `if lang == "xxx"` in there doesn't help in that regard. Having said that, I think we can also fix this after this PR and ideally split them into to separate generator classes that each handle exactly one language.

How exactly would we design this?

python/podio_gen/generator_utils.py
Ananya2003Gupta: @tmadlener Can't make this change as it is throwing error in build itself for example datamodel.
```
File "/home/user/podio/python/podio_gen/generator_utils.py", line 39, in get_julia_type
    if cpp_type.startswith("::"):
AttributeError: 'NoneType' object has no attribute 'startswith'
CMake Error at cmake/podioMacros.cmake:219 (message):
  Could not generate datamodel 'datamodel'.  Check your definition in
  'datalayout.yaml'
  tests/CMakeLists.txt:8 (PODIO_GENERATE_DATAMODEL)
```

python/podio_gen/generator_utils.py
Ananya2003Gupta: I feel we should keep these for the clarity of contributors and maintainers as this will help in easy debugging and understanding of code.

python/podio_gen/generator_utils.py
Ananya2003Gupta: Here, as well, since we are specifying for others and not for `self.bare_type`, it will look a little odd.
So either, we keep it as 
```
self.julia_type = get_julia_type(cpp_type=self.bare_type, is_array=self.is_array,
                                     array_type=self.array_type, array_size=self.array_size)
```
Or
```
self.julia_type = get_julia_type(self.bare_type, self.is_array,
                                     self.array_type, self.array_size)
```

python/podio_gen/generator_utils.py
tmadlener: This looks a bit like a logic error somewhere, as just from the calling sites there should not be any way to call this with a `cpp_type` that is `None`. Is that the complete stacktrace you got? It is hard to tell just from this bit where the original call came from.

python/podio_gen/generator_utils.py
tmadlener: I think that is OK. In this case the difference between the two is that `cpp_type` is always necessary, whereas all the others are in optional.

tmadlener:
The basic idea for splitting classes would be to have a sort of base class that does the common processing and them two dedicated classes for cpp and julia respectively that deal with the specifics. However, as I said, I would first get this PR merged as it is with the things mixed and then we can separate things later, when there are no multiple other PRs touching code generation in flight, so that we don't have to keep all of them synchronized.

python/podio_gen/generator_utils.py
Ananya2003Gupta: I feel there is an else block too that's checking for array type. So if cpp_type is None, it will not work for all the three ifs and we have to pass the array_type present in else condition.

```
if cpp_type and cpp_type.startswith("::"):
    cpp_type = cpp_type[2:]
  if cpp_type in builtin_types_map:
    return builtin_types_map[cpp_type]

  if not is_array:
    if cpp_type.startswith('std::'):
      cpp_type = cpp_type[5:]
    if cpp_type in ALLOWED_FIXED_WIDTH_TYPES:
      regex_string = re.split("(u|)int(8|16|32|64)_t", cpp_type)
      cpp_type = regex_string[1].upper() + "Int" + regex_string[2]
      return cpp_type

  else:
    array_type = get_julia_type(cpp_type=array_type)
    if '::' in array_type:
      array_type = array_type.split('::')[1]
    if array_type not in builtin_types_map.values():
      array_type = array_type + 'Struct'
    return f"MVector{{{array_size}, {array_type}}}"

  return cpp_type
```

Ananya2003Gupta:
> The basic idea for splitting classes would be to have a sort of base class that does the common processing and them two dedicated classes for cpp and julia respectively that deal with the specifics. However, as I said, I would first get this PR merged as it is with the things mixed and then we can separate things later, when there are no multiple other PRs touching code generation in flight, so that we don't have to keep all of them synchronized.

Got it. 
Also, I am not sure why the key4hep / build-and-test (sw-nightlies.hsf.org/key4hep, ON) (pull_request) is failing.
```
 [374/381] Linking CXX shared library tests/schema_evolution/libTestDataModel_v3SioBlocks.so
  [375/381] Generating TestDataModel_v3Dict.cxx, TestDataModel_v3DictDict.rootmap
  Warning: Unused class rule: ExampleWithUserInitv2
  Warning: Unused class rule: ex42::ExampleWithNamespacev2
  Warning: Unused class rule: ExampleWithArrayv2
  [376/381] Building CXX object tests/schema_evolution/root_io/CMakeFiles/write_old_data_root.dir/write_old_data_root.cpp.o
  [377/381] Linking CXX executable tests/schema_evolution/root_io/write_old_data_root
  [378/381] Building CXX object tests/schema_evolution/root_io/CMakeFiles/read_new_data_root.dir/read_new_data_root.cpp.o
  [379/381] Building CXX object tests/schema_evolution/CMakeFiles/TestDataModel_v3Dict.dir/TestDataModel_v3Dict.cxx.o
  [380/381] Linking CXX shared library tests/schema_evolution/libTestDataModel_v3Dict.so
  [381/381] Linking CXX executable tests/schema_evolution/root_io/read_new_data_root
  ninja: build stopped: cannot make progress due to previous errors.
  Error: Process completed with exit code 1.
```

tmadlener:
That looks like a temporary issue unrelated to your changes here: https://github.com/AIDASoft/podio/actions/runs/7003341864/job/19048875848?pr=473#step:4:986

We will simply use the next run if there are more changes, or rerun this later and it should pass.

Ananya2003Gupta:
> I would like to have an option `ENABLE_JULIA` in the top level CMakeLists.txt to toggle Julia support manually, rather than relying on finding the Julia executable alone.

For this, should I keep an ENABLE_JULIA toggle like ENABLE_SIO ? If julia executable is found, enable the toggle to on and then run the tests written for julia in tests/CMakeLists.txt?

Ananya2003Gupta:
> That looks like a temporary issue unrelated to your changes here: https://github.com/AIDASoft/podio/actions/runs/7003341864/job/19048875848?pr=473#step:4:986
> 
> We will simply use the next run if there are more changes, or rerun this later and it should pass.

Okay. Yes I need to make one more change for ENABLE_JULIA. I may do the rebase as well. And force push the changes.

tmadlener:
Yes, exactly.

python/podio_gen/generator_utils.py
Ananya2003Gupta: This member: `std::array<int, 4> p [mm]` caused the error. I think that's why we have an else condition that handles `array_type` like these and for such members `cpp_type` is None.

tmadlener: OK. That is slightly weird as at least from the parsing point of view, `std::array<int, 4> p [mm]` should definitely parse into a `MemberVariable` that has `array_type` set. I have just briefly checked the unit tests in [`test_MemberParser.py`](https://github.com/AIDASoft/podio/blob/master/python/podio_gen/test_MemberParser.py). It looks like we do not yet have an explicit test for this, so maybe this is indeed not working as intended.

I think the quickest test would be to add a unittest using exactly this type and see if fails, i.e. add something similar to, e.g. this:
https://github.com/AIDASoft/podio/blob/858c0ff0b841705d1b18aafd57569fcbd1beda91/python/podio_gen/test_MemberParser.py#L230-L234

to the `test_parse_unit` function using `std::array<int, 4> p [mm]` as input and see if it works. (You can run the python unittests even without a successul build, via `python test_MemberParser.py`)

Ananya2003Gupta:
I have added the ENABLE_JULIA toggle button.
We can re run the two failed tests.

python/podio_gen/generator_utils.py
Ananya2003Gupta: This is how the tests would look like:
```
    parsed = parser.parse('std::array<int, 4> p [mm]', False)
    self.assertEqual(parsed.full_type, 'std::array<int, 4>')
    self.assertEqual(parsed.name, 'p')
    self.assertEqual(parsed.array_type, 'int')
    self.assertTrue(parsed.is_builtin_array)
```
I am unable to run the tests though. Getting this error.
```
from podio_gen.podio_config_reader import MemberParser, DefinitionError
ModuleNotFoundError: No module named 'podio_gen'
```

tmadlener: I just ran them quickly o my end, and this is working. So the error has to be in somewhere in the whole chain of calling things, such that the `arra_type` is either overwritten, or you end up calling `get_julia_type` with something where you should not call it.

You might have to run the tests in the `python` subdirectory via `python -m unittest podio_gen/test_MemberParser.py`

Ananya2003Gupta: I'll see what's going wrong.

tmadlener: Ah right, that makes sense (sort of at least). I have the feeling that the `is_array` argument is redundant in this case as `bare_type` will only be `None` in case of `std::array` variables, but I am not entirely sure. If you want you can try to see if that is actually the case. However, I would be happy with the following (much less time consuming) as well:

- Make the argument required (i.e. remove the `=None` from `cpp_type` in the argument list)
- Keep the check in the function body as it is and add a comment that summarizes what you have figured out in this thread (something like, cpp_type can be None in case of array members)

Ananya2003Gupta: Okay. And for the other two ?
>array_type = get_julia_type(array_type)
>self.julia_type = get_julia_type(self.bare_type, is_array=self.is_array,

Ananya2003Gupta:
@tmadlener I have made all the asked changes.

tests/CMakeLists.txt
```diff
@@ -63,6 +63,35 @@ add_subdirectory(schema_evolution)
...
+  find_program(Julia_EXECUTABLE julia)
+  if (Julia_EXECUTABLE)
+    message(STATUS "Found Julia, will add the Julia tests")
+    set(UNITTEST_SCRIPT_PATH ${CMAKE_CURRENT_SOURCE_DIR}/unittests/unittest.jl)
+    add_test(NAME julia-unittests COMMAND julia ${UNITTEST_SCRIPT_PATH})
+    set_tests_properties(julia-unittests PROPERTIES
+      WORKING_DIRECTORY ${CMAKE_CURRENT_LIST_DIR}
+      )
+  else()
+    message(WARNING "Julia not found.")
```
tmadlener, line 89:
```suggestion
    message(WARNING "Julia not found. Cannot run the Julia tests")
```

tests/CMakeLists.txt
```diff
@@ -63,6 +63,35 @@ add_subdirectory(schema_evolution)
...
+    IO_BACKEND_HANDLERS ${PODIO_IO_HANDLERS}
+    OUTPUT_FOLDER ${CMAKE_CURRENT_SOURCE_DIR}/extension_model
+    LANG julia
+    )
+
+  find_program(Julia_EXECUTABLE julia)
+  if (Julia_EXECUTABLE)
+    message(STATUS "Found Julia, will add the Julia tests")
+    set(UNITTEST_SCRIPT_PATH ${CMAKE_CURRENT_SOURCE_DIR}/unittests/unittest.jl)
+    add_test(NAME julia-unittests COMMAND julia ${UNITTEST_SCRIPT_PATH})
```
tmadlener, line 84:
```suggestion
    add_test(NAME julia-unittests COMMAND julia ${CMAKE_CURRENT_SOURCE_DIR}/unittests/unittest.jl)
```
`UNITTEST_SCRIPT_PATH` is a bit too generic, and it's not reused more than once, so there is no need to introduce it.

.github/workflows/key4hep.yml
```diff
@@ -28,6 +28,7 @@ jobs:
           mkdir build install
           cd build
           cmake -DENABLE_SIO=ON \
+            -DENABLE_JULIA=ON \
```
tmadlener, line 31:
```suggestion
```
This stack doesn't come with julia, so no need to run it here
```suggestion
            -DENABLE_JULIA=OFF \
```

tmadlener (approved):
Very nice. Thank you. I have just a few final minor comments. I can also just do them via the github web interface.

I will let @hegner or @graeme-a-stewart press the button.

.github/workflows/key4hep.yml
Ananya2003Gupta: Wait, let me make all these changes that you suggested.

PR #477, opened 2023-09-08
Reduce unnecessary template instantiations

### LINKED ISSUE #475: PODIO_JSON_OUTPUT causes nlohmann/json.hpp in each include file, could be json_fwd.hpp instead
- OS version: ubuntu 23.04
 - Compiler version: GCC 12.3
 - PODIO version: 0.16.5
 - Reproduced by: compiling github.com/eic/EICrecon ;-) (sorry)

When we compile EDM4hep with json support, the installed CMake config includes the line:
```cmake
set_target_properties(EDM4HEP::edm4hep PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "PODIO_JSON_OUTPUT"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "podio::podio;nlohmann_json::nlohmann_json"
)
```
This causes `PODIO_JSON_OUTPUT` to be defined for all code that uses EDM4hep. This potentially affects a lot of files.

In the case of EICrecon, this results in the following build analysis (`clang -ftime-trace` with https://github.com/aras-p/ClangBuildAnalyzer):
```
**** Templates that took longest to instantiate:
134874 ms: nlohmann::basic_json<>::parse<const char *> (163 times, avg 827 ms)
 93952 ms: nlohmann::basic_json<>::basic_json (815 times, avg 115 ms)
 85261 ms: nlohmann::detail::parser<nlohmann::basic_json<>, nlohmann::detail::i... (163 times, avg 523 ms)
 71201 ms: nlohmann::detail::parser<nlohmann::basic_json<>, nlohmann::detail::i... (163 times, avg 436 ms)
 53404 ms: nlohmann::basic_json<>::json_value::json_value (817 times, avg 65 ms)
```
I.e. we spend almost 500s on just instantiating the json template in `json.hpp` out of the total compilation run. And we don't use json anywhere in this code :-\

It would be useful if downstream code could make the decision whether or not to enable this json support, or to use the `json_fwd.hpp` forwarding header, e.g. https://github.com/nlohmann/json#integration (this potentially requires some specific features of the nlohmann-json install, to ensure this header is available).

Happy to move this to EDM4hep too. Not sure where this is best addressed, but since the include lines live in the jinja templates here, I figured this would be the best place to start the discussion.

tmadlener:
BEGINRELEASENOTES
- Use `nlohmann/json_fwd.hpp` in headers to reduce unnecessary template instantiations. Fixes #475 

ENDRELEASENOTES

@wdconinc could you give this a try and see if this solves things already? ~Currently there are only the components left with the original include. That is mainly because they do not have a `.cc` file yet, but if necessary we can also introduce that.~

### DIFF
--- a/python/podio_class_generator.py
+++ b/python/podio_class_generator.py
@@ -209,7 +209,6 @@ have resolvable schema evolution incompatibilities:")
 
     endings = {
         'Data': ('h',),
-        'Component': ('h',),
         'PrintInfo': ('h',)
         }.get(template_base, ('h', 'cc'))
 
--- a/python/templates/CMakeLists.txt
+++ b/python/templates/CMakeLists.txt
@@ -4,6 +4,7 @@ set(PODIO_TEMPLATES
   ${CMAKE_CURRENT_LIST_DIR}/CollectionData.cc.jinja2
   ${CMAKE_CURRENT_LIST_DIR}/CollectionData.h.jinja2
   ${CMAKE_CURRENT_LIST_DIR}/Component.h.jinja2
+  ${CMAKE_CURRENT_LIST_DIR}/Component.cc.jinja2
   ${CMAKE_CURRENT_LIST_DIR}/Data.h.jinja2
   ${CMAKE_CURRENT_LIST_DIR}/Obj.h.jinja2
   ${CMAKE_CURRENT_LIST_DIR}/Obj.cc.jinja2
--- a/python/templates/Collection.cc.jinja2
+++ b/python/templates/Collection.cc.jinja2
@@ -13,6 +13,10 @@
 {{ include }}
 {% endfor %}
 
+#if defined(PODIO_JSON_OUTPUT) && !defined(__CLING__)
+#include "nlohmann/json.hpp"
+#endif
+
 // standard includes
 #include <stdexcept>
 #include <iomanip>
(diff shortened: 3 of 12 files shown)


tmadlener:
Failing edm4hep workflow should be fixed with https://github.com/key4hep/EDM4hep/pull/224

wdconinc:
I'll try to fit in a test. I haven't set up my local environments to make it easy to swap out podio without recompiling all of DD4hep...

tmadlener:
Thanks. In principle this should be visible e.g. when compiling anything that links against edm4hep, right? Maybe I can do some before-/after tests myself to see if I see a difference.

wdconinc:
We've found it to be most noticeable in our reconstruction, where every file includes at least some data model file somewhere. But it also includes dd4hep and acts with EDM4hep, so spack will be wanting to recompile those too...

wdconinc:
Confirmed that this branch removes the expensive json template instantiations. (We're back to being dominated by Eigen template instantiations...)

tmadlener:
Thank you for the very quick turnaround. Quickly triggering CI again before merging.

PR #480, opened 2023-09-12
Introduce a createBuffers macro for the jinja template

hegner:
BEGINRELEASENOTES
- Move the code generation of buffers into the 'create_buffers' macro

ENDRELEASENOTES

This extracts the create_buffers change from #472 as discussed in there.

PR #482, opened 2023-09-15
rename CMake macro createBuffers into create_buffers

hegner:
BEGINRELEASENOTES
- rename CMake macro createBuffers into create_buffers

ENDRELEASENOTES

As discussed in #480


tmadlener (approved):
Thanks!

PR #483, opened 2023-09-15
Clarify error message in case of not implemented schema changes

hegner:
BEGINRELEASENOTES
- Clarify error message in case of not implemented schema changes
ENDRELEASENOTES
follow up to #472


tmadlener:
The string representation of `schema_change` also contains information about the involved datatype/component as well as the member, right?

hegner:
so far is has all the information. We have to make sure for the future ones that this remains the case.

PR #486, opened 2023-09-15
Fix reading relations to unpersisted objects

tmadlener:
BEGINRELEASENOTES

- Make sure to initialize `ObjectID`s to *untracked* to properly track whether they have been added to a Frame or not
- Change `CollectionIDTable` interfaces of `name` and `collectionID` to return `optional` to signal whether a collection (ID) is known to the table.
  - Avoids having to do the lookup twice to check existence and a subsequent retrieval
- Fix bug of overly shared CollectionIDTable in `ROOTNTupleReader` that was uncovered by the CollectionIDTable switch to optional returns.
- Switch tests from EventStore to Frame based I/O.
- Fix bug in Frame based I/O that lead to crashes when trying to resolve relations to unpersisted objects.

ENDRELEASENOTES

### DIFF
--- a/include/podio/CollectionIDTable.h
+++ b/include/podio/CollectionIDTable.h
@@ -4,6 +4,7 @@
 #include <cstdint>
 #include <memory>
 #include <mutex>
+#include <optional>
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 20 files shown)


tmadlener:
This seems to be a breaking change at least for the legacy I/O algorithms in k4FWCore. Should we aim for a smoother migration or do we just merge the two PRs "in sync"?

hegner (approved):
Thanks. Looks good to. In particular adding the default initialization

PR #493, opened 2023-09-28
Make `Object::id()` return an ObjectID since unsigned has become useless

### LINKED ISSUE #438: Generated `id` method for user facing types has become useless
The `id` method we generate for user facing object has become very hard to interpret with #412 :

https://github.com/AIDASoft/podio/blob/d8294d2a2aad3c737c7b9f07ad9e5b2d826f50b7/python/templates/macros/declarations.jinja2#L96

tmadlener:
BEGINRELEASENOTES
- Make `[Mutable]Object::id()` return a `podio::ObjectID` instead of `unsigned`, since the latter has become useless with #412. Fixes #438 
- Add an `operator<<(std::ostream&)` for `podio::ObjectID`

ENDRELEASENOTES

### DIFF
--- a/include/podio/ObjectID.h
+++ b/include/podio/ObjectID.h
@@ -2,6 +2,8 @@
 #define PODIO_OBJECTID_H
 
 #include <cstdint>
+#include <iomanip>
+#include <ostream>
 
 namespace podio {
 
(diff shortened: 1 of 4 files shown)


jmcarcell:
Note that this makes comparisons between `id()` fail since it now returns a `podio::ObjectID` instead of an unsigned that doesn't implement at least some of the comparison operators

hegner:
Can you open a ticket for that? (currently absent from work)

PR #497, opened 2023-10-03
Move podio_PYTHON_DIR to the top level CMakeLists

jmcarcell:
BEGINRELEASENOTES
- Move podio_PYTHON_DIR to the top level CMakeLists so that it is set even when BUILD_TESTING is off

ENDRELEASENOTES

This is useful when building other packages at the same time and having BUILD_TESTING=OFF. Alternatively, to reduce the number of lines in the top level CMakeLists the `if(BUILD_TESTING)` could be moved inside the `tests/CMakeLists.txt` and this variable could be always be set there

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -178,6 +178,11 @@ install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/NOTICE
 
 #--- project specific subdirectories -------------------------------------------
 add_subdirectory(src)
+
+# Set the podio_PYTHON_DIR manually here because the macros in tests expect it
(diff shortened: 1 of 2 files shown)


tmadlener:
Shouldn't dependent packages get this via the `find_package`, where it is set? This is mainly a curiosity, I don't see any reason to not merge this.

jmcarcell:
Yes if you build podio, install it and then build something else. No if you use `add_directory` or `FetchContent` to build podio together with some other packages. In general everything that is not being hardcoded like this path works fine with this approach

tmadlener (approved):

PR #498, opened 2023-10-03
Silence when SIO is missing and exit if the podio dictionary is not found

jmcarcell:
BEGINRELEASENOTES
- Silence when SIO is missing so that the messages are not shown every time a datamodel is generated when a podio installation already exists
- Simplify `__init__.py`, remove an unneeded `import ROOT`

ENDRELEASENOTES

### DIFF
--- a/python/podio/__init__.py
+++ b/python/podio/__init__.py
@@ -5,28 +5,19 @@ from .__version__ import __version__
 
 from .podio_config_reader import *  # noqa: F403, F401
 
-import ROOT  # pylint: disable=wrong-import-order
-
-# Track whether we were able to dynamially load the library that is built by
-# podio and enable certain features of the bindings only if they are actually
-# available.
-_DYNAMIC_LIBS_LOADED = False
-
-# Check if we can locate the dictionary wthout loading it as this allows us to
-# silence any ouptput. If we can find it, we can also safely load it
-if ROOT.gSystem.DynamicPathName("libpodioDict.so", True):
-  ROOT.gSystem.Load("libpodioDict.so")
+# Try to load podio, this is equivalent to trying to load libpodio.so and will
+# error if libpodio.so is not found but work if it's found
+try:
   from ROOT import podio
-
-  _DYNAMIC_LIBS_LOADED = True
-
-if _DYNAMIC_LIBS_LOADED:
+except ImportError:
+  print('Unable to load podio, make sure that libpodio.so is in LD_LIBRARY_PATH')
+else:
   from .frame import Frame
   from . import root_io, reading
 
   try:
     # We try to import the sio bindings which may fail if ROOT is not able to
-    # load the dictionary in this case they have most likely not been built and
+    # load the dictionary. In this case they have most likely not been built and
     # we just move on
     from . import sio_io
   except ImportError:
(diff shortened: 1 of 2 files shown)


tmadlener (commented):
This would break in completely pristine environments, where there is no previous podio installed, I think. We are only building the `libpodioDict.so`, but we still need some parts of the python bindings to run the code generation (which is done at CMake stage, so no chance that we can build the dictionary before hand).

jmcarcell:
PR updated. I'm a bit upset that there isn't a way of loading a library silently so that we wouldn't need to check first and then load (that will do the same check again)

tmadlener:
>  However, for the cases when there is a mismatch between ROOT versions `ROOT.gInterpreter` may not say anything but `ROOT.gSystem` does

By "not say anything" you mean that `gInterpreter` doesn't give us a different return value in case there is a version mismatch, or simply that it doesn't print a warning message?

python/podio/__init__.py
```diff
@@ -5,28 +5,19 @@
...
+# Try to load podio, this is equivalent to trying to load libpodio.so and will
+# error if libpodio.so is not found but work if it's found
+try:
   from ROOT import podio
-
-  _DYNAMIC_LIBS_LOADED = True
-
-if _DYNAMIC_LIBS_LOADED:
+except ImportError:
+  print('Unable to load podio, make sure that libpodio.so is in LD_LIBRARY_PATH')
```
tmadlener, line 13: Do we have to re-raise this here, or does this already propagate up after the log output?

jmcarcell:
> > However, for the cases when there is a mismatch between ROOT versions `ROOT.gInterpreter` may not say anything but `ROOT.gSystem` does
> 
> By "not say anything" you mean that `gInterpreter` doesn't give us a different return value in case there is a version mismatch, or simply that it doesn't print a warning message?

I'm not sure about this. I just tested building podio with 6.28.04 and then loading the podio library with a new build with master and none of the two ways complain but I remember seeing some message when the ROOT versions were not the same... I initially thought that maybe it doesn't say anything because `gSystem.Load` calls `gInterpreter.Load` with an extra check about the versions: https://root.cern.ch/doc/master/TSystem_8cxx_source.html#l01929 and https://root.cern.ch/doc/master/TSystem_8cxx_source.html#l01933. When the library doesn't exist `gSystem.Load` complains with a message while `gInterpreter.Load` just returns -1. So maybe it would be fine to use `gInterpreter.Load` to do the lookup once only

python/podio/__init__.py
tmadlener, line 13:
```suggestion
  print('Unable to load podio, make sure that libpodio.so is in LD_LIBRARY_PATH')
  raise
```
Otherwise the `except` block will swallow the `ImportError` and we will probably fail later with some more or less horrible c++ / python intermixed output

python/podio/__init__.py
```diff
@@ -5,28 +5,19 @@
...
+else:
```
tmadlener, line 14: If `from ROOT import podio` also the rest will fail probably, as that will try to load at least one library that wants `libpodio` as well.

python/podio/__init__.py
jmcarcell: `raise` added, now it fails importing when `libpodio.so` is not in `LD_LIBRARY_PATH`, for example

tmadlener:
Picked up by #511 and being merged as part of that.

PR #502, opened 2023-10-07
Make it possible to dispose of unconsumed collection buffers

### LINKED ISSUE #500: ROOTFrameData leaks collection buffers that are not requested by the Frame
When reading files via the `ROOTFrameReader` memory increases steadily. ~(This most likely also applies to other readers)~. The minimal reproducer looks something like this

```cpp
#include "podio/ROOTFrameReader.h"

void demo() {
  auto reader = podio::ROOTFrameReader();
  reader.openFile("example_frame.root");
    
  for (int i = 0; i < reader.getEntries("events"); ++i) {
     auto data = reader.readNextEntry("events"); 
  }  
}
```

A quick run to valgrind seems to indicate that some of the data that is allocated in `createBuffers` becomes unreachable somewhere in the chain from creating it to constructing `CollectionData` from the buffers. Currently trying to figure out where things go wrong and whether this is a ROOT exclusive problem or whether this is a general podio issue (which I assume currently).

tmadlener:
BEGINRELEASENOTES

- Add a `deleteBuffers` function that is populated at generation time to the `CollectionReadBuffers` to make it possible to dispose of unconsumed buffers. This fixes the main leaks described in #500 
- Make the `ROOTFrameData` clean up all unconsumed buffers at desctruction.
- Make sure to delete buffers that are no longer necessary in the CollectionData constructor. This fixes some more leaks described in #500 

ENDRELEASENOTES

Fixes #500

### DIFF
--- a/include/podio/CollectionBuffers.h
+++ b/include/podio/CollectionBuffers.h
@@ -53,15 +53,19 @@ struct CollectionReadBuffers {
   using CreateFuncT = std::function<std::unique_ptr<podio::CollectionBase>(podio::CollectionReadBuffers, bool)>;
   using RecastFuncT = std::function<void(CollectionReadBuffers&)>;
 
+  using DeleteFuncT = std::function<void(CollectionReadBuffers&)>;
+
   CollectionReadBuffers(void* d, CollRefCollection* ref, VectorMembersInfo* vec, SchemaVersionT version,
-                        std::string_view typ, CreateFuncT&& createFunc, RecastFuncT&& recastFunc) :
+                        std::string_view typ, CreateFuncT&& createFunc, RecastFuncT&& recastFunc,
+                        DeleteFuncT&& deleteFunc) :
       data(d),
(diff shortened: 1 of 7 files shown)


tmadlener:
This should fix #500 (still have to test that on a larger scale) but it is a bit cumbersome and the "proper" solution would be to make the `CollectionReadBuffers` owners of their data and giving them move-only semantics such that the flow of ownership of the data is expressed much more clearly.

I would be happy to merge this as is to fix the current leak and then address the "proper" solution after v1.0, because I am not entirely sure whether it can be made to work with the EventStore.

hegner:
Is it WIP or are you finished with this now? I am asking because of more commits coming in. Thanks

tmadlener:
Just finished a few more larger scale tests and valgrind and this fixes the leaks of unconsumed collection buffers.

hegner (approved):

PR #503, opened 2023-10-07
fix: install podio-vis

### LINKED ISSUE #501: podio-vis not installed by make install
- OS version: ubuntu 23.04
 - Compiler version: gcc 12.3.0
 - PODIO version: master
 - Reproduced by: cmake -Bbuild -S. && make -Cbuild install
 - Goal: While `podio-dump` from `tools/` is installed by the CMakeLists.txt in that directory, `podio-vis` is not installed. This makes it impossible to use `podio-vis` unless you keep the source tree around and use it from there.

wdconinc:
Fixes #501 by installing podio-vis.

BEGINRELEASENOTES
- Install podio-vis

ENDRELEASENOTES


tmadlener (approved):
Thanks a lot.

Issue #499, opened 2023-10-05
Allow to limit the collections that are read


tmadlener:
Currently all collections will always be available, but it might be useful to limit the collections that are read / become available after reading them.

PR #504, opened 2023-10-11
Make it possible to read only a subset of available collections in readers

tmadlener:
BEGINRELEASENOTES
- Add a `collsToRead` argument to `readEntry` and `readNextEntry` for the `ROOTFrameReader` to allow limiting the collections that are actually read.
- **This requires a bump in the minimum version of ROOT for RNTuple support to 6.32** as the previous RNTuple implementation apparently cannot handle switching between reading different sets of fields.

ENDRELEASENOTES

This is an (early) proposal for introducing this functionality which would in the end address #499 

- [x] Interface design: Do we want to have the possibility to change this frame-by-frame?
  - Alternatively would have to introduce a configuration method that takes a category name and a list of collections.
- [x] Should we only pass on the limited collection id table?
  - Interplay with Frame and assumptions there
- [x] Add tests and checks to avoid breaking things when collection names are passed that are not even available from the data
  - The way the check is implemented only names that exist in the frame will be checked. Non-existent names in `collsToRead` will be **silently** ignored.
- [x] Same capabilities for RNTuple reader
- [x] Documentation (once the rest has settled)
- [x] Ensure that files that have been read with a limited set of collections and are then written again, can be read without issues.
- [x] Needs #698 as it also starts using `std::ranges`.
- [x] Needs #719 to pass on `dev3` stacks
- [x] Needs #724 
- [x] Includes #730 
- [x] Check whether this can be used to ignore collections with unsupported schema evolutions.
- [x] python bindings
- [x] Includes #732

### DIFF
--- a/src/RNTupleReader.cc
+++ b/src/RNTupleReader.cc
@@ -138,11 +138,13 @@ std::vector<std::string_view> RNTupleReader::getAvailableCategories() const {
   return cats;
 }
 
-std::unique_ptr<ROOTFrameData> RNTupleReader::readNextEntry(const std::string& name) {
-  return readEntry(name, m_entries[name]);
+std::unique_ptr<ROOTFrameData> RNTupleReader::readNextEntry(const std::string& name,
+                                                            const std::vector<std::string>& collsToRead) {
+  return readEntry(name, m_entries[name], collsToRead);
 }
 
-std::unique_ptr<ROOTFrameData> RNTupleReader::readEntry(const std::string& category, const unsigned entNum) {
+std::unique_ptr<ROOTFrameData> RNTupleReader::readEntry(const std::string& category, const unsigned entNum,
+                                                        const std::vector<std::string>& collsToRead) {
   if (m_totalEntries.find(category) == m_totalEntries.end()) {
     getEntries(category);
   }
@@ -179,6 +181,9 @@ std::unique_ptr<ROOTFrameData> RNTupleReader::readEntry(const std::string& categ
   const auto& collInfo = m_collectionInfo[category];
 
   for (size_t i = 0; i < collInfo.id.size(); ++i) {
+    if (!collsToRead.empty() && std::ranges::find(collsToRead, collInfo.name[i]) == collsToRead.end()) {
+      continue;
+    }
     const auto& collType = collInfo.type[i];
     const auto& bufferFactory = podio::CollectionBufferFactory::instance();
     auto maybeBuffers =
--- a/src/RNTupleReader.cc
+++ b/src/RNTupleReader.cc
@@ -178,15 +178,11 @@ std::unique_ptr<ROOTFrameData> RNTupleReader::readEntry(const std::string& categ
   auto readerIndex = upper - 1 - m_readerEntries[category].begin();
 
   ROOTFrameData::BufferMap buffers;
-#if ROOT_VERSION_CODE >= ROOT_VERSION(6, 31, 0)
   // We need to create a non-bare entry here, because the entries for the
   // parameters are not explicitly (re)set and we need them default initialized.
   // In principle we would only need a bare entry for the collection data, since
   // we set all the fields there in any case.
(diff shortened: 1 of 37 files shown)


tmadlener:
Revived this as this is likely to become an issue for moving EDM4hep forwards otherwise. I have settled on making it possible to pass the collections that should be made available on a frame-by-frame basis as that is the most flexible and is also rather simple to implement.

I still have to make sure that the idTable that is available from the constructed Frames is properly cleaned up and also add some tests that passing in unavailable collection names doesn't break the whole thing.

src/RNTupleReader.cc
```diff
@@ -176,6 +178,10 @@ std::unique_ptr<ROOTFrameData> RNTupleReader::readEntry(const std::string& categ
 #endif
 
   for (size_t i = 0; i < m_collectionInfo[category].id.size(); ++i) {
+    if (!collsToRead.empty() &&
+        std::ranges::find(collsToRead, m_collectionInfo[category].name[i]) == collsToRead.end()) {
+      continue;
+    }
```
jmcarcell, line 184:
```suggestion
    if (std::ranges::find(collsToRead, m_collectionInfo[category].name[i]) == collsToRead.end()) {
      continue;
    }
```
Probably nothing is gained by checking if `find` checks or does a for loop over its elements? Same below

jmcarcell: I just edited; I just meant the first empty check as it's deleted in the suggestion

tmadlener: Ah, sorry then I completely misunderstood. The check is negated, so we only want to do the `find` if there are collections to check. If we remove the check, we will stop reading any collections (because if `collsToRead` is empty then `find` will never find a name in it).

jmcarcell: I think my suggestion is still incorrect because how it is now if `collsToRead` is empty then the first check will always fail and will always continue. So it should be like this, only continue if at least one collection is passed to be read and it is not found there:
```
    if (!collsToRead.empty()) {
      if (std::ranges::find(collsToRead, m_collectionInfo[category].name[i]) == collsToRead.end()) {
        continue;
      }
    }
```
then the existing calls will just pass an empty `collsToRead` and the `continue` will not trigger.

tmadlener: Unless I am missing something the nested `if` and the current implementation with the `&&` are equivalent?

The main question this discussion has triggered for me is whether the `continue` should actually be an exception, because arguably, if I put a collection name into the `collsToRead`, I expect it to be in the `Frame` I get in the end.

jmcarcell: Ah yes, it's correct. Probably an exception is better than simply not reading it.

tmadlener: Now throwing an exception in case a non-existent collection is requested.

tmadlener:
It looks like we don't strictly need to trim down the collection id table, as it is by now really only used for the lookups between hashes and collection names. I will keep it as is and leave trimming that down for now as an optimization that can also be done later.

tmadlener:
It looks like the RNTuple implementation up to (and including) ROOT 6.30 cannot really deal with switching between different sets of Fields when reading. It's possible to read full frames and frames with only a subset of collections separately, but it is not possible to first read full frames and then frames with only a subset of collections with the same reader. I have decided to bump the minimum ROOT version to 6.32 for RNTuple support to make things work for now.

tmadlener:
I have done some (manual) checks with EDM4hep and schema changes that lead to breaks when reading back collections, that are handled gracefully if the affected collections are ignored. I have not yet found a really good way to add automated tests for that in podio directly. They would be nice to have, but I am not sure if it is worth the effort now. I would almost be in favor of merging this rather soon, after some review, and then adding them later, resp. checking that via the EDM4hep build & tests we also do in CI.

tmadlener:
Merging this later today, unless there are comments.

PR #507, opened 2023-10-16
Copy .clang-format to the dumpmodel test directory

jmcarcell:
BEGINRELEASENOTES
- Copy .clang-format to the dumpmodel test directory and fix some tests when the build directory is not a subdirectory of the main directory. In some tests `clang-format` will try to find a `.clang-format` looking in the directories above and if it doesn't exist it will format files differently and some tests will fail.

ENDRELEASENOTES


tmadlener (approved):
Nice spot. Never thought about true out-of-source builds apparently.

PR #508, opened 2023-10-30
Use the cmake ExternalData module to manage test data

jmcarcell:
BEGINRELEASENOTES
- Use the cmake ExternalData module to manage test data. This lets cmake take care of downloading the tests by hash so they can be version controlled. In addition, it's possible to set up a local store using `-DExternalData_OBJECT_STORES=/path/to/store` and it will download the test files there if they are not there but otherwise use them from there, so building from scratch won't download the test files again. 

ENDRELEASENOTES

The downside is that the test files are now hashes since we have a very simple web server to retrieve them. And there is now a file for each test file with its hash, so changing a file means also the hash should be changed. 

I wanted to have a look at this because I didn't like the fact that test files are always downloaded at configuration time and I'm happy now with the local store and will add this to the other key4hep repos with test files.


jmcarcell:
It also seems to be quite fast; since it downloads files as cmake targets I think it may be also doing downloads in parallel

hegner:
@jmcarcell - what's the compilation problem in the sanitizer?

jmcarcell:
I have no idea, there isn't an error message and it seems to be able to build everything. Maybe something transient? Nothing changes for how C++ files are built so it also doesn't make sense that it would fail with this PR.

jmcarcell:
If you click the link now it's fine and in my tests I have not yet seen that problem 🤷 
I hashed every file that was on EOS so it has to work. By the way the SIO test files are not used in tests I think so they will never be downloaded.

tmadlener:
Seems to have been transient.

tmadlener (approved):
Very nice. Thanks :)

tmadlener:
I think that is a pretty nice solution, especially because the symbolic links don't seem to show up in the web interface:
image

PR #509, opened 2023-10-30
Create a variant using clang format and set it to off by default

jmcarcell:
BEGINRELEASENOTES
- Create a variant using clang format and set it to off by default. It is significantly slower to run cmake with this option on.

ENDRELEASENOTES

I don't think most users care and we could run clang-format for the builds in the stacks, but if you are building I think you would rather get to whatever you want to do sooner.

I have some times for configure without downloading the test files. Rebuild means when for example you make a change in a CMakeLists.txt file so that configure has to run again when you try to build:

On my machine (all local): 

|                      | From scratch | Rebuild |
|----------------------|--------------|---------|
| With clang-format    | 16 s         | 14 s    |
| Without clang-format | 6 s          | 4 s     |

On a build node (Alma 9):

|                      | From scratch | Rebuild |
|----------------------|--------------|---------|
| With clang-format    | 27 s         | 20 s    |
| Without clang-format | 20 s         | 13 s    |

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,6 +69,7 @@ option(CREATE_DOC        "Whether or not to create doxygen doc target." OFF)
 option(ENABLE_SIO        "Build SIO I/O support" OFF)
 option(PODIO_RELAX_PYVER "Do not require exact python version match with ROOT" OFF)
 option(ENABLE_RNTUPLE    "Build with support for the new ROOT NTtuple format" OFF)
+option(PODIO_USE_CLANG_FORMAT "Use clang-format to format the code" OFF)
 
 
 #--- Declare ROOT dependency ---------------------------------------------------


tmadlener:
I think `cmake -DPODIO_USE_CLANG_FORMAT=OFF` should already work right now without these changes, since we set this here:
https://github.com/AIDASoft/podio/blob/e44af47e44b595439eb4a62ac9a6893bc46f9b9e/cmake/podioMacros.cmake#L108-L109

jmcarcell:
Yes, the change here would be to have it OFF by default and for stacks and whoever wants it to have it ON since it doesn't have any impact unless you look at the files (and how often does this happen anyway?)

tmadlener:
If we set this to OFF by default, can we at least have one CI workflow where we have it `AUTO` / `ON` with a suitable `clang-format` available? Just to avoid inadvertently breaking anything in the future.

PR #510, opened 2023-11-05
Fix legacy tests

jmcarcell:
BEGINRELEASENOTES
- Fix legacy tests; an extra argument was being passed and default in the .cpp file `example_frame.root` was being used which (almost) always exists (because there is a another test creating it) so it was hard to notice.
ENDRELEASENOTES

Maybe this default should be changed so as to be more explicit so this can't happen but it would make the implementation in the CMakeLists in root_io a bit less clean.

### DIFF
--- a/tests/root_io/CMakeLists.txt
+++ b/tests/root_io/CMakeLists.txt
@@ -59,8 +59,7 @@ target_link_libraries(read-legacy-files-root PRIVATE TestDataModel TestDataModel
 macro(ADD_PODIO_LEGACY_TEST version base_test input_file)
   ExternalData_Add_Test(legacy_test_cases
     NAME ${base_test}_${version}
-    COMMAND ${base_test} ${CMAKE_CURRENT_BINARY_DIR}/../input_files/${input_file}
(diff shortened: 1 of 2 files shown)


tmadlener (approved):
Good catch. Not sure I understand the diff though or why it worked before. Previously it got 3 arguments, so how did it end up reading the default `example_frame.root`?

jmcarcell:
It's the default value and it's changed based on the number of arguments:
https://github.com/AIDASoft/podio/blob/master/tests/root_io/read_frame_root.cpp#L10
And since there is another test creating it with the current version of podio, reading that will never fail. I only noticed it because when running the tests in parallel it didn't work which makes no sense; it was reading the file that was just being created in all the legacy tests and that wasn't working.

tmadlener:
Ah, its a `argc == 2`. Then the fix that would at least have caught this issue would be to also have an additional check for `> 2` in the executable, right?

jmcarcell:
Yep, I added the check

tmadlener (approved):
Thanks.

tmadlener:
closing and re-opening to trigger actions, which seem to have gone missing somehow for the last commit.

PR #485, opened 2023-09-15
Remove deprecated EventStore based I/O functionality

tmadlener:
BEGINRELEASENOTES
- **Remove the deprecated EventStore functionality as announced in #429**

ENDRELEASENOTES

~This is currently only the first steps to remove the EventStore based I/O functionality (as outlined in #429). It is quite far from complete. The main purpose at this point is to get a feeling of how many things need to be touched in the end.~ This removes all traces of the `EventStore` and related readers and writers.

- [x] Needs #486 
- [x] Pre-release (v00-17-01)
- [x] Remove python bindings 
- [x] Fix tests
  - [x] Make unittests independent of EventStore
  - [x] Make some of the I/O tests use Frame based I/O
  - [x] Cleanup python tests
- [x] Cleanup documentation
- [x] #505 
- [x] Needs https://github.com/key4hep/k4FWCore/pull/171

PR #511, opened 2023-11-06
Restructure python parts, splitting code generation and python bindings into separate modules

jmcarcell:
BEGINRELEASENOTES
- Decouple generation tools and files from the rest of the podio python files by creating a new folder called `podio_gen`. **This is a transparent change for users that only use the generator script.**
  - This makes the configuration / generation times negligible again, because we don't load libraries unnecessarily any longer for the generation.
- Simplify the python bindings (`podio`) `__init__.py` and remove the `test_utils` from it.
- Move the tests for writing frames in python to the `tests` folder, where they belong and also test the SIO python writer.

ENDRELEASENOTES

Generation tools do not need to load or know anything about the rest of podio. I think the remaining question is if the tests have to be moved or not, some of them only test generation tools.


jmcarcell:
I didn't move any tests - I mean moving tests from the current folder `podio` to the `podio_gen` folder; for example https://github.com/AIDASoft/podio/blob/master/python/podio/test_ClassDefinitionValidator.py only uses classes from `podio_gen` in this PR but it's in `podio`. It only matters from an organization point of view, and if any test is moved then there is some extra cmake code needed to also find those. On that I don't have an opinion to where this one goes, I don't think there are many more.

tmadlener:
Ah, apologies. Misread the diff. I would move them, just to have them closer to the code they are testing.

python/CMakeLists.txt
```diff
@@ -23,17 +23,25 @@ else()
...
+  )
+
 #--- install templates ---------------------------------------------------------
 install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/templates
   DESTINATION ${podio_PYTHON_INSTALLDIR})
 
 IF (BUILD_TESTING)
   add_test( NAME pyunittest COMMAND python3 -m unittest discover -s ${PROJECT_SOURCE_DIR}/python/podio)
   PODIO_SET_TEST_ENV(pyunittest)
+  add_test( NAME pyunittest_gen COMMAND python3 -m unittest discover -s ${PROJECT_SOURCE_DIR}/python/podio_gen)
```
tmadlener, line 39: `python3 -m unittest discover -s '${PROJECT_SOURCE_DIR}/python -p "test_*.py"` doesn't work to find all tests in both modules?

jmcarcell: Hmm that works, it seems without the `-p` it also picks up other files; I thought it would only pick up the subclassed classes

python/CMakeLists.txt
```diff
@@ -23,12 +23,18 @@ else()
...
+  PATTERN __pycache__ EXCLUDE  # Or pythons caches
+  )
+
 #--- install templates ---------------------------------------------------------
 install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/templates
   DESTINATION ${podio_PYTHON_INSTALLDIR})
 
 IF (BUILD_TESTING)
-  add_test( NAME pyunittest COMMAND python3 -m unittest discover -s ${PROJECT_SOURCE_DIR}/python/podio)
+  add_test( NAME pyunittest COMMAND python3 -m unittest discover -s ${PROJECT_SOURCE_DIR}/python/podio -p "test_*.py")
```
tmadlener, line 37:
```suggestion
  add_test( NAME pyunittest COMMAND python3 -m unittest discover -s ${PROJECT_SOURCE_DIR}/python -p "test_*.py")
```
I think otherwise we are missing the ones that are now in `podio_gen`, right?

jmcarcell: Ah yes, of course, that is what I tested before and it doesn't work. I'm checking...

jmcarcell: So I have no idea how to make it work with a single test. First, when trying to discover from the top directory python will call `dir(podio)` and will fail like in https://github.com/AIDASoft/podio/issues/512. Even finding a workaround some tests fail with errors related to some errors when importing like this
``` python
ImportError: Failed to import test module: podio.test_EventStore
  File "/usr/lib/python3.11/unittest/loader.py", line 407, in _find_test_path
    module = self._get_module_from_name(name)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/unittest/loader.py", line 350, in _get_module_from_name
    __import__(name)
  File "/usr/lib/python3.11/site-packages/ROOT/_facade.py", line 154, in _importhook
    return _orig_ihook(name, *args, **kwds)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ModuleNotFoundError: No module named 'podio.test_EventStore'; 'podio' is not a package
```
so I reverted back to the version that works except pre-commit...

tmadlener:
pre-commit complains about a cyclic import, but it doesn't really make sense to me, as the script should now only import from `podio_gen`, right? 

https://github.com/AIDASoft/podio/actions/runs/6776714508/job/18418800193?pr=511#step:4:427

jmcarcell:
I believe the diagnosis is correct; podio wants to import test_utils and test_utils wants to import podio. Actually the try - except block only seems to work during test time. I'm not sure why it appears now though

tmadlener:
Ah then we can maybe switch to a

```python
from .frame import Frame
```
in `test_utils.py`?

jmcarcell:
That breaks some tests due to relative imports. Another proposal: move test_utils.py to `podio/tests` so that the tests that use it can use it directly. Then the import can be removed from `__init__.py` (and it makes sense that you don't always import something that you only need for tests, it just happens that the import doesn't work now at runtime). It's a bit less organized but anyway there are some python tests already there

tmadlener:
I tried that, but that breaks with the weird c++ error in the comment (or at least it did in the past). I agree, this should really live in the `tests` directory though. Maybe you can give it another try and see if things just magically work now?

tmadlener:
I think I found the issue. If I do

```python
from podio.test_utils import Frame
print(Frame)
```
I get `<class 'podio.frame.Frame'>`. However, if I do
```python
from podio import Frame
print(Frame)
```
I get `<class cppyy.gbl.podio.Frame at 0x5569c6f3fc60>`. So it seems like outside of the `test_utils` the importing of `Frame` does not pick up the wrapper, but rather the c++ class instead.

tmadlener:
I have removed the `sys.modules["podio"] = podio` bit from the non-generator part `__init__.py` to make sure that we always get our wrapper classes in python rather than picking up c++ classes with the same name via the ROOT python bindings accidentally. This also makes it possible to move the `test_utils` more or less completely to the `tests` folder into `write_frame.py` and simply calling that for the different I/O backends.

I think in this case not having all of the c++ bits of podio directly available in python is OK, since for the main use cases we provide wrappers in any case and all the others are effectively mainly there to support the implementation of those, and I don't think we gain too much by exposing them to people in python.

Since this has now become a bit more than just "decoupling the generation files from the rest of the podio python files" I would actually also like to pick up #498 in this PR and make it another general overhaul of the python structure of podio. Mainly because #498 needs to be touched again in any case and we might as well resolve the merge conflict in this PR, rather than later.

tmadlener (approved):

python/podio/test_utils.py
```diff
@@ -2,59 +2,5 @@
 """Utilities for python unittests"""
 
 import os
```
jmcarcell, line 4: Hmm this file only has like 3 lines after all the removals, I'm not sure if it's needed...

tmadlener:
I think `GenericParameters` should be seen as an internal implementation detail of how the Frame supports parameters.

`UserDataCollection`s on the other hand are a good point. Maybe we can try to address them in another PR since Andre is waiting on this for the LCG stacks.

python/podio/test_utils.py
tmadlener: I think I can clean this up in the course of #485

PR #513, opened 2023-11-08
Make sure that the ROOT writers enforce consistency for the Frame contents they write

### LINKED ISSUE #382: ROOTFrameWriter can produce unreadable files without warning
The documentation states that the contents of the first Frame of a category determine the contents of all Frames of a given category:
https://github.com/AIDASoft/podio/blob/8488574fb8b948a53f618887d13040e79e8f9d24/include/podio/ROOTFrameWriter.h#L31-L37

However, there is no currently no warning in case this happens, and the writer will simply ignore additional collections. This also has the potential to break the reader side, especially if the missing collections are not simply the last ones to be added to the frame, e.g.

```cpp
#include "write_frame.h"  // from the tests directory for easier creation of collections

#include "podio/Frame.h"
#include "podio/ROOTFrameWriter.h"

int main() {
  auto writer = podio::ROOTFrameWriter("broken_frame.root");

  auto frame = podio::Frame();
  frame.put(createHitCollection(10), "hits");
  frame.put(createMCCollection(), "mcparticles");
  writer.writeFrame(frame, "events");

  frame = podio::Frame();
  const auto& hits = frame.put(createHitCollection(10), "hits");
  frame.put(createClusterCollection(hits), "clusters");
  frame.put(createMCCollection(), "mcparticles");
  writer.writeFrame(frame, "events");

  writer.finish();

  return 0;
}
```
produces a file for which
```bash
podio-dump -e1 broken_frame.root
```
will run into a segmentation fault, because the collection ID for that is necessary for resolving the relations of the `ExampleMC`s has changed between the frames.

The cause for the segmentation fault is actually a different issue (#381). Nevertheless, there should probably at least be a warning if different contents are present in Frames that should be written with the `ROOTFrameWriter`.

tmadlener:
BEGINRELEASENOTES
- Introduce checks in `ROOTFrameWriter::writeFrame` and `ROOTNTupleWriter::writeFrame` that ensure consistent contents for all Frames of a given category. If inconsistent contents are found an exception is thrown. Before these changes this might lead to a crash or to unreadable files. Fixes #382 
- Refactor `ROOTNTupleWriter` internals to have only one map that keeps track of categories instead of two maps and a set that need to be kept consistent.

ENDRELEASENOTES


tmadlener:
@andresailer @Zehvogel any thoughts about the contents of the exception messages? In principle the only exception that we can trigger from the k4FWCore PodioOutput is the one where a collection is not available. The reason is that there we are guaranteed to have a consistent set of `collsToWrite` as that is dictated by the framework.

src/rootUtils.h
```diff
@@ -272,6 +272,30 @@ inline std::vector<std::string> sortAlphabeticaly(std::vector<std::string> strin
...
 }
 
+/**
+ * Check whether existingColls and candidateColls both contain the same
+ * collection names. Returns false if the two vectors differ in content. Inputs
+ * can have random order wrt each other, but the assumption is that each vector
+ * only contains unique names.
+ */
+inline bool checkConsistentColls(const std::vector<std::string>& existingColls,
+                                 const std::vector<std::string>& candidateColls) {
```
andresailer, line 282: We should print out which collections are missing, and not return on the first one we don't find.

tmadlener: Should we make that part of the exception message? (That might become quite long then). An alternative approach would be to keep the exception in its terse form and then give access to the `collsToWrite` for each category so that analysis can be done on the user side, resp. we could provide something like
```cpp
using namespace std;
tuple<vector<string>, vector<string>> checkConsistency(const vector<string>& collsToWrite, const string& category);
```
which returns the collection names that only exist in either the passed `collsToWrite` or inside the `category`. This could in principle then also be used as a pre-check, before calling `writeFrame`.

tests/unittests/unittest.cpp
```diff
@@ -26,6 +29,13 @@
   #include "podio/SIOReader.h"
 #endif
 
+#ifndef PODIO_ENABLE_RNTUPLE
+  #define PODIO_ENABLE_RNTUPLE 0
+#endif
+#if PODIO_ENABLE_RNTUPLE
```
andresailer, line 35: Why not just `#ifdef PODIO_ENABLE_RNTUPLE` ? usually the pre-compiler flags are either set or not?

src/rootUtils.h
andresailer: Making it part of the exception is also good. But we must log it somewhere! Rather a too long exception message than a too short one.

tests/unittests/unittest.cpp
tmadlener: Simply copied from the SIO approach, where the library with the definition might not be present. However, the root library is always here, so I will make this always defined to the approriate value.

tmadlener: Done.

src/rootUtils.h
tmadlener: This is now part of the exception message and there is the possibility to check for consistency.

src/ROOTNTupleWriter.cc
```diff
@@ -64,22 +64,50 @@ void ROOTNTupleWriter::writeFrame(const podio::Frame& frame, const std::string&
...
+
+    for (const auto& [name, coll] : collections) {
+      catInfo.id.emplace_back(coll->getID());
+      catInfo.type.emplace_back(coll->getTypeName());
+      catInfo.isSubsetCollection.emplace_back(coll->isSubsetCollection());
+      catInfo.schemaVersion.emplace_back(coll->getSchemaVersion());
+    }
+  } else {
+    if (!root_utils::checkConsistentColls(catInfo.name, collsToWrite)) {
+      throw std::runtime_error("Trying to write category '" + category + "' with inconsistent collection content." +
```
andresailer, line 105:
```suggestion
      throw std::runtime_error("Trying to write category '" + category + "' with inconsistent collection content. " +
```
Like in the other place

src/rootUtils.h
```diff
@@ -272,6 +273,75 @@ inline std::vector<std::string> sortAlphabeticaly(std::vector<std::string> strin
...
+  }
+
+  // Since we are guaranteed to have unique names here, we can just look for
+  // collisions brute force, which seems to be quickest approach for vector
+  // sizes we typically have here (few hundred)
+  for (const auto& id : candidateColls) {
+    if (std::find(existingColls.begin(), existingColls.end(), id) == existingColls.end()) {
+      return false;
+    }
+  }
```
jmcarcell, line 295:
```suggestion
  // Since we are guaranteed to have unique names here, we can just look for
  // collisions brute force, which seems to be quickest approach for vector
  // sizes we typically have here (few hundred)
  for (const auto& id : candidateColls) {
    if (!std::binary_search(existingColls.begin(), existingColls.end(), id, [](const auto& lhs, const auto& rhs) { std::lexicographical_compare(
        lhs.begin(), lhs.end(), rhs.begin(), rhs.end(),
        [](const auto& cl, const auto& cr) { return std::tolower(cl) < std::tolower(cr); }))) {
      return false;
    }
  }
```
We can use that `existingColls` is ordered, and maybe make a helper function for the comparison lambda, looks like it should be faster.

src/rootUtils.h
```diff
@@ -272,6 +273,75 @@ inline std::vector<std::string> sortAlphabeticaly(std::vector<std::string> strin
...
+    sstr << sep << name;
+    sep = ",";
+  }
+  sep = "";
+  sstr << "], superfluous: [";
+  for (const auto& name : onlyCands) {
+    sstr << sep << name;
+    sep = ",";
+  }
+  sstr << "]";
```
jmcarcell, line 340:
```suggestion
  std::stringstream sstr;
  if (onlyExisting.size()) {
    sstr << "missing: [";
    std::string sep = "";
    for (const auto& name : onlyExisting) {
      sstr << sep << name;
      sep = ",";
    }
    sstr << "], ";
  }
  
  sep = "";
  if (onlyCands.size()) {
    sstr << "superfluous: [";
    for (const auto& name : onlyCands) {
      sstr << sep << name;
      sep = ",";
    }
    sstr << "]";
  }
```
Say only when needed?

src/rootUtils.h
tmadlener: I added the `binary_search` version to benchmarks[^1] I did before to see if a more involved approach works better for us. It looks like a `binary_search` approach **without lower-casing things** is the best (depending on the number of collections). On the other hand, in most cases this check will probably exit simply because the two vectors have unequal size.

**Results with one vector orderd as we do in the ROOTFrameWriter** (and the second one unordered). The numbers are the sizes of the vectors. `CompareLinear` is the current implementation, `CompareBinary` and `CompareBinaryLowerCase` use `binary_search`. The former one uses `binary_search(existingColls.begin(), existingColls.end(), id)`, the second one the proposal from above.
```
---------------------------------------------------------------------
Benchmark                           Time             CPU   Iterations
---------------------------------------------------------------------
CompareLinear/2                  11.8 ns         11.8 ns     57782514
CompareLinear/8                  50.6 ns         50.5 ns     10000000
CompareLinear/64                 1153 ns         1151 ns       603770
CompareLinear/512               61044 ns        60996 ns        11362
CompareBinary/2                  27.8 ns         27.8 ns     24459705
CompareBinary/8                   142 ns          141 ns      4901529
CompareBinary/64                  525 ns          525 ns      1375137
CompareBinary/512                 706 ns          705 ns       983974
CompareBinaryLowerCase/2          251 ns          251 ns      2779374
CompareBinaryLowerCase/8         1646 ns         1644 ns       422281
CompareBinaryLowerCase/64       19288 ns        19274 ns        36780
CompareBinaryLowerCase/512     173434 ns       173318 ns         4026
```

[^1]: Basically what the benchmark does is to select `N` collection names randomly from all the collection names that we collected during #412. This is done twice, in order to have vectors to compare. In this way the cheap check for the sizes is effectively always skipped and depending on `N` the actual loop is executed at least a few times. The biggest `N=512` in this case represents the worst case for the runtime as it will do the full N^2 of work. This is not entirely our use case, but it should represent sort of the worst case scenario.

tmadlener: After some further checking it turns out that `binary_search` without taking into account that `existingColls` are lexicographically sorted case insensitve doesn't work because the pre-conditions of `binary_search` are broken. Hence, it looks like `std::find` and linear searching is still the way to go.

src/rootUtils.h
tmadlener: Done

src/rootUtils.h
tmadlener: Just for documentation of our short internal discussion. Making the `binary_search` approach exit early in case the strings are of unequal sizes, speeds that up quite a bit and it becomes comparable in speed at 64 elements already and scales much better:

```
-----------------------------------------------------------------------------
Benchmark                                   Time             CPU   Iterations
-----------------------------------------------------------------------------
CompareLinear/2                          11.3 ns         11.3 ns     58053921
CompareLinear/8                          48.3 ns         48.3 ns     14333123
CompareLinear/64                         1130 ns         1130 ns       618685
CompareLinear/512                       60220 ns        60216 ns        11686
CompareBinaryLowerCaseShortCut/2          150 ns          150 ns      4636032
CompareBinaryLowerCaseShortCut/8          498 ns          498 ns      1373998
CompareBinaryLowerCaseShortCut/64        1159 ns         1159 ns       599214
CompareBinaryLowerCaseShortCut/512       6149 ns         6149 ns       111620
````

Given that our happy path will always have to go through the complete vector to see whether the contents are the same, I have switched to this approach and also made sure that the collection names are sorted accordingly for the RNTuple writer.

PR #514, opened 2023-11-10
[WIP] Better memory model for user facing objects

### LINKED ISSUE #174: heap-use-after-free in ExampleCluster

### LINKED ISSUE #492: AddressSanitizer: heap-use-after-free in object destructor
- OS version: Linux or macOS
 - Compiler version: gcc version 11.3.0
 - PODIO version: main or older releases
 - Reproduced by:

```
CXXFLAGS=-fsanitize=address LDFLAGS=-fsanitize=address cmake -DCMAKE_BUILD_TYPE=Debug -S . -B build
cmake --build build
build/tests/unittests/unittest_podio
```
 - Input: none
 - Output:
 ```output
Randomness seeded to: 3728155408
=================================================================
==3061454==ERROR: AddressSanitizer: heap-use-after-free on address 0x606000020068 at pc 0x7feda4d2a6ec bp 0x7ffcfea450a0 sp 0x7ffcfea45098
READ of size 4 at 0x606000020068 thread T0
    #0 0x7feda4d2a6eb in podio::ObjBase::release() ./include/podio/ObjBase.h:27
    #1 0x7feda4d2a6eb in ExampleCluster::~ExampleCluster() ./tests/src/ExampleCluster.cc:48
    #2 0x7feda4d8ee13 in ExampleWithOneRelationObj::~ExampleWithOneRelationObj() ./tests/src/ExampleWithOneRelationObj.cc:23
    #3 0x7feda4d8ee48 in ExampleWithOneRelationObj::~ExampleWithOneRelationObj() ./tests/src/ExampleWithOneRelationObj.cc:24
    #4 0x7feda4d8fab9 in ExampleWithOneRelationCollectionData::clear(bool) ./tests/src/ExampleWithOneRelationCollectionData.cc:50
    #5 0x7feda4d81a2c in ExampleWithOneRelationCollection::~ExampleWithOneRelationCollection() ./tests/src/ExampleWithOneRelationCollection.cc:40
    #6 0x7feda4d81a88 in ExampleWithOneRelationCollection::~ExampleWithOneRelationCollection() ./tests/src/ExampleWithOneRelationCollection.cc:41
    #7 0x7feda5a56722 in podio::EventStore::~EventStore() ./src/EventStore.cc:14
    #8 0x4b6384 in CATCH2_INTERNAL_TEST_6 ./tests/unittests/unittest.cpp:105
    #9 0x557269 in Catch::RunContext::invokeActiveTestCase() (./build/tests/unittests/unittest_podio+0x557269)
    #10 0x5576bc in Catch::RunContext::runCurrentTest(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&) (./build/tests/unittests/unittest_podio+0x5576bc)
    #11 0x557b11 in Catch::RunContext::runTest(Catch::TestCaseHandle const&) (./build/tests/unittests/unittest_podio+0x557b11)
    #12 0x560bd4 in Catch::Session::runInternal() (./build/tests/unittests/unittest_podio+0x560bd4)
    #13 0x560edb in Catch::Session::run() (./build/tests/unittests/unittest_podio+0x560edb)
    #14 0x43d3eb in main (./build/tests/unittests/unittest_podio+0x43d3eb)
    #15 0x7feda342924d in __libc_start_call_main (/nix/store/bzd91shky9j9d43girrrj6vmqlw7x9m8-glibc-2.35-163/lib/libc.so.6+0x2924d)
    #16 0x7feda3429308 in __libc_start_main_impl (/nix/store/bzd91shky9j9d43girrrj6vmqlw7x9m8-glibc-2.35-163/lib/libc.so.6+0x29308)
    #17 0x43d484 in _start (./build/tests/unittests/unittest_podio+0x43d484)

0x606000020068 is located 8 bytes inside of 64-byte region [0x606000020060,0x6060000200a0)
freed by thread T0 here:
    #0 0x7feda50b3db7 in operator delete(void*, unsigned long) (/nix/store/4v2bk6almk03mfnz4122dfz8vcxynvs3-gcc-11.3.0-lib/lib/libasan.so.6+0xb3db7)
    #1 0x7feda4d3bf39 in ExampleClusterCollectionData::clear(bool) ./tests/src/ExampleClusterCollectionData.cc:74
    #2 0x61500000287f  (<unknown module>)

previously allocated by thread T0 here:
    #0 0x7feda50b2f37 in operator new(unsigned long) (/nix/store/4v2bk6almk03mfnz4122dfz8vcxynvs3-gcc-11.3.0-lib/lib/libasan.so.6+0xb2f37)
    #1 0x7feda4d25c79 in ExampleClusterCollection::create() ./tests/src/ExampleClusterCollection.cc:82
    #2 0x7ffcfea4581f  ([stack]+0x1d81f)
    #3 0x7ffcfea461bf  ([stack]+0x1e1bf)

SUMMARY: AddressSanitizer: heap-use-after-free ./include/podio/ObjBase.h:27 in podio::ObjBase::release()
Shadow bytes around the buggy address:
  0x0c0c7fffbfb0: fd fd fd fd fa fa fa fa fd fd fd fd fd fd fd fd
  0x0c0c7fffbfc0: fa fa fa fa fd fd fd fd fd fd fd fd fa fa fa fa
  0x0c0c7fffbfd0: fd fd fd fd fd fd fd fd fa fa fa fa fd fd fd fd
  0x0c0c7fffbfe0: fd fd fd fd fa fa fa fa fd fd fd fd fd fd fd fd
  0x0c0c7fffbff0: fa fa fa fa fd fd fd fd fd fd fd fd fa fa fa fa
=>0x0c0c7fffc000: fd fd fd fd fd fd fd fd fa fa fa fa fd[fd]fd fd
  0x0c0c7fffc010: fd fd fd fd fa fa fa fa fd fd fd fd fd fd fd fd
  0x0c0c7fffc020: fa fa fa fa fd fd fd fd fd fd fd fd fa fa fa fa
  0x0c0c7fffc030: fd fd fd fd fd fd fd fd fa fa fa fa fd fd fd fd
  0x0c0c7fffc040: fd fd fd fd fa fa fa fa fd fd fd fd fd fd fd fd
  0x0c0c7fffc050: fa fa fa fa fd fd fd fd fd fd fd fd fa fa fa fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
  Shadow gap:              cc
==PID==ABORTING
```
 - Goal: resolve sporadic crashes in https://github.com/eic/EICrecon/issues/949

It looks like there are some issues with lifetimes (~~R in Rust stands for Rewrite~~) when Object tries to update ref counter in the destructor:
https://github.com/AIDASoft/podio/blob/babec745c12727cd5da590abbe1e3cf1ab8743ad/python/templates/macros/implementations.jinja2#L35-L39

cc @wdconinc @nathanwbrei

### PR DESCRIPTION
**Pull request by tmadlener**
BEGINRELEASENOTES
- ...

ENDRELEASENOTES

This is a resurrected version of #220 with some minor modifications. I will summarize things again here, so that things are in one place and easier to follow. Similar to that one this should address the heap-use-after-free issues
- Fixes #174 
- Fixes #492 
- Supersedes #221 (see below for why)

The main thing this PR introduces is the `MaybeSharedPtr`, a smart pointer, that can relinquish ownership of the managed object while still being safely destructible even if the originally managed object has been destroyed. It achieves this by having a simple control block that tracks the lifetime of the `MaybeSharedPtr` and optionally that of the managed object. Ownership of the managed pointer can only move from the `MaybeSharedPtr` to something else, but never the other way around. It avoids the heap-use-after-free issues by keeping the control block alive until all references to it are destroyed. Previously this control block lived inside the managed object, so that the potentially separate lifetimes caused issues.

It is only used for the `Mutable` user facing objects. The main reason is to avoid unnecessary overhead for the immutable default classes (the size of the `Mutable` handles doubles to effectively two pointers). Additionally, the assumption is that immutable default classes are mainly obtained from collections, which act as their managers. To this end this PR also removes any constructor of the default (immutable) classes that allow to create a new handle. Instead, only the copy and other necessary constructors starting from an existing object are present. Additionally, a new `makeEmpty` function is introduced to make it possible to still create "empty handles" for handling unpersisted relations (without having to expose a public constructor from an `Obj*`). These object will always return `false` to a call to `isAvailable()`.

### Potential issues
- This has the potential to introduce sort of subtle lifetime issues, especially if users fully lean into the *value-semantics* promise, e.g. in the following case a dangling reference is created. Once the `mcP` goes out of scope it will destroy its `Obj` and the element in the vector is left with a handle that points to an already destroyed object, resulting again in a heap-use-after-free.
```cpp
std::vector<MCParticle> particles;
{
  auto mcP = MutableMCParticle{};
  clusters.push_back(mcP);
}
auto p = particles[0];
```
- The removal of the default constructor from the immutable default objects makes the `map::insert` (and family) not compile any longer as that requires a default constructor to be available. However, I think this is a minor issue as `emplace` still works:
```cpp
std::map<int, MCParticle> particleMap;
MutableMCParticle someParticle{};
particleMap.insert(42, someParticle); // <-- does not compile
particleMap.emplace(42, someParticle); // <-- this does.
```

- [ ] Release notes
- [ ] Documentation


tmadlener:
@veprbl does this look like something that could work for EIC / Jana2? I have added the minimal reprodcuer from [the original issue](https://github.com/eic/EICrecon/issues/949#issuecomment-1742283654) as a test-case that previously triggered under builds with AddressSanitizer enabled, but now no longer does.

Do the list of potential issues so far look like something that would make it hard / impossible for you to adapt this?

veprbl:
I've tried adapting our code to this branch, and the resulting diff is quite small, however, the `makeEmpty()` doesn't allow it to be forward compatible. And, I found that even though `std::map::operator[]`, like you say, doesn't compile, the `at()` method still works. I do get a random crash in runtime:
```
       `- JSignalHandler::handle_sigsegv(int, siginfo_t*, void*) (0x7ffff7f6b7a1)
        `- libc.so.6 (0x7ffff563dbf0)
         `- edm4eic::Cluster::getHits(unsigned long) const (0x7ffff4f144cf)
          `- eicrecon::ImagingClusterReco::fit_track(std::vector<edm4eic::Cluster, std::allocator<edm4eic::Cluster> > const&) const (0x7ffff3223571)
```
Haven't debugged it, but our access pattern looks quite reasonable:
https://github.com/eic/EICrecon/blob/fdc79ac42bf27b9000e2a9c752b5b9cfc6a85ba4/src/algorithms/calorimetry/ImagingClusterReco.h#L271

tmadlener:
Thanks for checking this.

It looks a bit like your crash is because your `Cluster` comes from a `makeEmpty` and hasn't been assigned a "proper" cluster yet. I am mainly guessing this from the stacktrace and the fact, that `Cluster::getHits(unsigned)`. It is generated from these bits here:
https://github.com/AIDASoft/podio/blob/858c0ff0b841705d1b18aafd57569fcbd1beda91/python/templates/macros/implementations.jinja2#L135-L140

Effectively the implementation will try to access the internal `m_obj` which is a `nullptr` if initialized via `makeEmpty` and will hence crash.

I am not sure what is the best solution in this case here. We could also bring back the default constructor. That would fix this, and other, crashes. On the other hand it might hide subtle issues where people access objects that are default initialized, expecting something more useful.

I think we should schedule a brief meeting to discuss which approach works best for you in this case. That might be quicker and easier to do than over github issues.

tmadlener:
I have now also made the default handles ref-counted / managed via a `MaybeSharedPtr` (as also suggested in #492). This doubles their size to two pointers. However, if they are in any form obtained from a collection the control block will not be initialized (it will remain `nullptr`). This means that almost all checks for the control block will be a simple check against `nullptr`. The main reasons for doing this are:
- Keep the existing behavior (but remove the use-after-free)
- Make dangling references a bit harder to create

This will now work as expected (this has become a test case):
```cpp
std::vector<ExampleHit> hits;
{
  auto hit = MutableExampleHit{};
  hits.push_back(hit);
}
auto e = hits[0].energy();
```

The only thing that is still creating dangling references is something along the lines of
```cpp
auto hit = ExampleHit::makeEmpty();
{
  auto hits = ExampleHitCollection();
  auto mHit = hits.create();
  hit = mHit;
}
auto e = hit.energy(); // <--- heap-use-after-free
```

However, since this is pre-existing behavior (although not documented well enough probably), I would argue that this is OK.

tmadlener:
I also brought back the pre-existing constructors for the default, immutable handles. With this in place this PR should now be completely transparent in fixing the reported use-after-free issues in destructors. Any other use-after-free from using "dangling references" are still there. These are not easily fixable in the current model, but fixing that requires some really breaking changes. I have, however, tried to at least document that these dangling references / invalid handles can happen if a handle outlives its collection.

python/templates/MutableObject.h.jinja2
```diff
@@ -1,3 +1,4 @@
+#include "podio/utilities/MaybeSharedPtr.h"
```
tmadlener, line 1:
```suggestion
```

included later at the appropriate place

python/templates/Obj.cc.jinja2
```diff
@@ -1,3 +1,4 @@
+#include "podio/ObjectID.h"
```
tmadlener, line 1:
```suggestion
```
include after the auto generation comment.

tmadlener:
I also just realized that this fixes #431 because it introduces a dedicated overload for `Collection::push_back` for `Mutable` types. Hence, there is no longer a need for an implicit conversion on the python side.

tmadlener:
Thanks a lot for testing. Also good that you see similar behavior than me. In that case I would vote for merging this and revisit this in case performance (ever) becomes an issue around this.

PR #516, opened 2023-11-22
Add support for interface types

tmadlener:
BEGINRELEASENOTES
- Add a new category of types that can be generated by podio: `interfaces`. These can be used to provide a type that can be initialized from several other datatypes and offers some common functionality. These interface types can be used in `OneToOneRelation`s and in `OneToManyRelation`s.
  - `interfaces` need to provide a list of types which they interface. Other types cannot be used with them.

ENDRELEASENOTES

Draft resurrection of #215 with the major difference that this no longer uses a `std::variant` of the `Obj*` types, but rather relies on a type-erased wrapping of the default handle types. Additionally, this also no longer supports mutable interface types, as I don't think we really have a use case for them.

- [x] Release notes
- [x] Needs #514 (builds directly on top of that)
- [x] Documentation

### DIFF
--- a/python/templates/macros/collections.jinja2
+++ b/python/templates/macros/collections.jinja2
@@ -73,9 +73,10 @@ std::vector<{{ member.full_type }}> {{ class.bare_type }}Collection::{{ member.n
       }
 {% if relation.interface_types %}
       // We need the concrete collection type to assign it to an InferenceWrapper
+{% set else = joiner("else") %}
 {% for int_type in relation.interface_types %}
-      if (auto concreteColl = dynamic_cast<{{ int_type.full_type }}Collection*>(coll)) {
-        const auto tmp = (*concreteColl)[id.index];
+      {{ else() }} if (auto {{ int_type.bare_type }}Coll = dynamic_cast<{{ int_type.full_type }}Collection*>(coll)) {
+        const auto tmp = (*{{ int_type.bare_type }}Coll)[id.index];
         m_rel_{{ relation.name }}->emplace_back(tmp);
       }
 {% endfor %}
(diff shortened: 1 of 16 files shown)


tmadlener:
Julia code generation cannot really deal with interface types at the moment. I will probably just filter them out at generation time to make the tests pass here.

tmadlener:
Waiting on #527 to have a cleaner basis for adding interface types only for c++.

python/templates/macros/collections.jinja2
```diff
@@ -71,9 +71,19 @@ std::vector<{{ member.full_type }}> {{ class.bare_type }}Collection::{{ member.n
         m_rel_{{ relation.name }}->emplace_back({{ relation.full_type }}::makeEmpty());
         continue;
       }
+{% if relation.interface_types %}
+      // We need the concrete collection type to assign it to an InferenceWrapper
+{% for int_type in relation.interface_types %}
+      if (auto concreteColl = dynamic_cast<{{ int_type.full_type }}Collection*>(coll)) {
```
tmadlener, line 77: This should result in an `if`-`else if` chain, rather than just a list of `if`s.

python/templates/macros/collections.jinja2
```diff
@@ -91,8 +101,16 @@ std::vector<{{ member.full_type }}> {{ class.bare_type }}Collection::{{ member.n
         entries[i]->m_{{ relation.name }} = nullptr;
         continue;
       }
+{% if relation.interface_types %}
+{% for int_type in relation.interface_types %}
+      if (auto concreteColl = dynamic_cast<{{ int_type.full_type }}Collection*>(coll)) {
```
tmadlener, line 106: Also this should be an `if`-`else if` chain

PR #484, opened 2023-09-15
Improve import structure of podio to feel more pythonic

tmadlener:
BEGINRELEASENOTES
- Make the podio python bindings import structure "feel more pythonic"

ENDRELEASENOTES

Fixes #459 

@jmcarcell is this enough already, or do you think we can still improve on this?

PR #521, opened 2023-12-01
Do not import ROOT when using `podio-dump --help`

jmcarcell:
BEGINRELEASENOTES
- Do not import ROOT when using `podio-dump --help`, otherwise it can take a while depending on the system only to print the help.

ENDRELEASENOTES

Small quality of life improvement to make it a bit less painful when running from cvmfs.


tmadlener:
Ah right, this changed with #484 I think. Previously the `from podio import __version__` was exactly there to have a quick `podio-dump --help` and `podio-dump --version`.

I would propose to keep the current `__version__` file because that is what "normal" python packages do and simply create a second `podio_version.py` with the same contents.

tmadlener (approved):
Thanks a lot for this.

PR #522, opened 2023-12-03
Add a reader and writer interface

jmcarcell:
BEGINRELEASENOTES
- Add a reader and writer interface and tests using those
- Add a new library `podioIO` with the interface readers and writers

ENDRELEASENOTES

- [x] Needs https://github.com/AIDASoft/podio/pull/568; Clang 12 doesn't like to return a std::unique_ptr without std::move (see the last commits) but if they are added then GCC 12 complains. Locally both GCC 13 and Clang 17 like both so I guess we stick to the most natural option.


tmadlener (commented):
I haven't yet looked into all the details, but one general comment that I already have now is: How hard would it be to make an I/O independent reader interfaces? The hardest part is most likely the different return types of `readNextEntry`/`readEntry`.

I would then make the user facing interface look something like this:
```cpp
struct Reader {
   podio::Frame readNextFrame(const std::string& category);
};
```
which then internally just calls the `readNextEntry` from the specific reader and constructs the Frame from the returned FrameData. Potentially the return type should be `std::optional<podio::Frame>` to make it easier to check whether we have run out of entries, because `readNextEntry` will return a `nullptr` in that case.

include/podio/Reader.h
```diff
@@ -0,0 +1,96 @@
...
+    size_t getEntries(const std::string& name) override {
+      return m_reader->getEntries(name);
+    }
+    size_t getEvents() override {
+      return getEntries(podio::Category::Event);
+    }
+    podio::version::Version currentFileVersion() const override {
+      return m_reader->currentFileVersion();
+    }
+    T* m_reader{nullptr};
```
tmadlener, line 60: I think this needs to be either just a `T` or a `unique_ptr<T>` as currently we seem to be leaking this.

include/podio/Reader.h
```diff
@@ -0,0 +1,96 @@
...
+    podio::Frame readNextFrame(const std::string& name) override {
+      auto maybeFrame = m_reader->readNextEntry(name);
+      if (maybeFrame) {
+        return std::move(maybeFrame);
+      }
+      throw std::runtime_error("Could not read frame (reading beyond bounds?)");
+    }
+    podio::Frame readNextEvent() override {
+      return readNextFrame(podio::Category::Event);
+    }
```
tmadlener, line 39: :heart: nice

include/podio/Reader.h
```diff
@@ -0,0 +1,96 @@
...
+  podio::Frame readFrame(const std::string& name, size_t index) {
+    return m_self->readFrame(name, index);
+  }
+  podio::Frame readEvent(size_t index) {
+    return readFrame(podio::Category::Event, index);
+  }
+  size_t getEntries(const std::string& name) {
+    return m_self->getEntries(name);
+  }
+  size_t getEvents() {
```
tmadlener, line 83:
```suggestion
  size_t getNEvents() {
```

Not sure, because `Entries` is also without the `N`, so either we have it very obvious that this returns the number of events or we are inconsistent with the entries version.

include/podio/Reader.h
```diff
@@ -0,0 +1,96 @@
...
+    return getEntries(podio::Category::Event);
+  }
+  podio::version::Version currentFileVersion() const {
+    return m_self->currentFileVersion();
+  }
+};
+
+std::unique_ptr<Reader> makeReader(const std::string& filename);
+std::unique_ptr<Reader> makeReader(const std::vector<std::string>& filename);
```
tmadlener, line 92: Looking at the implementation of these it should be possible to have these return a `Reader` by value and simply throw if something is not working as expected. Currently some paths through the implementation throw, while others return a `nullptr`. 

A value would also be more in tune with the rest of podio which usually tries to have value semantics.

include/podio/Writer.h
```diff
@@ -0,0 +1,67 @@
...
+  }
+  void writeEvent(const podio::Frame& frame) {
+    return writeFrame(frame, podio::Category::Event);
+  }
+  void writeEvent(const podio::Frame& frame, const std::vector<std::string>& collections) {
+    return writeFrame(frame, podio::Category::Event, collections);
+  }
+};
+
+std::unique_ptr<Writer> makeWriter(const std::string& filename, const std::string& type = "default");
```
tmadlener, line 63: Same here as for the reader. Can we have this by value instead?

tmadlener (commented):
Can you move the `RNTuple` related renaming commit to a separate PR? It would make this a bit more "atomic" and would yield a slightly cleaner diff for the review of the interface related changes.

I have a few comments for the interface already below as well. I think it looks good in general, just a few details that could be improved.

include/podio/Reader.h
```diff
@@ -0,0 +1,96 @@
...
+public:
+  struct ReaderConcept {
+    virtual ~ReaderConcept() = default;
+
+    virtual podio::Frame readNextFrame(const std::string& name) = 0;
+    virtual podio::Frame readFrame(const std::string& name, size_t index) = 0;
+    virtual size_t getEntries(const std::string& name) = 0;
+    virtual podio::Frame readNextEvent() = 0;
+    virtual podio::Frame readEvent(size_t index) = 0;
+    virtual size_t getEvents() = 0;
```
tmadlener, line 19: I think these could be removed from the concept, as they are in any case implemented via other methods, so they could just live in the Model, or potentially even the outer `Reader` class itself(?).

tmadlener (commented):
I am not sure if we want to replicate the reader / writer interfaces on the python side exactly. In the end the whole Mixin business is there to abstract that away and we have duck typing in python, so we probably should just aim at providing (renaming?) the `makeReader` or `makeWriter` functionality in python to do the right thing.

Another advantage of that would be that the `FrameCategoryIterator` will always get some `FrameDataT` and will not have to differentiate between dealing with the reader interface or with a concrete reader.

If we replicate the reader / writer interfaces they have to go into some generic module that is independent of the backend (i.e. it would have to be moved out of `root_io` where it currently is).

include/podio/Reader.h
jmcarcell: Changed

include/podio/Reader.h
jmcarcell: Now they are on the outer class

tests/root_io/read_interface.cpp
```diff
@@ -0,0 +1,110 @@
...
+    return 1;
+  }
+#endif
+
+#if PODIO_ENABLE_SIO
+  auto readerSIO = podio::makeReader("example_frame_sio_interface.sio");
+  if (read_frames(readerSIO)) {
+    return 1;
+  }
+#endif
```
tmadlener, line 107: This should go into the `sio_io` tests directory, also for the writing. That might require shuffling some of the code in the .cpp files into headers and putting them into the top level `tests` directory.

tmadlener (commented):
I only have some test organization and nitpicks here.

include/podio/Reader.h
```diff
@@ -0,0 +1,109 @@
...
+      }
+      throw std::runtime_error("Could not read frame (reading beyond bounds?)");
+    }
+
+    podio::Frame readFrame(const std::string& name, size_t index) override {
+      auto maybeFrame = m_reader->readEntry(name, index);
+      if (maybeFrame) {
+        return maybeFrame;
+      }
+      throw std::runtime_error("Could not read frame (reading beyond bounds?)");
```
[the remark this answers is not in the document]
jmcarcell: Changed

include/podio/Reader.h
```diff
@@ -0,0 +1,109 @@
...
+    virtual podio::Frame readFrame(const std::string& name, size_t index) = 0;
+    virtual size_t getEntries(const std::string& name) = 0;
+    virtual podio::version::Version currentFileVersion() const = 0;
+    virtual std::vector<std::string_view> getAvailableCategories() const = 0;
+    virtual const std::string_view getDatamodelDefinition(const std::string& name) const = 0;
+    virtual std::vector<std::string> getAvailableDatamodels() const = 0;
+  };
+
+  template <typename T>
+  struct ReaderModel : public ReaderConcept {
```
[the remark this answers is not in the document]
jmcarcell: Changed

include/podio/Reader.h
```diff
@@ -0,0 +1,109 @@
...
+    }
+    ReaderModel(const ReaderModel&) = delete;
+    ReaderModel& operator=(const ReaderModel&) = delete;
+
+    podio::Frame readNextFrame(const std::string& name) override {
+      auto maybeFrame = m_reader->readNextEntry(name);
+      if (maybeFrame) {
+        return maybeFrame;
+      }
+      throw std::runtime_error("Could not read frame (reading beyond bounds?)");
```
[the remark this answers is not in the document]
jmcarcell: Changed

include/podio/Writer.h
```diff
@@ -0,0 +1,86 @@
...
+    virtual void writeFrame(const podio::Frame& frame, const std::string& category) = 0;
+    virtual void writeFrame(const podio::Frame& frame, const std::string& category,
+                            const std::vector<std::string>& collections) = 0;
+    virtual void writeEvent(const podio::Frame& frame) = 0;
+    virtual void writeEvent(const podio::Frame& frame, const std::vector<std::string>& collections) = 0;
+    virtual void finish() = 0;
+  };
+
+  template <typename T>
+  struct WriterModel : public WriterConcept {
```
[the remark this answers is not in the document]
jmcarcell: Added

src/Writer.cc
```diff
@@ -0,0 +1,46 @@
...
+#endif
+  }
+  if ((type == "default" && endsWith(filename, ".sio")) || lower(type) == "sio") {
+#if PODIO_ENABLE_SIO
+    return Writer{std::make_unique<SIOFrameWriter>(filename)};
+#else
+    throw std::runtime_error("SIO writer not available. Please recompile with SIO support.");
+#endif
+  }
+  throw std::runtime_error("Unknown file type");
```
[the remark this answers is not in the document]
jmcarcell: Changed

tests/root_io/read_interface.cpp
jmcarcell: I added some more general includes in the top tests directory and then the actual tests in `tests/root_io` and `tests/sio_io`. I've added a new library `podioIO` since I don't think there is a good place for the reader and writer (but then one will need to link to this one to use the interfaces). I think this should eventually replace the current instances since the interface should cover most of the use cases and you can also change backend very easily.

include/podio/Reader.h
```diff
@@ -0,0 +1,109 @@
...
+  };
+
+  template <typename T>
+  struct ReaderModel : public ReaderConcept {
+    ReaderModel(T* reader) : m_reader(reader) {
+    }
+    ReaderModel(const ReaderModel&) = delete;
+    ReaderModel& operator=(const ReaderModel&) = delete;
+
+    podio::Frame readNextFrame(const std::string& name) override {
```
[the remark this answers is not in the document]
jmcarcell: Since the result is not conclusive and this is not necessary from the syntax nor conceptual point of view (there can't be a derived class because of the class `final`) I prefer to leave the override here

tests/root_io/read_interface.cpp
tmadlener: About the new library / target. I think this should be the new default target to link against. It brings along all the correct dependencies in any case (it links publicly against `podioRootIO` and potentially also `podioSioIO`). So users should just use `podioIO` in their setups. We will have to document that somewhere. However, I don't think we have a fitting section yet to put it into. I only ever put the cmake macros on to slides so far, it seems.

src/Writer.cc
```diff
@@ -0,0 +1,46 @@
...
+    return Writer{std::make_unique<ROOTWriter>(filename)};
+  }
+  if (lower(type) == "rntuple") {
+#if PODIO_ENABLE_RNTUPLE
+    return Writer{std::make_unique<RNTupleWriter>(filename)};
+#else
+    throw std::runtime_error("ROOT RNTuple writer not available. Please recompile with ROOT RNTuple support.");
+#endif
+  }
+  if ((type == "default" && endsWith(filename, ".sio")) || lower(type) == "sio") {
```
tmadlener, line 36: I am not entirely sure whether we should check the `type` here as well. currently it would be possible to write `example.root` as an `sio` file, which would then break when trying to read it back, probably with some rather incomprehensible error.

I think it would be less error prone, if we only check the `type` for the `.root` file ending and ignore it entirely for `sio`. Otherwise we would at least also have to foresee the possibility to customize the `type` in the reader interface.

tmadlener:
I don't understand why the `read_interface_sio` test is flagged by the AddressSanitizer, claiming there is a leak of effectively the complete Frame. Since reading with sio works without leaks this would have to be introduced somewhere in the interface, but I don't see a place where that would be possible or where it would make sense. Especially since the AddressSanitizer effectively points to the Frame construction in `readEvent`.

jmcarcell:
Reading a frame with SIO is also not free from leaks and this is why it's commented out in the sanitizer tests, right? 
https://github.com/jmcarcell/podio/blob/main/tests/CTestCustom.cmake#L83-L83
Simply reading a frame with SIO:
```
SUMMARY: AddressSanitizer: 7328 byte(s) leaked in 342 allocation(s).
```
The `read_interface_sio.cpp` test:
```
SUMMARY: AddressSanitizer: 7328 byte(s) leaked in 342 allocation(s).
```
so it doesn't have anything to do with the interface. I'll just add it to the list of excluded

andresailer:
> Seems like there is still another issue with clang12, that doesn't completely like the implicit construction in the interface:

What keeps us from dropping clang 12 and moving to clang16?

tmadlener:
Nothing really, other than losing some backwards compatibility.

jmcarcell:
> Seems like there is still another issue with clang12, that doesn't completely like the implicit construction in the interface:
> 
> [AIDASoft/podio/actions/runs/8964777459/job/24618838604?pr=522#step:4:340](https://github.com/AIDASoft/podio/actions/runs/8964777459/job/24618838604?pr=522#step:4:340)

Yeah, this is what I comment above in https://github.com/AIDASoft/podio/pull/522#issuecomment-2070762263

tmadlener:
CI update coming with #597

src/Writer.cc
jmcarcell: Done

src/CMakeLists.txt
```diff
@@ -92,6 +92,8 @@ SET(root_headers
   ${PROJECT_SOURCE_DIR}/include/podio/ROOTReader.h
   ${PROJECT_SOURCE_DIR}/include/podio/ROOTLegacyReader.h
   ${PROJECT_SOURCE_DIR}/include/podio/ROOTWriter.h
+  ${PROJECT_SOURCE_DIR}/include/podio/Reader.h
+  ${PROJECT_SOURCE_DIR}/include/podio/Writer.h
```
tmadlener, line 96: I think these should go into a separate list of headers, distinct from the root ones? That should also make the selection.xml things work?

src/root_selection.xml
```diff
@@ -5,5 +5,7 @@
     <class name="podio::ROOTWriter"/>
     <class name="podio::RNTupleReader"/>
     <class name="podio::RNTupleWriter"/>
+    <class name="podio::Reader"/>
+    <class name="podio::Writer"/>
```
tmadlener, line 9: Shouldn't these go into `io_selection.xml`?

src/CMakeLists.txt
```diff
@@ -135,6 +137,18 @@ if(ENABLE_SIO)
...
+  )
+
+PODIO_ADD_LIB_AND_DICT(podioIO "${core_headers}" "${io_sources}" io_selection.xml)
+target_link_libraries(podioIO PUBLIC podio::podio podio::podioRootIO)
+if(ENABLE_SIO)
+  target_link_libraries(podioIO PUBLIC podio::podioSioIO)
+endif()
+
 # --- Install everything
 install(TARGETS podio podioDict podioRootIO podioRootIODict ${INSTALL_LIBRARIES}
```
tmadlener, line 153:
```suggestion
install(TARGETS podio podioDict podioRootIO podioRootIODict podioIO ${INSTALL_LIBRARIES}
```

I don't think I saw it anywhere else that would install it(?).

src/io_selection.xml
```diff
@@ -0,0 +1,4 @@
+<lcgdict>
+  <selection>
+  </selection>
```
tmadlener, line 3: I suppose this is here mainly to make the whole CMake functions work? I think this should be populated with the `Reader` and `Writer` classes, which should (hopefully) be possible with the changes suggested in the other comments.

tmadlener (commented):
I have found a few more smaller things that I have missed before. They are related to the organization into libraries / targets and root dictionaries, which currently seems to mix a few things with the root specific ones and I think it should be possible to split that more consistently, but maybe there is a reason you have done it as it is done at the moment that I am missing.

src/CMakeLists.txt
jmcarcell: Fixed

src/root_selection.xml
jmcarcell: Fixed

src/io_selection.xml
jmcarcell: Fixed

src/CMakeLists.txt
jmcarcell: Fixed, I checked the target, dictionaries and headers are being installed

tmadlener (approved):

PR #527, opened 2023-12-07
Refactor the code generation and split it into language specific generators

tmadlener:
BEGINRELEASENOTES
- Split the `ClassGenerator` into a base class (mixin) and two specific c++ and julia code generators that only deal with their language specific needs.
  - Instantiate and configure the correct reader in the `podio_class_generator.py` main script depending on the desired language.
- Slightly cleanup the `MemberVariable` to declutter its `__init__` method a bit.

ENDRELEASENOTES

This resolves the main design issue mentioned in https://github.com/AIDASoft/podio/pull/473#pullrequestreview-1747913649

Apart from implementing the basics for the mixin this is mainly just shuffling code around.


python/podio_gen/generator_base.py
```diff
@@ -0,0 +1,229 @@
...
+  The following members are initialized and accessible from inheriting classes
+  - yamlfile (the path to the yamlfile)
+  - install_dir (top level directory into which the code should be generated)
+  - package_name (the name of the package)
+  - verbose (whether to print some information about the code gen process)
+  - dryrun (whether to actually generate the datamodel or to only run the
+    processing without filling the contents)
+  - upstream_edm (an optional upstream datamodel)
+  - datamodel (the current datamodel read from the yamlfile)
+  - get_syntax (whether to use get synatx or not)
```
tmadlener, line 58:
```suggestion
  - env (the jinja2 environment for generating code)
  - get_syntax (whether to use get synatx or not)
```

tmadlener:
@hegner does anything speak against merging this? Should make experimenting with julia a bit easier.

tmadlener:
Merging this to let the nightlies pick it up before a new tag for the next LCG release

PR #528, opened 2023-12-07
Use black to format python sources

tmadlener:
BEGINRELEASENOTES
- Use `black` to format all python source files
- Add a pre-commit hook for running black in CI

ENDRELEASENOTES

- [x] Includes #527 
- [x] Add CMake target to make it possible to run it as part of the build -> #544 

**IMPORTANT: Do not squash-merge this because then we cannot add the format only commit to `.git-ignore-blame-refs`**

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,5 +1,5 @@
 [flake8]
-max-line-length = 88
+max-line-length = 99
 extend-ignore = E203
 
 per-file-ignores =
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,6 @@
+[flake8]
+max-line-length = 88
+extend-ignore = E203
+
+per-file-ignores =
+  python/podio_gen/test_MemberParser.py: E501
--- a/.github/scripts/pylint.rc
+++ b/.github/scripts/pylint.rc
@@ -144,7 +144,7 @@ property-classes=abc.abstractproperty
 [FORMAT]
 
 # Maximum number of characters on a single line.
-max-line-length=120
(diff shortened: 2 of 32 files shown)


tmadlener:
Keep this open until #516 is merged to not have to many conflicts between the two.

jmcarcell:
Sorry for the late comment but now that we are getting at this there is a similar tool called ruff: https://github.com/astral-sh/ruff
ruff was initially a linter but after some time is also a formatter with support for black and others: https://astral.sh/blog/the-ruff-formatter. One of the reasons to use it would be the speed although we don't have repos large enough to benefit from that a lot, probably. Another reason is the fact that it is a tool that can do several things. This is what I use as linter in my editor and tells me issues without having to run other things and can also format, in this case being fast and reacting to changes quickly is important. I added it to the key4hep stack not too long ago with the intention to use it in key4hep one day...

Maybe I'm biased but I hear people using ruff a lot. For comparison, ruff is 1.5 years old and has 22k stars on github, black is almost 6 years old and has 36k (and part of the PSF) so it's a quite popular tool.

tmadlener:
If we can make `ruff` format to the same as `black` in this configuration, we can always switch the linters later, because that would presumably also replace `pylint` and `flake8`, IIUC.

tmadlener:
Merging this now. From a quick test it looks like `ruff` can be made to reproduce this formatting with `black`.

tmadlener:
Failing CI, due to #545

PR #536, opened 2024-01-08
Add a getAvailableCollections method in python

jmcarcell:
BEGINRELEASENOTES
- Add a `getAvailableCollections` method in python that does the same thing as in C++

ENDRELEASENOTES

Currently `frame.collections` is the way of getting the collections. I'd prefer the C++ method and the python one to have the same name so it's easier to remember.

I'm also in favor of removing the current `.collections` which is probably only used in podio (in podio-dump).

### DIFF
--- a/python/podio/frame.py
+++ b/python/podio/frame.py
@@ -1,9 +1,8 @@
 #!/usr/bin/env python3
 """Module for the python bindings of the podio::Frame"""
 
-# pylint: disable-next=import-error # gbl is a dynamic module from cppyy
-import cppyy
 import sys
+import cppyy
 
 import ROOT
 
--- a/python/podio/frame.py
+++ b/python/podio/frame.py
@@ -125,7 +125,7 @@ class Frame:
     Returns:
         tuple(str): The names of the available collections from this Frame.
     """
-    warnings.warn('WARNING: collections is deprecated use getAvailableCollections()'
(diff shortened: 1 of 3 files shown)


tmadlener (commented):
No objections from my side to add this (and potentially deprecate the `collections` property and removing it after the next tag. Unless, you want to keep that for the "pythonic interface"?

jmcarcell:
`.collections` now prints a warning and I changed it to `.getAvailableCollections()` where I found it. I'm fine with removing it, it's one less thing to remember or check when it's different in C++ and python.

python/podio/frame.py
```diff
@@ -110,14 +110,24 @@ def __init__(self, data=None):
...
   @property
   def collections(self):
     """Get the currently available collection (names) from this Frame.
 
     Returns:
         tuple(str): The names of the available collections from this Frame.
     """
-    return tuple(str(s) for s in self._frame.getAvailableCollections())
+    print('WARNING: collections is deprecated, use getAvailableCollections() (like in C++) instead',
+          file=sys.stderr)
```
tmadlener, line 129: Can we make this a `warnings.warn("text", FutureWarning)`? That would give a bit more context as to where the warning is coming from

tmadlener (approved):

PR #539, opened 2024-01-16
Fix crash when a writer is in the global namespace

jmcarcell:
BEGINRELEASENOTES
- Fix crash when a writer is in the global namespace by adding a class that will manage all the writers and finish them before exiting

ENDRELEASENOTES

The following code crashes (either `Writer` or `RNTupleWriter`):

``` python
import podio
writer = podio.root_io.RNTupleWriter('rntuple.root')
```

The workaround is to add a `del writer` at the end. This happens because the writers do something when they are destructed but podio and/or ROOT may get destructed before. Error message:

``` c++
Error in <TKey::Create>: Cannot allocate 658 bytes for ID =  Title =
SysError in <TFile::Seek>: cannot seek to position 0 in file rntuple.root, retpos=-1 Bad file descriptor
Fatal: !rv violated at line 1152 of `/root/tree/ntuple/v7/src/RMiniFile.cxx'
aborting
```

### DIFF
--- a/python/podio/base_writer.py
+++ b/python/podio/base_writer.py
@@ -2,6 +2,31 @@
 """Python module for defining the basic writer interface that is used by the
 backend specific bindings"""
 
+import atexit
+
+
+class AllWriters:
+  """Class to manage all writers in the program
+     so that they can be properly finished at the end of the program
+  """
+  writers = []
+
+  def add(self, writer):
+    """Add a writer to the list of managed writers"""
+    self.writers.append(writer)
+
+  def finish(self):
+    """Finish all managed writers"""
+    for writer in self.writers:
+      try:
+        writer._writer.finish()  # pylint: disable=protected-access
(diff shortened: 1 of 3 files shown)


tmadlener (commented):
This does also affect all the readers, right? Could we the `_all_writers.add` into the `BaseReaderMixin` constructor?

jmcarcell:
Hm nope, at least they don't crash. I've moved the management to `base_writer.py`, in lines of code it's the same since every writer has now to do `super().__init__()`

tmadlener (approved):
Not really sure about the `AllWriters` name, but I can't think of anything better. Since this is internal in any case we should be able to change this later if we come up with something better.

jmcarcell:
`WriterManager`, `WriterDeleter` or `WriterFinisher` or even `WriterDestroyer` for maximum effect

PR #540, opened 2024-01-17
Add a check using ROOT_CXX_STANDARD

jmcarcell:
BEGINRELEASENOTES
- Add a check for the C++ standard using `ROOT_CXX_STANDARD` that was introduced yesterday in ROOT (https://github.com/root-project/root/commit/d487a42b311c5d0c7544031e3071a388c488c429)

ENDRELEASENOTES

currently nightly builds with HEAD of ROOT are not working in the LCG stacks

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -86,10 +86,18 @@ else()
 endif()
 
 # Check that root is compiled with a modern enough c++ standard
-get_target_property(ROOT_COMPILE_FEATURES ROOT::Core INTERFACE_COMPILE_FEATURES)
-if (NOT "cxx_std_17" IN_LIST ROOT_COMPILE_FEATURES AND NOT "cxx_std_20" IN_LIST ROOT_COMPILE_FEATURES)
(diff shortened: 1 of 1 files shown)


tmadlener:
Do we already have a check somewhere that makes sure that building podio with c++20 also requires root to have c++20?

tmadlener:
Does this fixe this error? https://github.com/AIDASoft/podio/actions/runs/7555260579/job/20569847736?pr=383#step:4:170

jmcarcell:
I added a check (prints a warning) when the C++ standard is different in ROOT from the required. It could be an error but it was somewhat useful to build some packages with C++20 when we didn't have ROOT with C++20 to see if they failed or not; anyway mismatches can't happen in spack so in user builds there would be a visible warning.

> Does this fixe this error? [AIDASoft/podio/actions/runs/7555260579/job/20569847736?pr=383#step:4:170](https://github.com/AIDASoft/podio/actions/runs/7555260579/job/20569847736?pr=383#step:4:170)

Yes

tmadlener (approved):
Thanks for the clarification. Merging this to unblock a few of the other PRs that are essentially ready.

PR #541, opened 2024-01-17
Add tool to transform between TTrees and RNTuples

jmcarcell:
BEGINRELEASENOTES
- Add tool to transform between TTrees and RNTuples

ENDRELEASENOTES

Installed only when compiling with RNTuple enabled


tmadlener:
Can you rebase this to pick up #540 for CI?

tmadlener (approved):
Looks good and simple enough to me. It also looks like this should copy over everything and just recreate metadata on the fly as necessary. I am not sure whether we need a dedicated test for the whole thing, as all the pieces are in principle already tested on their own. Unless, there are any concerns, I would merge this as is later today.

jmcarcell:
Yeah they are tested independently, at least I used this in the two directions and it worked. When there is an interface maybe it can be changed so there is automatic detection of the direction but it's not that important

tmadlener:
As far as I can see this can handle #543 transparently, right?

jmcarcell:
Yes, in python we already had that name, so another reason to change it from the original one

PR #543, opened 2024-01-22
Change ROOTNTuple{Reader,Writer} to RNTuple{Reader,Writer}

jmcarcell:
BEGINRELEASENOTES
- Change ROOTNTuple{Reader,Writer} to RNTuple{Reader,Writer}

ENDRELEASENOTES

So it's clear it's RNTuple (ROOT also has [NTuples](https://root.cern.ch/doc/master/classTNtuple.html). An alternative would be removing ROOT and having simply RNTuple{Reader, Writer}, I guess when there is an interface it won't matter much.

### DIFF
--- a/include/podio/GenericParameters.h
+++ b/include/podio/GenericParameters.h
@@ -20,8 +20,8 @@ using version_type = uint32_t; // from sio/definitions
 
 #if PODIO_ENABLE_RNTUPLE
 namespace podio {
-class ROOTRNTupleReader;
-class ROOTRNTupleWriter;
+class RNTupleReader;
+class RNTupleWriter;
 } // namespace podio
 #endif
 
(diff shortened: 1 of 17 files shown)


tmadlener:
I actually like `RNTuple{Reader,Writer}` a bit more then with `ROOT`. I suppose it is sufficiently clear that we talk about the RNTuple from ROOT. However, I am not sure how much work that would create for #522 in the end, so if you want to avoid that I am also happy just keeping `ROOTRNTuple`.

tmadlener (approved):
Thanks. Looks good to me.

Let's clarify during the meetings tomorrow and Wednesday, whether we need some sort of backward compatibility for a migration period, or whether we can just press the button as it is.

jmcarcell:
It should be quite safe to merge. Probably I'm the only one who has been using these writers and readers...

PR #548, opened 2024-01-30
Remove unnecessary usage of TClass from RNTupleReader

tmadlener:
BEGINRELEASENOTES
- Remove an unnecessary usage of the `TClass` machinery inside the `RNTupleReader` as all the necessary information is also available from metadata that we carry around in any case.

ENDRELEASENOTES

### DIFF
--- a/src/RNTupleReader.cc
+++ b/src/RNTupleReader.cc
@@ -7,8 +7,8 @@
 #include "podio/GenericParameters.h"
 #include "rootUtils.h"
 
-#include "TClass.h"
 #include <ROOT/RError.hxx>
+
 #include <memory>
 
 namespace podio {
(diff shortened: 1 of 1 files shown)


src/RNTupleReader.cc
```diff
@@ -141,15 +141,10 @@ std::unique_ptr<ROOTFrameData> RNTupleReader::readEntry(const std::string& categ
   auto dentry = m_readers[category][0]->GetModel()->GetDefaultEntry();
 
   for (size_t i = 0; i < m_collectionInfo[category].id.size(); ++i) {
-    const auto collectionClass = TClass::GetClass(m_collectionInfo[category].type[i].c_str());
-
-    auto collection =
-        std::unique_ptr<podio::CollectionBase>(static_cast<podio::CollectionBase*>(collectionClass->New()));
```
jmcarcell, line 147: Was this line problematic?

jmcarcell (approved):

src/RNTupleReader.cc
tmadlener: No, not really per se. It has just become unnecessary with the introduction of our own buffer factory. Now it was only used to get the type name basically, which we already have available otherwise.

Issue #551, opened 2024-02-05
Inconsistent reference access for Mutable types


tmadlener:
The generated interface for our mutable types allows to access component members by reference, but not for primitive types, i.e.

```yaml
components:
  SimpleStruct:
    Members:
      -  int i // integer

datatypes:
  SimpleDatatype:
    Members:
      - int j // integer
      - SimpleStruct s // component
```

will generate to (assuming getter syntax is turned on)

```cpp
class MutableSimpleDataType {
  // many other things
  // ...
  
  int getJ() const;
  const  SimpleStruct& getS() const;

  // this mutable accessor only exists for non-builtin types
  SimpleStruct& s();
};
```

This is inconsistent in two ways with the rest of the generated code:
- the mutable accessors do not respect the `getSyntax` setting
- the mutable accessors do not expose their primitive members by non-const reference

After some discussion with @paulgessinger (who actually found this), I think the easiest solution is to
- deprecate the existing mutable reference accessors
- introduce new mutable accessors for **all** members that respect the `getSyntax` option (i.e. effectively adding the non-const overloads for our mutable classes).

This doesn't affect const-correctness of generated EDMs and will make the Mutable handles more consistent (and a bit more useful).

PR #553, opened 2024-02-05
Generate mutable reference getters for all members

tmadlener:
BEGINRELEASENOTES
- Generate mutable reference getters for all members on the `Mutable` types
  - Previously this was only generated for non-builtin (i.e. non-trivial) members. 
  - Additionally, deprecate the API that does not respect the `getSyntax` option.
  - Fixes https://github.com/AIDASoft/podio/issues/551

ENDRELEASENOTES

- [x] tests

@paulgessinger this should be more or less what we discussed today. A minor drawback that I just realized is that for the `const` versions of these getters we return by value for builtin types rather than by `const&` (mainly premature optimization I would say). In any case, it shouldn't really matter for you, and I think I am OK with that small inconsistency.

### DIFF
--- a/python/templates/macros/declarations.jinja2
+++ b/python/templates/macros/declarations.jinja2
@@ -32,8 +32,11 @@
 {% if member.is_array %}
   void {{ member.setter_name(get_syntax) }}(size_t i, {{ member.array_type }} value);
 {% endif %}
-{% if not member.is_builtin %}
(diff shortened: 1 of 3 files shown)


hegner:
@tmadlener - add the tests and I can accept :-)

tmadlener:
Done.

PR #557, opened 2024-02-12
Remove deprecated Frame functionality for v1.0

tmadlener:
BEGINRELEASENOTES
- Remove deprecated `collections` property from Frame python bindings
- Remove deprecated `getParametersForWrite` method from Frame interface

ENDRELEASENOTES


hegner:
In dev3 the inner namespace `Detail` was dropped for many of the items - which leads to the failing integration tests here. I think that should be fixed in a separate PR.

hegner (approved):

PR #561, opened 2024-02-21
feat: podio-vis --upstream-edm and some tests

wdconinc:
This adds some basic support for extension data models into podio-vis.

Also adds three tests for podio-vis to make sure it doesn't fail. May require graphviz python to be installed in the CI environment.

BEGINRELEASENOTES
- support podio-vis for extension data models

ENDRELEASENOTES

### DIFF
--- a/python/podio/__init__.py
+++ b/python/podio/__init__.py
@@ -1,4 +1,5 @@
 """Python module for the podio EDM toolkit and its utilities"""
+
 from .__version__ import __version__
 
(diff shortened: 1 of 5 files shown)


tmadlener:
Key4hep nightlies should pick up the py-graphviz dependency soon-ish. I would wait until then and run the key4hep workflow once that is in. Unless, you need this very urgently.

wdconinc:
Not urgently needed; we are also propagating py-graphviz into our environments still (FYI @bspage912).

PR #564, opened 2024-02-22
Fix: mutable to immutable object conversion with python bindings

m-fila:
BEGINRELEASENOTES
- Fix mutable to immutable object conversion with python bindings

ENDRELEASENOTES

As reported in key4hep/EDM4hep#270 the python bindings were unable to implicitly convert mutable to immutable objects with the converting constructor. An alternative approach with converting constructor instead of conversion function is viable.

Converting constructor was added and conversion function removed to avoid ambiguity in precedence.

I'm not sure if the test is placed in a correct place

Closes key4hep/EDM4hep#270

### DIFF
--- a/python/templates/Object.h.jinja2
+++ b/python/templates/Object.h.jinja2
@@ -43,6 +43,8 @@ public:
   using collection_type = {{ class.bare_type }}Collection;
 
 {{ macros.constructors_destructors(class.bare_type, Members) }}
+  /// converting constructor from mutable object
+  {{ class.bare_type }}(const Mutable{{ class.bare_type }}& other);
 
   static {{ class.bare_type }} makeEmpty();
 
--- a/tests/unittests/unittests.py
+++ b/tests/unittests/unittests.py
@@ -1,0 +1,43 @@
+#!/usr/bin/env python3
+"""cppyy python binding tests"""
+
+import unittest
+import ROOT
+
+res = ROOT.gSystem.Load("libTestDataModel.so")
+if res < 0:
+    raise RuntimeError("Failed to load libTestDataModel.so")
+
+from ROOT import MutableExampleMC  # pylint: disable=wrong-import-position # noqa: E402
+from ROOT import ExampleMCCollection  # pylint: disable=wrong-import-position # noqa: E402
+
+
+class ObjectConversionsTest(unittest.TestCase):
+    """Object conversion binding tests"""
+
+    def test_conversion_mutable_to_immutable(self):
+        ROOT.gInterpreter.Declare(
+            """
+            void test_accepts_immutable(ExampleMC) {}
+            """
+        )
+        accepts_immutable = ROOT.test_accepts_immutable
+        mutable_obj = MutableExampleMC()
+        accepts_immutable(mutable_obj)
+
+
+class OneToManyRelationsTest(unittest.TestCase):
+    """OneToManyRelations binding tests"""
+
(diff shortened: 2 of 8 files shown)


tests/unittests/unittests.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python3
```
tmadlener, line 1: I would put this into `python/podio/test_CodeGen.py` (or something similar that get's automatically picked up by the `python -m unittest discover` we run in there). That should also automatically fix the issues with the test environment which makes some tests fail at the moment.

python/templates/Object.h.jinja2
```diff
@@ -43,6 +43,8 @@ public:
   using collection_type = {{ class.bare_type }}Collection;
 
 {{ macros.constructors_destructors(class.bare_type, Members) }}
+  /// converting constructor from mutable object
+  {{ class.bare_type }}(const Mutable{{ class.bare_type }}& other);
```
tmadlener, line 47: If we mark this `explicit` things do not work any longer, right? Because then we lose the implicit conversion that we rely on in many places, I suppose.

m-fila: Exactly, plenty of 'no known conversion' errors

tmadlener:
It looks like having a converting constructor instead of a conversion operator makes clang-tidy a bit more sensitive: https://github.com/AIDASoft/podio/actions/runs/8016200074/job/21897701922?pr=564#step:4:450

m-fila:
wlav/cppyy#218 explains inefficiencies of implicit casting in cppyy.  
I tried a naive benchmark:

```python
from timeit import timeit
from ROOT import MutableExampleMC
from cppyy.ll import static_cast
 
def bench_cast():
    parent = MutableExampleMC()
    daughter = MutableExampleMC()
    daughter.addparents(static_cast['ExampleMC'](parent))
 
def bench_no_cast():
    parent = MutableExampleMC()
    daughter = MutableExampleMC()
    daughter.addparents(parent)

def report(result, reps, name):
    print(f'Average time: {result/reps:.1E} seconds\t({name})')

if __name__ == '__main__':
    reps= 100000
    result = timeit('bench_cast()', setup='from __main__ import bench_cast; bench_cast()', number=reps)
    report(result, reps, 'explicit cast')
    result = timeit('bench_no_cast()', setup='from __main__ import bench_no_cast; bench_no_cast()', number=reps)
    report(result, reps, 'implicit cast')
```

With the results:

```txt
Average time: 2.7E-06 seconds   (explicit cast)
Average time: 2.2E-06 seconds   (implicit cast)
```

Not sure why implicit seems to be faster  (very little overloads to consider?).  
Regardless, I'd say there is no visible drop in performance for this case

tmadlener:
Thanks for checking some potential performance implications. That looks reasonable to me. In the end, I don't think performance is the main concern when you go to the python bindings ;)

jmcarcell:
This works fine in EDM4hep without the casts, I tested the script that is in https://github.com/key4hep/EDM4hep/pull/272. Re the benchmarks wlav mentions all the process where you have to first fail (in a try - except block probably) and then find the correct operator and then use it. So it's going to be worse than your .5 / 2.2 = 23% loss

wdconinc:
> this should be a completely transparent change as far as we are aware

Hyrum's law strikes again. Go ahead and merge. We'll consider it a bug on our end :-)

PR #568, opened 2024-03-13
Switch to more recent OS and compilers for sanitizer workflows

tmadlener:
BEGINRELEASENOTES
- Switch to gcc13 and clang16 for sanitizer workflows. Also switch to EL9 as OS to run on.

ENDRELEASENOTES


andresailer:
Need to change the standard for the sanitizer builds when using gcc13 or clang16
```diff
--DCMAKE_CXX_STANDARD=17 
+-DCMAKE_CXX_STANDARD=20 
```

Issue #458, opened 2023-07-21
Better out-of-bounds checks for collection access


tmadlener:
I feel like here going out of range may be more of a problem than in C++ (this is not related to this PR but just indexing out of range a `std::deque`):
``` python
event = Frame()
hits = ExampleHitCollection()
hits[0] # seg fault
hits[1] # <cppyy.gbl.MutableExampleHit object at 0x55dcd5d30d50>
hits[1].x() # seg fault
hits.create(0, 1, 2. 3, 4)
hits[0] # <cppyy.gbl.MutableExampleHit object at 0x55dcc4324040>
hits[1] # <cppyy.gbl.MutableExampleHit object at 0x55dcd5d572b0>
hits[2] # seg fault
```
 It's inconsistent since the behavior is undefined, could be a bit hard to debug when you go out of range and it may seem to work but then it doesn't later on. I'm not sure if there are options, not checking in C++ I think is a design choice since there is also `.at`, but in python no one is going to use that probably

_Originally posted by @jmcarcell in https://github.com/AIDASoft/podio/pull/447#discussion_r1269738805_

PR #570, opened 2024-03-18
Add pythonizations for collection subscript

m-fila:
BEGINRELEASENOTES
- Added mechanism to load cppyy pythonizations
- Added pythonization for bound-check subscript operation in collections

ENDRELEASENOTES

#458 and key4hep/EDM4hep#288 show there is some interest in tuning the cppyy generated bindings. The cppyy  has a feature for this called [pythonization](https://cppyy.readthedocs.io/en/latest/pythonizations.html) -> a callback can be registered for all the classes that fulfill some predicate (e.g. derived from `podio::CollectionBase`).

This PR contains:
- an abstraction for adding pythonizations (each pythonization implemented in a class derived from `podio.pythonizations.utils.Pythonizer`)
-  pythonization adding bound check `__getitem__` for classes derived from `podio::CollectionBase`

The mechanism of loading the pythonizations is inspired by [ROOT pythonizations](https://github.com/root-project/root/blob/master/bindings/pyroot/pythonizations/python/ROOT/_pythonization/__init__.py) with a change that de order of loading of the pythonizations is defined.

To load the pythonizations in a downstream projects, for instance for project utilizing a "edm4hep" namespace
```python
from podio.pythonizations import load_pythonizations

load_pythonizations("edm4hep")
```
alternatively a selection of pythonizations can be loaded by importing and applying them one by one

The alternative mechanism I tough about:
- [c++ callback pythonization](https://cppyy.readthedocs.io/en/latest/pythonizations.html#c-callbacks) - instead of in python the callback to cppyy could be defined in C++ using the Python API. The downsides are extra dependency for C++ code (`Python::Module`) and knowing the Python C API is required
- generating python wrappers with jinja2 - it would be rather cumbersome to add another set jinja2 templates for tuning the python bindings

### DIFF
--- a/doc/python.md
+++ b/doc/python.md
@@ -1,0 +1,56 @@
+# Python interface for data models
+
+Podio provides support for a Python interface for the generated data models. The [design choice](design.md) to create Python interface resembling the C++ interface is achieved by generating Python bindings from the C++ interface using
+[cppyy](https://cppyy.readthedocs.io/en/latest/index.html).
+
+It's important to note that cppyy loads the bindings and presents them lazily at runtime to the Python interpreter, rather than writing Python interface files. Consequently, the Python bindings have a runtime dependency on both cppyy and the data model's C++ interface.
+
+To load the Python bindings from a generated C++ model dictionary, first make sure the model's library and headers can be found in `LD_LIBRARY_PATH` and `ROOT_INCLUDE_HEADERS` respectively, then:
+
+```python
+import ROOT
+
+res = ROOT.gSystem.Load('libGeneratedModelDict.so')
+if res < 0:
+    raise RuntimeError('Failed to load libGeneratedModelDict.so')
+```
+
+For reference usage, see [Python module of EDM4HEP](https://github.com/key4hep/EDM4hep/blob/main/python/edm4hep/__init__.py) data model.
+
+## Pythonizations
+
+Python as a language uses different constructions and conventions than C++, perfectly fine C++ code translated one to one to Python could be clunky by Python's standard. cppyy offers a mechanism called [pythonizations](https://cppyy.readthedocs.io/en/latest/pythonizations.html) to make the resulting bindings more pythonic. Some basic pythonizations are included automatically (for instance `operator[]` is translated to `__getitem__`) but others can be specified by a user.
+
+Podio comes with its own set pythonizations useful for the data models generated with it. To apply all the provided pythonizations to a `model_namespace` namespace:
+
+```python
+from podio.pythonizations import load_pythonizations
+
+load_pythonizations("model_namespace")
+```
+
+If only specific pythonizations should be applied:
+
+```python
+from  podio.pythonizations import collection_subscript # specific pythonization
+
+collection_subscript.CollectionSubscriptPythonizer.register("model_namespace")
+```
+
+### Developing new pythonizations
+
+To be discovered by `load_pythonizations`, any new pythonization should be placed in `podio.pythonizations` and be derived from the abstract class `podio.pythonizations.utils.pythonizer.Pythonizer`.
+
+A pythonization class should implement the following three class methods:
+
+- `priority`: The `load_pythonizations` function applies the pythonizations in increasing order of their `priority`
+- `filter`: A predicate to filter out classes to which given pythonization should be applied. See the [cppyy documentation](https://cppyy.readthedocs.io/en/latest/pythonizations.html#python-callbacks).
+- `modify`: Applying the modifications to the pythonized classes.
+
+### Considerations
+
+The cppyy pythonizations come with some considerations:
+
+- The general cppyy idea to lazily load only things that are needed applies only partially to the pythonizations. For instance, a pythonization modifying the `collection[]` will be applied the first time a class of `collection` is used, regardless if `collection[]` is actually used.
+- Each pythonization is applied to all the entities in a namespace and relies on a conditional mechanism (`filter` method) inside the pythonizations to select entities they modify. With a large number of pythonizations, the overheads will add up and slow down the usage of any class from a pythonized namespace.
+- The cppyy bindings hooking to the C++ routines are characterized by high performance compared to ordinary Python code. The pythonizations are written in Python and are executed at ordinary Python code speed.
--- a/doc/python.md
+++ b/doc/python.md
@@ -1,9 +1,9 @@
 # Python interface for data models
 
 Podio provides support for a Python interface for the generated data models. The [design choice](design.md) to create Python interface resembling the C++ interface is achieved by generating Python bindings from the C++ interface using
-[cppyy](https://cppyy.readthedocs.io/en/latest/index.html).
+[cppyy](https://cppyy.readthedocs.io/en/latest/index.html). To make pyROOT aware of the bindings, the cppyy functionality bundled with ROOT can be used.
 
-It's important to note that cppyy loads the bindings and presents them lazily at runtime to the Python interpreter, rather than writing Python interface files. Consequently, the Python bindings have a runtime dependency on both cppyy and the data model's C++ interface.
+It's important to note that cppyy loads the bindings and presents them lazily at runtime to the Python interpreter, rather than writing Python interface files. Consequently, the Python bindings have a runtime dependencies on ROOT, cppyy and the data model's C++ interface.
 
 To load the Python bindings from a generated C++ model dictionary, first make sure the model's library and headers can be found in `LD_LIBRARY_PATH` and `ROOT_INCLUDE_HEADERS` respectively, then:
 
@@ -15,13 +15,13 @@ if res < 0:
     raise RuntimeError('Failed to load libGeneratedModelDict.so')
 ```
 
-For reference usage, see [Python module of EDM4HEP](https://github.com/key4hep/EDM4hep/blob/main/python/edm4hep/__init__.py) data model.
+For reference usage, see the [Python module of EDM4hep](https://github.com/key4hep/EDM4hep/blob/main/python/edm4hep/__init__.py).
 
 ## Pythonizations
 
 Python as a language uses different constructions and conventions than C++, perfectly fine C++ code translated one to one to Python could be clunky by Python's standard. cppyy offers a mechanism called [pythonizations](https://cppyy.readthedocs.io/en/latest/pythonizations.html) to make the resulting bindings more pythonic. Some basic pythonizations are included automatically (for instance `operator[]` is translated to `__getitem__`) but others can be specified by a user.
 
-Podio comes with its own set pythonizations useful for the data models generated with it. To apply all the provided pythonizations to a `model_namespace` namespace:
+Podio comes with its own set of pythonizations useful for the data models generated with it. To apply all the provided pythonizations to a `model_namespace` namespace:
 
 ```python
 from podio.pythonizations import load_pythonizations
--- a/python/podio/pythonizations/__init__.py
+++ b/python/podio/pythonizations/__init__.py
@@ -1,0 +1,16 @@
+"""cppyy pythonizations for podio"""
+
+from importlib import import_module
+from pkgutil import walk_packages
+from .utils.pythonizer import Pythonizer
+
+
+def load_pythonizations(namespace):
+    """Register all available pythonizations for a given namespace"""
+    module_names = [name for _, name, _ in walk_packages(__path__)]
+    module_names = filter(lambda x: not x.startswith("test_"), module_names)
+    for module_name in module_names:
+        import_module(__name__ + "." + module_name)
+    pythonizers = sorted(Pythonizer.__subclasses__(), key=lambda x: x.priority())
+    for i in pythonizers:
+        i.register(namespace)
--- a/python/podio/pythonizations/__init__.py
+++ b/python/podio/pythonizations/__init__.py
@@ -7,8 +7,7 @@ from .utils.pythonizer import Pythonizer
 
 def load_pythonizations(namespace):
     """Register all available pythonizations for a given namespace"""
-    module_names = [name for _, name, _ in walk_packages(__path__)]
-    module_names = filter(lambda x: not x.startswith("test_"), module_names)
+    module_names = [name for _, name, _ in walk_packages(__path__) if not name.startswith("test_")]
     for module_name in module_names:
         import_module(__name__ + "." + module_name)
     pythonizers = sorted(Pythonizer.__subclasses__(), key=lambda x: x.priority())
--- a/python/podio/pythonizations/collection_subscript.py
+++ b/python/podio/pythonizations/collection_subscript.py
@@ -1,0 +1,27 @@
+"""Pythonize subscript operation for collections"""
+
+import cppyy
+from .utils.pythonizer import Pythonizer
+
+
+class CollectionSubscriptPythonizer(Pythonizer):
+    """Bound-check __getitem__ for classes derived from podio::CollectionBase"""
+
+    @classmethod
+    def priority(cls):
+        return 50
+
+    @classmethod
+    def callback(cls, class_, name):
+        def get_item(self, i):
+            if i >= len(self):
+                raise IndexError("collection index out of range")
+            return getitem_raw(self, i)
+
+        if (
+            issubclass(class_, cppyy.gbl.podio.CollectionBase)
+            and hasattr(class_, "__getitem__")
+            and hasattr(class_, "__len__")
+        ):
+            getitem_raw = class_.__getitem__
+            class_.__getitem__ = get_item
--- a/python/podio/pythonizations/collection_subscript.py
+++ b/python/podio/pythonizations/collection_subscript.py
@@ -16,8 +16,8 @@ class CollectionSubscriptPythonizer(Pythonizer):
         def get_item(self, i):
             try:
                 return self.at(i)
-            except cppyy.gbl.std.out_of_range as exc:
-                raise IndexError("collection index out of range") from exc
+            except cppyy.gbl.std.out_of_range:
+                raise IndexError("collection index out of range") from None
 
         if issubclass(class_, cppyy.gbl.podio.CollectionBase) and hasattr(class_, "at"):
             class_.__getitem__ = get_item
--- a/python/podio/pythonizations/collection_subscript.py
+++ b/python/podio/pythonizations/collection_subscript.py
@@ -12,12 +12,15 @@ class CollectionSubscriptPythonizer(Pythonizer):
         return 50
 
     @classmethod
-    def callback(cls, class_, name):
+    def filter(cls, class_, name):
+        return issubclass(class_, cppyy.gbl.podio.CollectionBase)
+
+    @classmethod
+    def modify(cls, class_, name):
         def get_item(self, i):
             try:
                 return self.at(i)
             except cppyy.gbl.std.out_of_range:
                 raise IndexError("collection index out of range") from None
 
-        if issubclass(class_, cppyy.gbl.podio.CollectionBase):
-            class_.__getitem__ = get_item
+        class_.__getitem__ = get_item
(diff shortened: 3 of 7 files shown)


tmadlener:
If I understand correctly this would effectively give an "opt-in" solution for generated EDMs, as the EDM has to explicitly load these pythonizations. For EDM4hep this would not really be a problem, as we have some python glue code already in any case. 

For me this looks like a good solution, but we should add a bit of documentation on how to use this in generated EDMs.

m-fila:
> If I understand correctly this would effectively give an "opt-in" solution for generated EDMs, as the EDM has to explicitly load these pythonizations. For EDM4hep this would not really be a problem, as we have some python glue code already in any case.
> 
yes, it's opt-in. `load_pythonizations` could be added to EDM4hep's `__init__.py` to load on import

> For me this looks like a good solution, but we should add a bit of documentation on how to use this in generated EDMs.

Thanks, I will add it

python/podio/pythonizations/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+
+from importlib import import_module
+from pkgutil import walk_packages
+from .utils.pythonizer import Pythonizer
+
+
+def load_pythonizations(namespace):
+    """Register all available pythonizations for a given namespace"""
+    module_names = [name for _, name, _ in walk_packages(__path__)]
+    module_names = filter(lambda x: not x.startswith("test_"), module_names)
```
jmcarcell, line 11:
```suggestion
    module_names = [name for _, name, _ in walk_packages(__path__) if not name.startswith("test_")]
```

python/podio/pythonizations/collection_subscript.py
```diff
@@ -0,0 +1,27 @@
...
+    def callback(cls, class_, name):
+        def get_item(self, i):
+            if i >= len(self):
+                raise IndexError("collection index out of range")
+            return getitem_raw(self, i)
+
+        if (
+            issubclass(class_, cppyy.gbl.podio.CollectionBase)
+            and hasattr(class_, "__getitem__")
+            and hasattr(class_, "__len__")
```
jmcarcell, line 24: This should always be true if something is a `podio.CollectionBase` right? For `__getitem__` its the collections who implement it themselves.

jmcarcell:
Note that this will make indexing significantly slower:
With this PR and the one in EDM4hep:
``` python
In [2]: import edm4hep

In [3]: coll = edm4hep.MCParticleCollection()

In [4]: coll.create()
Out[4]: <cppyy.gbl.edm4hep.MutableMCParticle object at 0x55f984f08120>

In [5]: %timeit coll[0]
3.45 µs ± 72.6 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```
and without this PR:
```
2.46 µs ± 26.6 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

python/podio/pythonizations/collection_subscript.py
```diff
@@ -0,0 +1,27 @@
...
+
+class CollectionSubscriptPythonizer(Pythonizer):
+    """Bound-check __getitem__ for classes derived from podio::CollectionBase"""
+
+    @classmethod
+    def priority(cls):
+        return 50
+
+    @classmethod
+    def callback(cls, class_, name):
```
jmcarcell, line 15: So every callback runs once per object in the namespace they are tied to? That would mean every EDM4hep object, collection or not

m-fila:
> Note that this will make indexing significantly slower

Thanks for pointing this out, I'll check if this can be improved by mapping `__getitem__` to `at` and raising `IndexError`

python/podio/pythonizations/collection_subscript.py
m-fila: Once registered, cppyy will try to lazily apply the callback to all the entities in a given namespace. So the callbacks usually start with a conditional to filter out when they should be applied.
Yes, it creates a small overhead for importing any definition from that namespace

python/podio/pythonizations/collection_subscript.py
m-fila: Yes
```python
            and hasattr(class_, "__getitem__")
            and hasattr(class_, "__len__")
```
are redundant. I still prefer to have them because:
- The methods are not used directly in a callback, instead the callback merely adds another method that calls them. So there will be no error that a pythonization can't be applied, the classes will just get broken.
- The exception from a  method broken with pythonization is rather cryptic: ` RecursionError: maximum recursion depth exceeded while calling a Python object`

If think it would be more readable to not apply the exception and learn from a test that it was not applied

m-fila: And yes both `__len__` and `__getitem__` are created automatically by cppyy if  `size()` and `operator()` are present

m-fila:
I  tried benchmarking on following function:
```python
def bench():
    try:
        coll[0]
    except:
        pass

coll = MCParticleCollection()
bench() # give cppyy time to load lazily
%timeit bench()

coll.create()
bench() # give cppyy time to load lazily
%timeit bench()
```

|    | Bound check | `__getitem__` implementation       | Average time, no-exception branch | Average time, exception branch |
|----|-------------|------------------------------------|---------------------------------|------------------------------|
| 1. | ❌           | Default cppyy mapping to `operator[]`    |               683 ns              |             -        |
| 2. | ✔️           | Current PR, bound-check in python  |               1.2 µs              |             779 ns             |
| 3. | ✔️           | Mapped to `at`                     |               730 ns              |             30.8 µs            |
| 4. | ✔️           | `at` wrapped to raise `IndexError` |               826 ns              |             32.1 µs            |

The implementations using `at` seems to be more performant in the usual case. Would it be acceptable to go with option 3. or 4.?

A C++ callback added to the Collection template could potentialy achieve the same as `at` (option 3.) on the no-exception branch and slightly better than python side bound-check (option 2.) on the exception branch. But then extending with other c++ callbacks would be very ugly

tmadlener:
I see there has been quite some discussion (and progress?) here. A few comments / questions from my side in no particular order:

- Did you by any chance check how the times compare to pure c++? Mainly for comparison, I don't think we learn too much from them other than a 30 % increase on the python side is pretty negligible compared to pure c++.
- I think the exceptional case should not be the (main) driver for performance considerations. After all you are dealing with an error at that point and not with something "usual".
- IIUC all of the presented methods are still possible with python callbacks, right?

m-fila:
>Did you by any chance check how the times compare to pure c++? Mainly for comparison, I don't think we learn too much from them other than a 30 % increase on the python side is pretty negligible compared to pure c++.

Pure C++ google benchmark between a single element access with `operator[]` vs `at`:
```txt
----------------------------------------------------------------
Benchmark                      Time             CPU   Iterations
----------------------------------------------------------------
squareBracketsNoError       4.59 ns         4.59 ns    152241304
atNoError                   5.26 ns         5.25 ns    131108856
atError                     1552 ns         1550 ns       452227
```
where
```cpp
static void squareBracketsNoError(benchmark::State& state) {
  edm4hep::MCParticleCollection coll;
  coll.create();
  for (auto _ : state) {
    try {
      benchmark::DoNotOptimize(coll[0]);
      benchmark::ClobberMemory();
    } catch (const std::out_of_range&) {
    }
  }
}

static void atNoError(benchmark::State& state) {
  edm4hep::MCParticleCollection coll;
  coll.create();
  for (auto _ : state) {
    try {
      benchmark::DoNotOptimize(coll.at(0));
      benchmark::ClobberMemory();
    } catch (const std::out_of_range&) {
    }
  }
}

static void atError(benchmark::State& state) {
  edm4hep::MCParticleCollection coll;
  coll.create();
  for (auto _ : state) {
    try {
      benchmark::DoNotOptimize(coll.at(1));
      benchmark::ClobberMemory();
    } catch (const std::out_of_range&) {
    }
  }
}
```
both this and python benchmark are isolated single access, looping over the collection will probably give different results

> IIUC all of the presented methods are still possible with python callbacks, right?

Yes, all the alternatives from that table were python callbacks

> I think the exceptional case should not be the (main) driver for performance considerations. After all you are dealing with an error at that point and not with something "usual".

In the last commit I changed the implementation proposed in PR to option 4 from the table. It seems to me as a reasonable compromise

python/podio/pythonizations/collection_subscript.py
```diff
@@ -0,0 +1,23 @@
...
+        def get_item(self, i):
+            try:
+                return self.at(i)
+            except cppyy.gbl.std.out_of_range as exc:
+                raise IndexError("collection index out of range") from exc
```
tmadlener, line 20: How decipherable will the stacktrace be in this case? I.e. how many levels do we add with this callback mechanism? Or is cppyy clever enough to hide a bunch of it's internals and make it quite obvious where the actual error is? I am asking because people usually don't read the full stacktrace, especially if it's long.

tmadlener (commented):
Thanks for also checking the c++ times. The orders of magnitude difference between c++ and python looks like expected, I would say.

I would also vote for option 4 here. I just have one minor question below.

python/podio/pythonizations/collection_subscript.py
m-fila: Example:
```python
>>> import edm4hep
>>> coll = edm4hep.MCParticleCollection()
>>> def foo(x):
...      x[0]
... 
>>> foo(coll)
  File "/scratch/podio/python/podio/pythonizations/collection_subscript.py", line 18, in get_item
    return self.at(i)
cppyy.gbl.std.out_of_range: none of the 2 overloaded methods succeeded. Full details:
  edm4hep::MCParticle edm4hep::MCParticleCollection::at(unsigned long index) =>
    out_of_range: deque::_M_range_check: __n (which is 0)>= this->size() (which is 0)
  edm4hep::MutableMCParticle edm4hep::MCParticleCollection::at(unsigned long index) =>
    out_of_range: deque::_M_range_check: __n (which is 0)>= this->size() (which is 0)

The above exception was the direct cause of the following exception:

  File "<stdin>", line 2, in foo
  File "/scratch/podio/python/podio/pythonizations/collection_subscript.py", line 20, in get_item
    raise IndexError("collection index out of range") from exc
IndexError: collection index out of range
>>> 
```
The context printout could be suppressed by raising `from None` if you find the following more readable:
```python
>>> foo(coll)
  File "<stdin>", line 2, in foo
  File "/scratch/podio/python/podio/pythonizations/collection_subscript.py", line 20, in get_item
    raise IndexError("collection index out of range") from None
IndexError: collection index out of range
```

In both cases I'd say it's rather easy to spot the culprit in `line 2, in foo`

tmadlener: I have a slight preference for the `from None` version, simply because it is less to read for essentially the same information. The other one just dumps a bunch of implementation details as well, IMHO.

python/podio/pythonizations/collection_subscript.py
jmcarcell: It's being called more often than one would think, for example, after adding a print in the callback:
```
In [1]: import edm4hep

In [2]: edm4hep.<TAB>
Calling callback from collection_subscript.py

In [2]: coll = edm4hep.MCParticleCollection()
Calling callback from collection_subscript.py

In [3]: coll.create()
Calling callback from collection_subscript.py
Out[3]: <cppyy.gbl.edm4hep.MutableMCParticle object at 0x59e0115f8cd0>

In [4]: edm4hep.<TAB>
Calling callback from collection_subscript.py
Calling callback from collection_subscript.py
Calling callback from collection_subscript.py
Calling callback from collection_subscript.py
```
and after creating more objects `edm4hep.<TAB>` will call the callback many many times. I'm not sure if much can be done about that, otherwise one has to be careful what is put in the callbacks or some parts of using python interactively will become unusable.

m-fila: Thanks for checking. I get a slightly different behavior and the callback seems to be indeed called only once for each class in namespace and only when cppyy loads it for the first time
```python
>>> import edm4hep
>>> edm4hep.<TAB>
Calling callback for Quantity
Calling callback for RawTimeSeriesData
Calling callback for MutableMCRecoCaloAssociation
...
Display all 123 possibilities? (y or n)
edm4hep.CaloHitContribution(
edm4hep.CaloHitContributionCollection(
edm4hep.CaloHitContributionData(
...
>>
>>> coll = edm4hep.MCParticleCollection()
>>> coll.create()
<cppyy.gbl.edm4hep.MutableMCParticle object at 0x99d8b70>
>>> edm4hep.<TAB>
Display all 123 possibilities? (y or n)
edm4hep.CaloHitContribution(
edm4hep.CaloHitContributionCollection(
edm4hep.CaloHitContributionData(
```
with modified callback
```python
    @classmethod
    def callback(cls, class_, name):
        def get_item(self, i):
            try:
                return self.at(i)
            except cppyy.gbl.std.out_of_range:
                raise IndexError("collection index out of range") from None

        print(f"Calling callback for {name}")
        if issubclass(class_, cppyy.gbl.podio.CollectionBase) and hasattr(class_, "at"):
            class_.__getitem__ = get_item
```

python/podio/pythonizations/collection_subscript.py
```diff
@@ -0,0 +1,23 @@
...
+
+    @classmethod
+    def callback(cls, class_, name):
+        def get_item(self, i):
+            try:
+                return self.at(i)
+            except cppyy.gbl.std.out_of_range:
+                raise IndexError("collection index out of range") from None
+
+        if issubclass(class_, cppyy.gbl.podio.CollectionBase) and hasattr(class_, "at"):
```
tmadlener, line 22:
```suggestion
        if issubclass(class_, cppyy.gbl.podio.CollectionBase):
```
I am not sure if this check is expensive enough to have any performance implications, but a `podio::CollectionBase` will always have an `at` method.

python/podio/pythonizations/collection_subscript.py
tmadlener: Could this be due to some `%autoreload` magic? I could imagine that would influence how often the callback is called.

m-fila: I see. So the difference is that plain python's tab-completion on a module forces cppyy to load everything now but ipython's loads only a few entities now and the rest is loaded lazily. In both scenarios each callback will be called at most once for each class. I was worried that somehow the callbacks are called over and over again for the same things, but it's not the case here.

It seems that `%autoreload`  doesn't reset the state of cppyy. So neither the shared libraries are read again nor callbacks applied again

jmcarcell (approved):
This functionality should be in podio. For EDM4hep, I would like to have only what's necessary since there can be some drawbacks for interactive usage

m-fila:
Sorry for the delay. Rebased and added documentation. I also split the callback into two functions:
- conditional to filter classes to by modified
- modification body doing actual work here

I think previously  it was not clear that callback should start with some conditional statement to select what should be modified

doc/python.md
```diff
@@ -0,0 +1,56 @@
+# Python interface for data models
+
+Podio provides support for a Python interface for the generated data models. The [design choice](design.md) to create Python interface resembling the C++ interface is achieved by generating Python bindings from the C++ interface using
+[cppyy](https://cppyy.readthedocs.io/en/latest/index.html).
+
+It's important to note that cppyy loads the bindings and presents them lazily at runtime to the Python interpreter, rather than writing Python interface files. Consequently, the Python bindings have a runtime dependency on both cppyy and the data model's C++ interface.
```
tmadlener, line 6: I think we should mention here that we automatically get them via ROOT, so that people are not confused when they suddenly see cppyy. Especially since we continue with ROOT below.

doc/python.md
```diff
@@ -0,0 +1,56 @@
...
+
+```python
+import ROOT
+
+res = ROOT.gSystem.Load('libGeneratedModelDict.so')
+if res < 0:
+    raise RuntimeError('Failed to load libGeneratedModelDict.so')
+```
+
+For reference usage, see [Python module of EDM4HEP](https://github.com/key4hep/EDM4hep/blob/main/python/edm4hep/__init__.py) data model.
```
tmadlener, line 18:
```suggestion
For reference usage, see the [Python module of EDM4hep](https://github.com/key4hep/EDM4hep/blob/main/python/edm4hep/__init__.py).
```

python/podio/test_CodeGen.py
```diff
@@ -31,3 +37,15 @@ def test_add(self):
...
+
+class CollectionSubscriptTest(unittest.TestCase):
+    """Collection subscript test"""
+
+    def test_bound_check(self):
+        collection = nsp.EnergyInNamespaceCollection()
+        _ = collection.create()
+        self.assertEqual(len(collection), 1)
+        with self.assertRaises(IndexError):
+            _ = collection[20]
```
tmadlener, line 50: This fails on `dev3` due to some overload resolution issue. I would assume it's a ROOT issue that we are hitting here, but I am not sure yet which one.

doc/python.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Pythonizations
+
+Python as a language uses different constructions and conventions than C++, perfectly fine C++ code translated one to one to Python could be clunky by Python's standard. cppyy offers a mechanism called [pythonizations](https://cppyy.readthedocs.io/en/latest/pythonizations.html) to make the resulting bindings more pythonic. Some basic pythonizations are included automatically (for instance `operator[]` is translated to `__getitem__`) but others can be specified by a user.
+
+Podio comes with its own set pythonizations useful for the data models generated with it. To apply all the provided pythonizations to a `model_namespace` namespace:
```
tmadlener, line 24:
```suggestion
Podio comes with its own set of pythonizations useful for the data models generated with it. To apply all the provided pythonizations to a `model_namespace` namespace:
```

python/podio/test_CodeGen.py
m-fila: Well an exception is thrown, but of a different type than expected:
```txt
    File "/home/runner/work/podio/podio/python/podio/test_CodeGen.py", line 50, in test_bound_check
      _ = collection[20]
    File "/home/runner/work/podio/podio/python/podio/pythonizations/collection_subscript.py", line 22, in get_item
      return self.at(i)
  TypeError: none of the 2 overloaded methods succeeded. Full details:
    nsp::EnergyInNamespace nsp::EnergyInNamespaceCollection::at(size_t index) =>
      out_of_range: deque::_M_range_check: __n (which is 20)>= this->size() (which is 1)
    nsp::MutableEnergyInNamespace nsp::EnergyInNamespaceCollection::at(size_t index) =>
      out_of_range: deque::_M_range_check: __n (which is 20)>= this->size() (which is 1)
```

According to the [docs](https://cppyy.readthedocs.io/en/latest/functions.html#overloads-and-exceptions) if there is an exception then cppyy checks the rest of overloads and if they all throw the same type then the exception is propagated, if they all throw different types then `TypeError` is risen
Here both overloads throws `std::out_of_range` but somehow it isn't recognized as the same type

The possible solutions I could think of now are:
- explicitly select overload - here the overloads are const and non-const. Selection between const and non-const was added in cppyy-1.7.0, nightlies have 1.6.2
- change pythonization to catch either `cppyy.gbl.std.out_of_range` or `TypeError` and raise `IndexError`- I don't like it becasue it's also used for other things, like actual type mismatch, eg. calling with `str` instead of `int`
- allow both types of exception in test - getting an exception is already an improvement, the message says it's out of range. The users get confused why it's `TypeError` not `IndexError`

tmadlener: Maybe this can be boiled down to a small reproducer that we can submit to ROOT because they probably didn't want to introduce too many "breaking" changes in their update of the cppyy they bundle.

m-fila: Opened an issue wlav/cppyy#230 as the problems appears also outside of ROOT bundle

tmadlener: I also made the ROOT folks aware via https://github.com/root-project/root/issues/15375

tmadlener:
Looks like all the necessary fixes have landed in upstream root (master and v6.32). IIRC this can be merged now, right?

m-fila:
> Looks like all the necessary fixes have landed in upstream root (master and v6.32). IIRC this can be merged now, right?

Awesome 😎 Yes, it's ready to go

hegner:
and off we go...

PR #573, opened 2024-03-20
Remove outdated information from top level README

tmadlener:
BEGINRELEASENOTES
- Remove outdated information from top level README

ENDRELEASENOTES


jmcarcell:
That the test directory has to be in LD_LIBRARY_PATH to run the tests manually I would say is a common pitfall

tmadlener:
Yes, but in this case `ctest` has the correct environment built-in. Some of the things that were mentioned have been gone for quite some time now (tests have been moved in #428 and removed with #485)

PR #575, opened 2024-04-02
Make sure that the `as_type` argument is respected for `Frame.put_parameters` for lists of values as well

tmadlener:
BEGINRELEASENOTES
- Make sure that the `as_type` argument is respected in `Frame.put_parameter` also for lists of doubles / floats.
- Fix test to actually check for this.

ENDRELEASENOTES

### DIFF
--- a/python/podio/frame.py
+++ b/python/podio/frame.py
@@ -249,7 +249,7 @@ class Frame:
                 raise ValueError(f"Cannot put a parameter of type {type_name} into a Frame")
 
             par_type = vec_types[0]
-            if isinstance(value[0], float):
+            if as_type is None and isinstance(value[0], float):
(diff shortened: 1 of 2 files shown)


tmadlener:
The recent updates to cppyy in root mater also broke the resolution of `Frame::put` from the python side. For some reason I haven't yet figured out it runs into a copy constructor for a `unique_ptr` which has not been the case before. Marking the argument as rvalue reference makes things work again for root@master.

tmadlener:
The issue relating to `Frame::put` has been reported to ROOT here: https://github.com/root-project/root/issues/15117

tmadlener:
Merging this as it fixes a false positive test and we have to wait for ROOT to see how to fix the other python issue for upcoming versions.

PR #577, opened 2024-04-09
Improve error message for push_back when using an immutable type

jmcarcell:
BEGINRELEASENOTES
- Improve error message for push_back when using an immutable type

ENDRELEASENOTES

Possible (wrong) meaning that can be inferred from the current one: this is a subset collection and you are trying to push a mutable object so "Can only add immutable objects to subset collections", therefore you should change your mutable object to immutable object (wrong conclusion).

With the new one I don't think the wrong meaning can be inferred; it means you are pushing an immutable object but "Immutable objects can only be added to subset collections", which means it's not a subset collection, and the answer here is to change from immutable to mutable.


tmadlener (approved):
Thanks. Indeed less confusing this way.

PR #578, opened 2024-04-09
Check if the data passed to the frame model is not a nullptr

jmcarcell:
BEGINRELEASENOTES
- Check if the data passed to the frame model is not a nullptr before doing anything with it.

ENDRELEASENOTES

It looks like a good idea to check if the pointer is null or not regardless. This makes opening corrupted or wrong files throw

```
terminate called after throwing an instance of 'std::invalid_argument'
  what():  Building a Frame failed; if you are reading from a file it may be corrupted
```

instead of crashing.

### DIFF
--- a/include/podio/Frame.h
+++ b/include/podio/Frame.h
@@ -391,11 +391,13 @@ const CollT& Frame::put(CollT&& coll, const std::string& name) {
 
 template <typename FrameDataT>
 Frame::FrameModel<FrameDataT>::FrameModel(std::unique_ptr<FrameDataT> data) :
-    m_mapMtx(std::make_unique<std::mutex>()),
-    m_data(std::move(data)),
-    m_dataMtx(std::make_unique<std::mutex>()),
-    m_idTable(std::move(m_data->getIDTable())),
-    m_parameters(std::move(m_data->getParameters())) {
+    m_mapMtx(std::make_unique<std::mutex>()), m_dataMtx(std::make_unique<std::mutex>()) {
+  if (!data) {
+    throw std::invalid_argument("Building a Frame failed; if you are reading from a file it may be corrupted");
+  }
+  m_data = std::move(data);
+  m_idTable = std::move(m_data->getIDTable());
+  m_parameters = std::move(m_data->getParameters());
 }
 
 template <typename FrameDataT>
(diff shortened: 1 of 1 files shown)


include/podio/Frame.h
```diff
@@ -391,11 +391,13 @@ const CollT& Frame::put(CollT&& coll, const std::string& name) {
...
 template <typename FrameDataT>
 Frame::FrameModel<FrameDataT>::FrameModel(std::unique_ptr<FrameDataT> data) :
-    m_mapMtx(std::make_unique<std::mutex>()),
-    m_data(std::move(data)),
-    m_dataMtx(std::make_unique<std::mutex>()),
-    m_idTable(std::move(m_data->getIDTable())),
-    m_parameters(std::move(m_data->getParameters())) {
+    m_mapMtx(std::make_unique<std::mutex>()), m_dataMtx(std::make_unique<std::mutex>()) {
+  if (!data) {
+    throw std::invalid_argument("Building a Frame failed; if you are reading from a file it may be corrupted");
```
jmcarcell, line 396:
```suggestion
    throw std::runtime_error("Building a Frame failed; if you are reading from a file it may be corrupted");
```
Or this, initially I was going to have a message about FrameModel but no one will know what that is so it would be a bit useless.

tmadlener:
In principle you could also be reading past the available number of entries, as I originally envisioned the user to check whether the `uniqe_ptr<FramaData>` they retrieve from the reader is a valid pointer. However, I see it's not documented, so we should also add something to the docstring of the constructor here:

https://github.com/AIDASoft/podio/blob/b94cc639b884356b8eb43b017586d86ed513558a/include/podio/Frame.h#L149-L155

I would then make the exception message mainly about receiving a `nullptr` for the `FrameData`.

include/podio/Frame.h
```diff
@@ -150,7 +150,8 @@ class Frame {
   ///
   /// @tparam FrameDataT Arbitrary data container that provides access to the
   ///                    collection buffers as well as the metadata, when
-  ///                    requested by the Frame.
+  ///                    requested by the Frame. The unique_ptr has to be checked
+  ///                    for validity or this constructor will throw an std::invalid_argument
```
tmadlener, line 154:
```suggestion
  ///                    requested by the Frame. The unique_ptr has to be checked
  ///                    for validity before calling this construtor.
  ///
  /// @throws std::invalid_argument if the passed pointer is a nullptr.
```

since doxygen has this tag.

tmadlener:
Not entirely sure why the `dev4` workflows are failing now. The issue looks like https://github.com/root-project/root/issues/15117 but this should have been fixed. Merging this for now.

PR #583, opened 2024-04-15
Fix OneToManyRelations and VectorMembers in cloned objects

jmcarcell:
The context is the following. Let's say we have a collection and each member has a OneToManyRelation to another type, for this example something that looks like a string. Let's say the first element in the collection has a relation to 'a', then the second to 'b' and so on. Each 'a', 'b', 'c', ... are in a different vector at this point. When saving the collection, all of 'a', 'b', 'c', ... go into the same vector, and there are some indexes to know which part of the vector corresponds to each object in the collection (for example for the first element it would be `start=0` and `end=1`. When reading, each object still points to this big vector ['a', 'b', 'c', ...] so if the object is cloned the clone will also point to this big vector. If I clone the first object and then do a `addToRelation('z')`, the big vector will have one more element ['a', 'b', 'c', ..., 'z'] and the indexes will be updated so that now the relations of the first element will be 'a' and 'b' (`start=0` and `end=2`), when the expected result is 'a' and 'z'.

BEGINRELEASENOTES
- Fix OneToManyRelations and VectorMembers in cloned objects. Currently, pushing back these fields to a cloned object does not give the expected result, because the objects that we get after reading are not the same as we had before writing.
- Add some code testing this behavior: pushing back to cloned objects that have been read from a file and also after cloning the cloned object. 

ENDRELEASENOTES

Initially I implemented it (check if this is the case and if it is then make a new vector) for the `clone()` method but in [this commit](https://github.com/AIDASoft/podio/pull/583/commits/3323f2370f0b85b8ededd8f07a6d3820d5b931b6) I have done it for the `addX` methods, since it's closer to on demand and won't make all the `clone()` a bit slower.

### DIFF
--- a/python/templates/macros/implementations.jinja2
+++ b/python/templates/macros/implementations.jinja2
@@ -28,8 +28,9 @@ Mutable{{ type }} {{ full_type }}::clone() const {
 {% for relation in multi_relations %}
   // If the current object has been read from a file, then the object may only have a slice of the relation vector
   // so this slice has to be copied in case we want to modify it
-  delete tmp->m_{{ relation.name }};
-  tmp->m_{{ relation.name }} = new std::vector<{{ relation.full_type }}>(m_obj->m_{{ relation.name }}->begin() + m_obj->data.{{ relation.name }}_begin, m_obj->m_{{ relation.name }}->begin() + m_obj->data.{{ relation.name }}_end);
+  for (size_t i = m_obj->data.{{ relation.name }}_begin; i < m_obj->data.{{ relation.name }}_end; i++) {
+    tmp->m_{{ relation.name }}->push_back((*m_obj->m_{{ relation.name }})[i]);
+  }
   tmp->data.{{ relation.name }}_begin = 0;
   tmp->data.{{ relation.name }}_end = tmp->m_{{ relation.name }}->size();
 {% endfor %}
--- a/python/templates/macros/implementations.jinja2
+++ b/python/templates/macros/implementations.jinja2
@@ -1,4 +1,4 @@
-{% macro constructors_destructors(type, members, relations=[], prefix='') %}
+{% macro constructors_destructors(type, members, multi_relations=[], prefix='') %}
 {% set full_type = prefix + type %}
 
 {{ full_type }}::{{ full_type }}() :
@@ -20,7 +20,7 @@
 
 Mutable{{ type }} {{ full_type }}::clone() const {
   auto tmp = Mutable{{ type }}(podio::utils::MaybeSharedPtr(new {{ type }}Obj(*m_obj), podio::utils::MarkOwned));
-{% for relation in relations %}
+{% for relation in multi_relations %}
   if (m_obj->data.{{ relation.name }}_end != m_obj->m_{{ relation.name }}->size()) {
     tmp.m_obj->m_{{ relation.name }} = new std::vector<{{ relation.full_type }}>(m_obj->m_{{ relation.name }}->begin() + m_obj->data.{{ relation.name }}_begin, m_obj->m_{{ relation.name }}->begin() + m_obj->data.{{ relation.name }}_end);
   }
--- a/python/templates/macros/implementations.jinja2
+++ b/python/templates/macros/implementations.jinja2
@@ -1,4 +1,4 @@
-{% macro constructors_destructors(type, members, prefix='') %}
+{% macro constructors_destructors(type, members, multi_relations=[], prefix='') %}
 {% set full_type = prefix + type %}
 
 {{ full_type }}::{{ full_type }}() :
@@ -20,6 +20,13 @@
 
 Mutable{{ type }} {{ full_type }}::clone() const {
   auto tmp = Mutable{{ type }}(podio::utils::MaybeSharedPtr(new {{ type }}Obj(*m_obj), podio::utils::MarkOwned));
+{% for relation in multi_relations %}
+  if (m_obj->data.{{ relation.name }}_end != m_obj->m_{{ relation.name }}->size()) {
+    tmp.m_obj->m_{{ relation.name }} = new std::vector<{{ relation.full_type }}>(m_obj->m_{{ relation.name }}->begin() + m_obj->data.{{ relation.name }}_begin, m_obj->m_{{ relation.name }}->begin() + m_obj->data.{{ relation.name }}_end);
+    tmp.m_obj->data.{{ relation.name }}_begin = 0;
+    tmp.m_obj->data.{{ relation.name }}_end = tmp.m_obj->m_{{ relation.name }}->size();
+  }
+{% endfor %}
   return tmp;
 }
 
@@ -100,11 +107,6 @@ void {{ class_type }}::{{ relation.setter_name(get_syntax) }}({{ relation.full_t
 {% for relation in relations %}
 {% if with_adder %}
 void {{ class_type }}::{{ relation.setter_name(get_syntax, is_relation=True) }}({{ relation.full_type }} component) {
-  if (m_obj->data.{{ relation.name }}_end != m_obj->m_{{ relation.name }}->size()) {
-    m_obj->m_{{ relation.name }} = new std::vector<{{ relation.full_type }}>(m_obj->m_{{ relation.name }}->begin() + m_obj->data.{{ relation.name }}_begin, m_obj->m_{{ relation.name }}->begin() + m_obj->data.{{ relation.name }}_end);
-    m_obj->data.{{ relation.name }}_begin = 0;
-    m_obj->data.{{ relation.name }}_end = m_obj->m_{{ relation.name }}->size();
-  }
   m_obj->m_{{ relation.name }}->push_back(component);
   m_obj->data.{{ relation.name }}_end++;
 }
--- a/python/templates/macros/implementations.jinja2
+++ b/python/templates/macros/implementations.jinja2
@@ -28,6 +28,7 @@ Mutable{{ type }} {{ full_type }}::clone() const {
 {% for relation in multi_relations %}
   // If the current object has been read from a file, then the object may only have a slice of the relation vector
   // so this slice has to be copied in case we want to modify it
+  tmp->m_{{ relation.name }}->reserve(m_obj->m_{{ relation.name }}->size());
   for (size_t i = m_obj->data.{{ relation.name }}_begin; i < m_obj->data.{{ relation.name }}_end; i++) {
     tmp->m_{{ relation.name }}->push_back((*m_obj->m_{{ relation.name }})[i]);
   }
--- a/python/templates/macros/implementations.jinja2
+++ b/python/templates/macros/implementations.jinja2
@@ -1,4 +1,4 @@
-{% macro constructors_destructors(type, members, multi_relations=[], prefix='') %}
+{% macro constructors_destructors(type, members, prefix='') %}
 {% set full_type = prefix + type %}
 
 {{ full_type }}::{{ full_type }}() :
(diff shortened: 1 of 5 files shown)


python/templates/macros/implementations.jinja2
```diff
@@ -99,6 +100,11 @@ void {{ class_type }}::{{ relation.setter_name(get_syntax) }}({{ relation.full_t
 {% for relation in relations %}
 {% if with_adder %}
 void {{ class_type }}::{{ relation.setter_name(get_syntax, is_relation=True) }}({{ relation.full_type }} component) {
+  if (m_obj->data.{{ relation.name }}_end != m_obj->m_{{ relation.name }}->size()) {
```
tmadlener, line 103: Can you add a comment here that says why we are doing this for the future? Should we just make this copy in the `clone` call directly without waiting for it to trigger here? The effect would be the same, but I think the code would be easier to follow.

tmadlener (commented):
Thanks for catching and fixing this. This seems to mainly fix things for cloning things that are read from file. I suppose it also works for things that have just been created in memory.

Is there an easy way to put the new tests into the Catch2 unittest harness?

python/templates/macros/implementations.jinja2
tmadlener: I think in this case I would go for the implementation in `clone` because it's easier to see why it happens there. We can always come back and move it to a slightly more unintuitive place once we see performance is actually an issue. In the end for most relations we are copying a vector of pointers and most vector members are also pretty small objects.

python/templates/macros/implementations.jinja2
```diff
@@ -19,7 +19,8 @@
 }
 
 Mutable{{ type }} {{ full_type }}::clone() const {
-  return Mutable{{ type }}(podio::utils::MaybeSharedPtr(new {{ type }}Obj(*m_obj), podio::utils::MarkOwned));
+  auto tmp = Mutable{{ type }}(podio::utils::MaybeSharedPtr(new {{ type }}Obj(*m_obj), podio::utils::MarkOwned));
+  return tmp;
```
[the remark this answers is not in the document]
jmcarcell: This is needed for implenting the fix in `clone()`, when the comment was made the implementation was in the `addX` methods and it doesn't make sense

python/templates/macros/implementations.jinja2
jmcarcell: I had a look at saving cloned objects and that's worse when it's implemented in the `addX` because after cloning each element will point to the big vector and for each element saved a big vector will also be saved (even though when reading since we have the indexes it's fine). (For example, if element 0 has a relation to 'a', 1 to 'b', 2 to 'c' and 3 to 'd', if 0 is cloned and saved then ['a', 'b', 'c', 'd'] is saved but if 0 and 1 are both cloned then ['a', 'b', 'c', 'd', 'a', 'b', 'c', 'd'] is saved. So certainly having it in `addX` with the same behaviour as in `clone()` is more complicated

jmcarcell:
I changed the fix to be in `clone()` and moved the tests in a unittest (maybe I should add an rntuple one...)

> I suppose it also works for things that have just been created in memory.

For memory only everything works fine because there is never this conversion to a single vector, that only happens when writing

python/templates/macros/implementations.jinja2
tmadlener: Yeah, I had this though over lunch as well. I think `clone` would have to copy contents to a different vector in order to make the writing side work again as well. I am currently having a look to remember which one should actually be used for this.

tmadlener: OK. I think the additional vector I was thinking about doesn't really play a role here, it's in `CollectionData` and should be handled correctly if we make sure to handle the vector correctly here.

In that case I agree with you and adding things in `clone` is the easier way, because otherwise we have to keep track of whether `addX` has been called and then do the copy in any case before writing in order for the rest to work as expected.

tmadlener (commented):
This looks good to me and I am fairly certain everything works as expected also in the case where the relations are empty (when read from file). Is it easily possible to have a test for that as well?

python/templates/macros/implementations.jinja2
```diff
@@ -19,7 +19,17 @@
 }
 
 Mutable{{ type }} {{ full_type }}::clone() const {
-  return Mutable{{ type }}(podio::utils::MaybeSharedPtr(new {{ type }}Obj(*m_obj), podio::utils::MarkOwned));
+  auto tmp = Mutable{{ type }}(podio::utils::MaybeSharedPtr(new {{ type }}Obj(*m_obj), podio::utils::MarkOwned));
+{% for relation in multi_relations %}
+  // If the current object has been read from a file, then the object may only have a slice of the relation vector
+  // so this slice has to be copied in case we want to modify it
+  if (m_obj->data.{{ relation.name }}_end - m_obj->data.{{ relation.name }}_begin != m_obj->m_{{ relation.name }}->size()) {
+    tmp.m_obj->m_{{ relation.name }} = new std::vector<{{ relation.full_type }}>(m_obj->m_{{ relation.name }}->begin() + m_obj->data.{{ relation.name }}_begin, m_obj->m_{{ relation.name }}->begin() + m_obj->data.{{ relation.name }}_end);
```
tmadlener, line 27: Are we leaking the `vector` that is created in the `tmp` constructor here by not getting rid of it explicitly? I think we are and the AddressSanitizer should in principle catch this in the unittests, but we would need to go through an I/O backend that has no memory issues. SIO and RNTuple should both work for that. Then we could remove the `ASAN-FAIL` (and hopefully also `UBSAN-FAIL` labels in the tests below.

jmcarcell: Here I would say that no, or at least that this is the same situation as when a new object is created since then also the vector is created with `new std::vector`. When cloning a collection that is read, the initial big vector is still managed by the read collection and its objects and the new vector by the cloned object.
Sanitizers seem to be happy. Locally all the rntuple tests fail for me with either gcc 13 or clang 17 and ROOT 6.30.06

python/templates/macros/implementations.jinja2
jmcarcell: > In that case I agree with you and adding things in `clone` is the easier way, because otherwise we have to keep track of whether `addX` has been called and then do the copy in any case before writing in order for the rest to work as expected.

I think no tracking needs to be done here. The workflow would be to have this fix in `addX`, then when writing we can have objects that are pointing to a big vector but we may not want to write it all and objects that are pointing to a "small" vector so we have to use the `X_begin` and `X_end` indexes to create a new big vector to only write what the objects are pointing to. Probably starting here: https://github.com/AIDASoft/podio/blob/master/python/templates/CollectionData.cc.jinja2#L134

python/templates/macros/implementations.jinja2
tmadlener: Makes sense. I think the cleanup is correct and since we run it through the sanitizers now this should be OK. I will check again locally to make sure that the tests are actually run.

tmadlener: OK. Turns out the cleanup is not OK, and we actually leak the pointer from `tmp`. Looking at the code what happens is the following:
- In the constructor for `tmp` we create a full copy of the big vector (that is managed by the collection)
https://github.com/AIDASoft/podio/blob/43330a8e1ea938f2c98b6473389bf317a6590ec3/python/templates/Obj.cc.jinja2#L32-L37
- In order to only get the slice that we want we create another vector with `new` (this snippet), but do not get rid of the copy we made in the first place

So we need to `delete` the original vector in `tmp` before we put in the new copy of the slice.

tmadlener: As far as I can tell from a quick look the `Obj(Obj const&)` copy constructor is only called via `clone`, so we could actually be a bit more clever with the unnecessary copying and only do the necessary work in `clone`.

tmadlener:
Can you rebase this to pick the changes from master?

jmcarcell:
After #588 and #589 I can easily test this and now the implementation doesn't have any leaks and does the minimum possible work; only copying the elements that are being pointed to. I was going to exclude the RNTuple test from the tests with sanitizers since it fails for me, but there are others that do so they will have to be changed anyway, I think the same will be seen when ROOT 6.30.06 is available.

python/templates/macros/implementations.jinja2
```diff
@@ -19,7 +19,24 @@
...
+{% else %}
+  auto tmp = new {{ type }}Obj();
+  tmp->id = podio::ObjectID();
+  tmp->data = m_obj->data;
+{% for relation in multi_relations %}
+  // If the current object has been read from a file, then the object may only have a slice of the relation vector
+  // so this slice has to be copied in case we want to modify it
+  tmp->m_{{ relation.name }}->reserve(m_obj->m_{{ relation.name }}->size());
+  for (size_t i = m_obj->data.{{ relation.name }}_begin; i < m_obj->data.{{ relation.name }}_end; i++) {
+    tmp->m_{{ relation.name }}->push_back((*m_obj->m_{{ relation.name }})[i]);
```
tmadlener, line 33:
```suggestion
    tmp->m_{{ relation.name }}->emplace_back((*m_obj->m_{{ relation.name }})[i]);
```
Might be able to skip one of the checks in `MabyeSharedPtr` if we get a move this way.

python/templates/macros/implementations.jinja2
```diff
@@ -19,7 +19,24 @@
 }
 
 Mutable{{ type }} {{ full_type }}::clone() const {
+{% if prefix %}
   return Mutable{{ type }}(podio::utils::MaybeSharedPtr(new {{ type }}Obj(*m_obj), podio::utils::MarkOwned));
+{% else %}
```
tmadlener, line 24: Do I understand correctly here that if we `clone` a `Mutable` type, we are guaranteed to not have slicing and skip the rest?

python/templates/macros/implementations.jinja2
```diff
@@ -19,7 +19,24 @@
...
+  auto tmp = new {{ type }}Obj();
+  tmp->id = podio::ObjectID();
+  tmp->data = m_obj->data;
```
tmadlener, line 27:
```suggestion
  auto tmp = new {{ type }}Obj(podio::ObjectID{}, m_obj->data);
```
This should be a public constructor that replaces **exactly these three lines**, i.e. it does not initialize any of the internal relations.

tests/unittests/unittest.cpp
```diff
@@ -1234,6 +1236,64 @@ void runCheckConsistencyTest(const std::string& filename) {
...
+  REQUIRE(nEmptyCluster.Hits().empty());
+  nEmptyCluster.addHits(hit);
+  REQUIRE(nEmptyCluster.Hits().size() == 1);
+  REQUIRE(nEmptyCluster.Hits()[0].cellID() == 420);
+  nEmptyCluster.addHits(anotherHit);
+  REQUIRE(nEmptyCluster.Hits().size() == 2);
+  REQUIRE(nEmptyCluster.Hits()[1].cellID() == 421);
+}
+
+TEST_CASE("Relations after cloning with TTrees", "[ASAN-FAIL][UBSAN-FAIL][relations]") {
```
tmadlener, line 1293:
```suggestion
TEST_CASE("Relations after cloning with TTrees", "[ASAN-FAIL][UBSAN-FAIL][relations][basics]") {
```
I would also add the `basics` tag here, since this is core functionality. Potentially even a `[memory_management]`.

Similar for the rntuple and sio based ones.

tmadlener (commented):
I just have a few minor comments / nitpicks.

The one thing that I don't see tested is whether the `clone`d objects can be written properly, but since they should now be properly initialized / constructed, they should behave just like any other freshly created object, so things should just work (TM).

tests/unittests/unittest.cpp
jmcarcell: Added

python/templates/macros/implementations.jinja2
jmcarcell: Changed

python/templates/macros/implementations.jinja2
jmcarcell: I think so, you can only get the pointers to the big vector when reading and then that's a non-mutable object which is now fixed so when you go to Mutable it should be fine. Previously I think it may have been possible by non-mutable (from reading) -> mutable (first clone) -> mutable (second clone) and the second clone would maybe be pointing to the big vector.

hegner:
Thanks! Quite important to fix this! If you fix the format check issues, I could merge it today

python/templates/macros/implementations.jinja2
jmcarcell: I'm not sure this can be a move, `[]` should return something movable no? In any case you don't want to move since you don't want to lose the original relation after a `clone()`

tmadlener: You are probably right, we won't get a move, but we should save creating a temporary for the `push_back`, and instead just get one copy constructor call.

veprbl:
This caused a regression for us https://github.com/AIDASoft/podio/issues/631

Issue #506, opened 2023-10-13
Cleanup CollectionReadBuffers


tmadlener:
This is a follow up to #502 which implemented a quick fix. Ideally the CollectionReadBuffers are refactored to a (move-only) type that clearly indicates the ownership of the contained buffers. Something like a `unique_ptr` with a custom (type-erased) deleter should work in this case, since we can populate the deleters at generation time.

tmadlener:
Enabling AddressSanitizer also reveals that there are a couple of indirect leaks in the whole machinery at the moment that could be related to this.

wdconinc:
I ran into AddressSanitizers leak warnings again related to this (well, caused by suppressions becoming invalid after #746)... Any further planning (excl. #813)? I'm actually surprised that our AddressSanitizers fail on leaks in `create*Buffers` (https://github.com/eic/EICrecon/actions/runs/17327690866/job/49196706907#step:7:2678), but the CI here doesn't flag it (https://github.com/AIDASoft/podio/actions/runs/17293775938/job/49086994087). Is it possible the ASAN_OPTIONS are too forgiving here? We use:
```
ASAN_OPTIONS: suppressions=/home/runner/work/EICrecon/EICrecon/.github/asan.supp:malloc_context_size=20:detect_leaks=1:verify_asan_link_order=0:detect_stack_use_after_return=1:detect_odr_violation=1:new_delete_type_mismatch=0:intercept_tls_get_addr=0
```

tmadlener:
Apologies, didn't intend to break anything downstream with that. We weren't aware you had those in your suppressions. I think for now we are through with cppcheck related fixes.

It's not entirely clear to me why at the moment why the CI flagging is different here and in your case. Our suppression basically amounts to LSAN only, we don't configure ASAN at the moment:

https://github.com/AIDASoft/podio/blob/5f5c6140c951970e06c3ec82dd0e6a15229df655/cmake/podioTest.cmake#L18

https://github.com/AIDASoft/podio/blob/5f5c6140c951970e06c3ec82dd0e6a15229df655/tests/root_io/leak_sanitizer_suppressions.txt#L2

And we essentially don't run any ROOT (TTree I/O) based tests when sanitizers are enabled:
https://github.com/AIDASoft/podio/blob/master/tests/CTestCustom.cmake

Let me try to configure ASAN similar to your setup here to see if things start to make more sense then.

wdconinc:
> Apologies, didn't intend to break anything downstream with that.

[...]

PR #585, opened 2024-04-15
Pass arguments by const reference to avoid making copies

jmcarcell:
BEGINRELEASENOTES
- Pass arguments by const reference to avoid making copies in a few places where passing by value was used

ENDRELEASENOTES

I think copies are not needed in these places

### DIFF
--- a/python/templates/macros/collections.jinja2
+++ b/python/templates/macros/collections.jinja2
@@ -206,7 +206,7 @@ podio::CollectionReadBuffers createBuffersV{{ schemaVersion }}(bool isSubset) {
 {% endfor %}
   }
 
-  readBuffers.createCollection = [](podio::CollectionReadBuffers buffers, bool isSubsetColl) {
+  readBuffers.createCollection = [](const podio::CollectionReadBuffers& buffers, bool isSubsetColl) {
     {{ collection_type }}Data data(buffers, isSubsetColl);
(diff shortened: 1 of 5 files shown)


python/templates/macros/collections.jinja2
```diff
@@ -206,7 +206,7 @@ podio::CollectionReadBuffers createBuffersV{{ schemaVersion }}(bool isSubset) {
 {% endfor %}
   }
 
-  readBuffers.createCollection = [](podio::CollectionReadBuffers buffers, bool isSubsetColl) {
+  readBuffers.createCollection = [](const podio::CollectionReadBuffers& buffers, bool isSubsetColl) {
```
tmadlener, line 209: I am not particularly liking this because it's a bit misleading, because something down the line will be stealing the data from these buffers:
https://github.com/AIDASoft/podio/blob/53ded81210d556638e4c50a761bc501ec2adf12e/python/templates/CollectionData.cc.jinja2#L35-L36

On the other hand this is the case already now because the copy we make at the moment will also have the same behavior. In the end all of this should be cleaned up better at some point (see https://github.com/AIDASoft/podio/issues/506).

Interestingly this also seems to work even though we technically assign to a different `std::function`: https://github.com/AIDASoft/podio/blob/53ded81210d556638e4c50a761bc501ec2adf12e/include/podio/CollectionBuffers.h#L51

tmadlener (approved):

PR #588, opened 2024-04-17
Set the right path for `PODIO_SIOBLOCK_PATH` and add an error message

jmcarcell:
BEGINRELEASENOTES
- Set the right path for `PODIO_SIOBLOCK_PATH`, fixing unit tests that read and write SIO files (there aren't any at the moment)
- Add an error message when `PODIO_SIOBLOCK_PATH` is set but none of the SioBlock libraries are found.

ENDRELEASENOTES

Adding a simple test that writes a frame to a SIO file won't work when ran with `ctest`. It looks like the environment is not correctly set because when running the test manually it will work fine. I don't know why but using `LD_LIBRARY_PATH` instead of `DL_PATHS` fixes this. This is the test

``` c++
+#if PODIO_ENABLE_SIO
+
+TEST_CASE("SIO writing a frame", "[relations]") {
+  auto frame = podio::Frame();
+  auto emptyColl = ExampleClusterCollection();
+  frame.put(std::move(emptyColl), "emptyClusters");
+  auto writer = podio::SIOWriter("unittests_frame_sio.sio");
+  writer.writeFrame(frame, podio::Category::Event);
+  writer.finish();
+}
+
+#endif
```

Here I don't think we need a unittest to simply test the environment, but for running with sanitizers, since all the ROOT ones fail then it's useful to have the SIO ones run from the unit tests.

### DIFF
--- a/tests/unittests/CMakeLists.txt
+++ b/tests/unittests/CMakeLists.txt
@@ -61,21 +61,11 @@ if (NOT FORCE_RUN_ALL_TESTS)
   endif()
 endif()
 
-option(SKIP_CATCH_DISCOVERY "Skip the Catch2 test discovery" OFF)
-
-# To work around https://github.com/catchorg/Catch2/issues/2424 we need the
-# DL_PATH argument for catch_discoer_tests which requires CMake 3.22 at least
-# The whole issue can be avoided if we skip the catch test discovery and set the
-# environment on our own
-if (CMAKE_VERSION VERSION_LESS 3.22)
-  set(SKIP_CATCH_DISCOVERY ON)
-endif()
-
-if (USE_SANITIZER MATCHES "Memory(WithOrigin)?" OR SKIP_CATCH_DISCOVERY)
+if (USE_SANITIZER MATCHES "Memory(WithOrigin)?")
   # Automatic test discovery fails with Memory sanitizers due to some issues in
   # Catch2. So in that case we skip the discovery step and simply run the thing
   # directly in the tests.
-  if (FORCE_RUN_ALL_TESTS OR SKIP_CATCH_DISCOVERY)
+  if (FORCE_RUN_ALL_TESTS)
     # Unfortunately Memory sanitizer seems to be really unhappy with Catch2 and
     # it fails to successfully launch the executable and execute any test. Here
     # we just include them in order to have them show up as failing
--- a/tests/unittests/CMakeLists.txt
+++ b/tests/unittests/CMakeLists.txt
@@ -90,7 +90,8 @@ else()
       TEST_SPEC ${filter_tests} # discover only tests that are known to not fail
       PROPERTIES
         ENVIRONMENT
+        PODIO_SIOBLOCK_PATH=${PROJECT_BINARY_DIR}/tests
+        ENVIRONMENT
         LD_LIBRARY_PATH=${CMAKE_CURRENT_BINARY_DIR}:${PROJECT_BINARY_DIR}/src:${PROJECT_BINARY_DIR}/tests:$<TARGET_FILE_DIR:ROOT::Tree>:$<$<TARGET_EXISTS:SIO::sio>:$<TARGET_FILE_DIR:SIO::sio>>:$ENV{LD_LIBRARY_PATH}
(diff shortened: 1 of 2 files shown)


tmadlener (commented):
I think I originally added the `DL_PATHS` like this because `catch_discover_tests` actually runs the executable during the build and it didn't find all the datamodel libraries without injecting the `DL_PATHS` like this.

[This is the original commit](https://github.com/AIDASoft/podio/pull/402/commits/cd509fa59b8229ff6f188f8593e67727a88c514e) that introduced it, it also contains a link to the original issue on Catch2.

At least for now the sanitizer builds all skip the test discovery because they run in an LCG environment where the CMake version is too old to have `DL_PATHS`. However, also on `dev3` with a newer environment things seem to work as expected **with** test discovery.

Did you check that the tests run in the correct environment with this change? I.e. `PODIO_SIOBLOCK_PATH` is also set? IIRC, there was also an issue with trying to set multiple environment variables this way with `catch_discover_tests`)

jmcarcell:
I noticed but since it works fine without `PODIO_SIOBLOCK_PATH` didn't give it much thought. Actually, based on `https://github.com/catchorg/Catch2/issues/2424`, it's possible to set several variables at the same time. But when doing that and setting both `LD_LIBRARY_PATH` and `PODIO_SIOBLOCK_PATH` the unittest crashes. So actually that tells us that the problem is not with `LD_LIBRARY_PATH` but with `PODIO_SIOBLOCK_PATH`, and indeed removing both also makes this new unittest pass (and all the others). Why do we need `PODIO_SIOBLOCK_PATH`? What do we want to do here? Since all the test pass without `PODIO_SIOBLOCK_PATH` I would just remove it.

jmcarcell:
So having a look I think the options are:
- Remove `PODIO_SIOBLOCK_PATH` and just use `LD_LIBRARY_PATH`. From the release notes it seems it was introduced when having multiple installations but then `LD_LIBRARY_PATH` also needs to be changed anyway.
- If `PODIO_SIOBLOCK_PATH` is not removed, a check should be added to not crash when the library is not found. Then the path can be modified since the current one is wrong and both `PODIO_SIOBLOCK_PATH` and `LD_LIBRARY_PATH` are set.

tests/unittests/CMakeLists.txt
```diff
@@ -88,9 +88,10 @@ else()
       WORKING_DIRECTORY ${CMAKE_CURRENT_LIST_DIR}
       TEST_PREFIX "UT_" # make it possible to filter easily with -R ^UT
       TEST_SPEC ${filter_tests} # discover only tests that are known to not fail
-      DL_PATHS ${CMAKE_CURRENT_BINARY_DIR}:${PROJECT_BINARY_DIR}/src:${PROJECT_BINARY_DIR}/tests:$<TARGET_FILE_DIR:ROOT::Tree>:$<$<TARGET_EXISTS:SIO::sio>:$<TARGET_FILE_DIR:SIO::sio>>:$ENV{LD_LIBRARY_PATH}
       PROPERTIES
         ENVIRONMENT
-        PODIO_SIOBLOCK_PATH=${CMAKE_CURRENT_BINARY_DIR}
+        PODIO_SIOBLOCK_PATH=${PROJECT_BINARY_DIR}/tests
+        ENVIRONMENT
```
tmadlener, line 94: Is this second `ENVIRONMENT` necessary?

tmadlener (commented):
Thanks for fixing this. It looks like I messed this up a bit in the whole test reorganization.

If we remove the `DL_PATHS`, I think we also need to adapt this comment here to not create too much confusion:

https://github.com/AIDASoft/podio/blob/82814a8800e246af71077982e270672b71ed6531/tests/unittests/CMakeLists.txt#L66-L72

tests/unittests/CMakeLists.txt
jmcarcell: Having a look at the comment, I think the `SKIP_CATCH_DISCOVERY` option can be removed; the point of that part is to check if the cmake version is less than 3.22 because in principle using `DL_PATH` is needed and this is supported only for 3.22 and above. But, it is actually not needed, because duplicating the `ENVIRONMENT` keyword works (it's described in the Catch2 issue).

tmadlener (approved):

PR #591, opened 2024-05-02
CI: use different dev4 stack

andresailer:
dev4 centos7 is now longer updating ROOT, so this gives misleading results

BEGINRELEASENOTES
- CI: alma9-gcc13 dev stack

ENDRELEASENOTES


tmadlener (approved):

hegner:
OK. Then let's pick it up :-)

PR #592, opened 2024-05-02
[WIP] Enable RNTuple for the sanitizer builds

tmadlener:
BEGINRELEASENOTES
- Enable RNTuple for the sanitizer CI builds.

ENDRELEASENOTES

These are currently not enabled by default. There is a leak in libCling.so where we don't have too much control over. I am not sure if we gain too much other than the information about which parts can actually run under sanitized builds (and which can't)


tmadlener:
Superseded by #695

PR #593, opened 2024-05-03
Add an RDataSource for podio files and collections

kjvbrt:
BEGINRELEASENOTES
- Add a `podio::DataSource` as an `RDataSource` for working with podio (generated EDM) files and collections. This `RDataSource` exposes the podio collections directly as columns for `RDataFrames`. **We consider this fairly stable, but still in early stages, there might still be some (breaking) changes to this**

ENDRELEASENOTES

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,6 +72,7 @@ option(CREATE_DOC        "Whether or not to create doxygen doc target." OFF)
 option(ENABLE_SIO        "Build SIO I/O support" OFF)
 option(PODIO_RELAX_PYVER "Do not require exact python version match with ROOT" OFF)
 option(ENABLE_RNTUPLE    "Build with support for the new ROOT NTtuple format" OFF)
+option(ENABLE_DATASOURCE "Build podio's ROOT DataSource" OFF)
 option(PODIO_USE_CLANG_FORMAT "Use clang-format to format the code" OFF)
 option(ENABLE_JULIA      "Enable Julia support. When enabled, Julia datamodels will be generated, and Julia tests will run." OFF)
 
@@ -79,14 +80,17 @@ option(ENABLE_JULIA      "Enable Julia support. When enabled, Julia datamodels w
 #--- Declare ROOT dependency ---------------------------------------------------
 list(APPEND CMAKE_PREFIX_PATH $ENV{ROOTSYS})
 set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
-if(NOT ENABLE_RNTUPLE)
-  find_package(ROOT REQUIRED COMPONENTS RIO Tree ROOTDataFrame)
-else()
-  find_package(ROOT REQUIRED COMPONENTS RIO Tree ROOTNTuple)
+set(root_components_needed RIO Tree)
+if(ENABLE_RNTUPLE)
+  list(APPEND root_components_needed ROOTNTuple)
   if(${ROOT_VERSION} VERSION_LESS 6.28.02)
     message(FATAL_ERROR "You are trying to build podio with support for the new ROOT NTuple format, but your ROOT version is too old. Please update ROOT to at least version 6.28.02")
   endif()
 endif()
+if(ENABLE_DATASOURCE)
+  list(APPEND root_components_needed ROOTDataFrame)
+endif()
+find_package(ROOT REQUIRED COMPONENTS ${root_components_needed})
 
 # ROOT_CXX_STANDARD was introduced in https://github.com/root-project/root/pull/6466
 # before that it's an empty variable so we check if it's any number > 0
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,7 +77,7 @@ option(ENABLE_JULIA      "Enable Julia support. When enabled, Julia datamodels w
 list(APPEND CMAKE_PREFIX_PATH $ENV{ROOTSYS})
 set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 if(NOT ENABLE_RNTUPLE)
-  find_package(ROOT REQUIRED COMPONENTS RIO Tree)
+  find_package(ROOT REQUIRED COMPONENTS RIO Tree ROOTDataFrame)
 else()
   find_package(ROOT REQUIRED COMPONENTS RIO Tree ROOTNTuple)
   if(${ROOT_VERSION} VERSION_LESS 6.28.02)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,6 +64,9 @@ option(FORCE_RUN_ALL_TESTS "Run all the tests even those with known problems" OF
 option(CLANG_TIDY "Run clang-tidy after compilation." OFF)
 ADD_CLANG_TIDY()
 
+# Export compile commands --- used by the tools from clang family
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+
 #--- Declare options -----------------------------------------------------------
 option(CREATE_DOC        "Whether or not to create doxygen doc target." OFF)
 option(ENABLE_SIO        "Build SIO I/O support" OFF)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,9 +64,6 @@ option(FORCE_RUN_ALL_TESTS "Run all the tests even those with known problems" OF
 option(CLANG_TIDY "Run clang-tidy after compilation." OFF)
 ADD_CLANG_TIDY()
 
-# Export compile commands --- used by the tools from clang family
-set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
-
 #--- Declare options -----------------------------------------------------------
 option(CREATE_DOC        "Whether or not to create doxygen doc target." OFF)
 option(ENABLE_SIO        "Build SIO I/O support" OFF)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,14 +83,14 @@ set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 set(root_components_needed RIO Tree)
 if(ENABLE_RNTUPLE)
   list(APPEND root_components_needed ROOTNTuple)
-  if(${ROOT_VERSION} VERSION_LESS 6.28.02)
-    message(FATAL_ERROR "You are trying to build podio with support for the new ROOT NTuple format, but your ROOT version is too old. Please update ROOT to at least version 6.28.02")
-  endif()
 endif()
 if(ENABLE_DATASOURCE)
   list(APPEND root_components_needed ROOTDataFrame)
 endif()
 find_package(ROOT REQUIRED COMPONENTS ${root_components_needed})
+if((ENABLE_RNTUPLE) AND (${ROOT_VERSION} VERSION_LESS 6.28.02))
+    message(FATAL_ERROR "You are trying to build podio with support for the new ROOT NTuple format, but your ROOT version is too old. Please update ROOT to at least version 6.28.02")
+endif()
 
 # ROOT_CXX_STANDARD was introduced in https://github.com/root-project/root/pull/6466
 # before that it's an empty variable so we check if it's any number > 0
--- a/include/podio/DataSource.h
+++ b/include/podio/DataSource.h
@@ -18,64 +18,61 @@
 #include <vector>
 
 namespace podio {
-class ROOTDataSource : public ROOT::RDF::RDataSource {
+class DataSource : public ROOT::RDF::RDataSource {
 public:
   ///
-  /// @brief Construct the podio::ROOTDataSource from the provided file.
+  /// @brief Construct the podio::DataSource from the provided file.
   ///
-  explicit ROOTDataSource(const std::string& filePath, int nEvents = -1);
+  explicit DataSource(const std::string& filePath, int nEvents = -1);
 
   ///
-  /// @brief Construct the podio::ROOTDataSource from the provided file
-  ///        list.
+  /// @brief Construct the podio::DataSource from the provided file list.
   ///
-  explicit ROOTDataSource(const std::vector<std::string>& filePathList, int nEvents = -1);
+  explicit DataSource(const std::vector<std::string>& filePathList, int nEvents = -1);
 
   ///
-  /// @brief Inform the podio::ROOTDataSource of the desired level of
-  ///        parallelism.
+  /// @brief Inform the podio::DataSource of the desired level of parallelism.
   ///
   void SetNSlots(unsigned int nSlots) override;
 
   ///
-  /// @brief Retrieve from podio::ROOTDataSource per-thread readers for the
-  ///        desired columns.
+  /// @brief Retrieve from podio::DataSource per-thread readers for the desired
+  /// columns.
   ///
   template <typename T>
   std::vector<T**> GetColumnReaders(std::string_view columnName);
 
   ///
-  /// @brief Inform podio::ROOTDataSource that an event-loop is about to
-  ///        start.
+  /// @brief Inform podio::DataSource that an event-loop is about to start.
   ///
   void Initialize() override;
 
   ///
-  /// @brief Retrieve from podio::ROOTDataSource a set of ranges of entries
-  ///        that can be processed concurrently.
+  /// @brief Retrieve from podio::DataSource a set of ranges of entries that
+  ///        can be processed concurrently.
   ///
   std::vector<std::pair<ULong64_t, ULong64_t>> GetEntryRanges() override;
 
   ///
-  /// @brief Inform podio::ROOTDataSource that a certain thread is about to
-  ///        start working on a certain range of entries.
+  /// @brief Inform podio::DataSource that a certain thread is about to start
+  ///        working on a certain range of entries.
   ///
   void InitSlot(unsigned int slot, ULong64_t firstEntry) override;
 
   ///
-  /// @brief Inform podio::ROOTDataSource that a certain thread is about to
-  ///        start working on a certain entry.
+  /// @brief Inform podio::DataSource that a certain thread is about to start
+  ///        working on a certain entry.
   ///
   bool SetEntry(unsigned int slot, ULong64_t entry) override;
 
   ///
-  /// @brief Inform podio::ROOTDataSource that a certain thread finished
-  ///        working on a certain range of entries.
+  /// @brief Inform podio::DataSource that a certain thread finished working
+  ///        on a certain range of entries.
   ///
   void FinalizeSlot(unsigned int slot) override;
 
   ///
-  /// @brief Inform podio::ROOTDataSource that an event-loop finished.
+  /// @brief Inform podio::DataSource that an event-loop finished.
   ///
   void Finalize() override;
 
@@ -141,7 +138,7 @@ private:
   std::vector<std::unique_ptr<podio::Frame>> m_frames = {};
 
   ///
-  /// @brief Setup input for the podio::ROOTDataSource.
+  /// @brief Setup input for the podio::DataSource.
   ///
   /// @param[in]  Number of events.
   /// @return     void.
@@ -153,8 +150,8 @@ private:
 /// Not used.
 ///
 template <typename T>
-std::vector<T**> ROOTDataSource::GetColumnReaders(std::string_view) {
-  // std::cout << "podio::ROOTDataSource: Getting column readers for column: " << columnName << std::endl;
+std::vector<T**> DataSource::GetColumnReaders(std::string_view) {
+  // std::cout << "podio::DataSource: Getting column readers for column: " << columnName << std::endl;
 
   std::vector<T**> readers;
 
--- a/include/podio/DataSource.h
+++ b/include/podio/DataSource.h
@@ -36,13 +36,6 @@ public:
   void SetNSlots(unsigned int nSlots) override;
 
   ///
-  /// @brief Retrieve from podio::DataSource per-thread readers for the desired
-  /// columns.
-  ///
-  template <typename T>
-  std::vector<T**> GetColumnReaders(std::string_view columnName);
-
-  ///
   /// @brief Inform podio::DataSource that an event-loop is about to start.
   ///
   void Initialize() override;
@@ -111,7 +104,7 @@ private:
   std::vector<std::string> m_filePathList = {};
 
   /// Total number of events
-  unsigned int m_nEvents = 0;
+  ULong64_t m_nEvents = 0;
 
   /// Ranges of events available to be processed
   std::vector<std::pair<ULong64_t, ULong64_t>> m_rangesAvailable = {};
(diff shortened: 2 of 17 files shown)


include/podio/ROOTDataSource.h
```diff
@@ -0,0 +1,117 @@
...
+      std::vector<std::unique_ptr<podio::Frame>> m_frames = {};
+      /// Setup input
+      void SetupInput(int nEvents);
+  };
+
+
+  /**
+   * \brief Retrieve from podio::ROOTDataSource per-thread readers for the
+   *        desired columns.
+   */
```
tmadlener, line 87: Just for some nitpicking to start off. I have (tried) to switch all of podios docstrings to using `///` and the `@` versions of the doxygen tags. Additionally, I think these need to go the declaration in the class for them to be picked up properly by doxygen.

src/ROOTDataSource.cc
```diff
@@ -0,0 +1,401 @@
...
+      // }
+
+      // Check if the provided file contains required metadata
+      std::unique_ptr<TFile> inFile(TFile::Open(filePath.data(), "READ"));
+      auto metadata = inFile->Get("podio_metadata");
+      if (!metadata) {
+        throw std::runtime_error(
+            "podio::ROOTDataSource: Provided file is missing podio metadata!");
+      }
+    }
```
tmadlener: This check is already done in `ROOTReader::openFile` and I think we can remove it here.

include/podio/ROOTDataSource.h
```diff
@@ -0,0 +1,117 @@
...
+  class ROOTDataSource : public ROOT::RDF::RDataSource {
+    public:
+      explicit ROOTDataSource(const std::string& filePath, int nEvents = -1);
+      explicit ROOTDataSource(const std::vector<std::string>& filePathList,
+                              int nEvents = -1);
+
+      void SetNSlots(unsigned int nSlots) override;
+
+      template<typename T>
+      std::vector<T**> GetColumnReaders(std::string_view columnName);
```
tmadlener, line 29: If I understand the documentation of `RDataSource` correctly, we don't need this at all. Also this is not overriding anything from the `RDataSource` base class. I think all of the "magic" is happening in the `GetColumnReadersImpl`, right?

src/ROOTDataSource.cc
```diff
@@ -0,0 +1,401 @@
...
+
+    for (size_t i = 0; i < m_nSlots; ++i) {
+      m_podioReaders[i]->openFiles(m_filePathList);
+    }
+
+    for (size_t i = 0; i < m_nSlots; ++i) {
+      m_frames.emplace_back(
+          std::make_unique<podio::Frame>(
+              podio::Frame(m_podioReaders[i]->readEntry("events", 0))));
+    }
```
tmadlener, line 157: Aren't we reading the same event `nSlots` times this way? As far as I can see we read the correct entry in any case later in `SetEntry`. Can we not just simply put an empty Frame here? Or even just populate the vector with a bunch of `nullptr`?

tmadlener (commented):
Thanks @kjvbrt. I wasn't aware this was possible in this generality. Very nice. I am probably still missing quite a few things, but I tried to summarize my current understanding below and added and a few comments / questions inline

For my general understanding, what this does is basically:
- Create a reader for each thread  / slot
- Keep a vector of Frames, again one per thread / slot
- Populate a vector of collections per thread / slot, where
  - each collection always is at the same position, so that we can simply index into it

I am wondering whether we could do with just one reader instead of multiple, but I suppose then we would probably have to do our own locking around that, because the `RDataSource` doesn't do that for us?

If I understand correctly, we could in the end even replace the current `ROOTReader` with the `Reader` that is introduced in #522 and effectively support all backends + RDataFrame?

include/podio/ROOTDataSource.h
kjvbrt: Done, all the doc strings moved to header.

tmadlener:
The sanitizer workflows are most likely failing because the tests that produce the input files are not running. You can disable them by adding the new tests also to the list in [`CTestCustom.cmake`](https://github.com/AIDASoft/podio/blob/master/tests/CTestCustom.cmake)

src/ROOTDataSource.cc
```diff
@@ -0,0 +1,308 @@
...
+  }
+
+  // Create probing frame
+  podio::Frame frame;
+  unsigned int nEventsInFiles = 0;
+  podio::ROOTReader podioReader;
+  podioReader.openFiles(m_filePathList);
+  nEventsInFiles = podioReader.getEntries("events");
+  frame = podio::Frame(podioReader.readEntry("events", 0));
```
Zehvogel, line 59:
```suggestion
  nEventsInFiles = podioReader.getEntries(podio::Category::Event);
  frame = podio::Frame(podioReader.readEntry(podio::Category::Event, 0));
```

src/ROOTDataSource.cc
```diff
@@ -0,0 +1,308 @@
...
+void ROOTDataSource::InitSlot([[maybe_unused]] unsigned int slot, [[maybe_unused]] ULong64_t firstEntry) {
+  // std::cout << "podio::ROOTDataSource: Initializing slot: " << slot
+  //           << " with first entry " << firstEntry << std::endl;
+}
+
+bool ROOTDataSource::SetEntry(unsigned int slot, ULong64_t entry) {
+  // std::cout << "podio::ROOTDataSource: In slot: " << slot << ", setting entry: "
+  //           << entry << std::endl;
+
+  m_frames[slot] = std::make_unique<podio::Frame>(podio::Frame(m_podioReaders[slot]->readEntry("events", entry)));
```
Zehvogel, line 186:
```suggestion
  m_frames[slot] = std::make_unique<podio::Frame>(podio::Frame(m_podioReaders[slot]->readEntry(podio::Category::Event, entry)));
```

src/ROOTDataSource.cc
```diff
@@ -0,0 +1,308 @@
...
+#include <filesystem>
+#include <iostream>
+#include <memory>
+
+// ROOT
+#include <TFile.h>
+
+// podio
+#include <podio/Frame.h>
+#include <podio/ROOTReader.h>
```
Zehvogel, line 16:
```suggestion
#include <podio/ROOTReader.h>
#include <podio/FrameCategories.h>
```

Zehvogel (changes requested):
to avoid the hardcoded strings :)

src/ROOTDataSource.cc
```diff
@@ -0,0 +1,308 @@
...
+  std::vector<std::string> collNames = frame.getAvailableCollections();
+  // std::cout << "podio::ROOTDataSource: Found following collections:\n";
+  for (auto& collName : collNames) {
+    const podio::CollectionBase* coll = frame.get(collName);
+    if (coll->isValid()) {
+      m_columnNames.emplace_back(collName);
+      m_columnTypes.emplace_back(coll->getValueTypeName());
+      // std::cout << "                - " << collName << "\n";
+    }
+  }
```
Zehvogel, line 98: Can we also make the parameters of the frame available as columns, e.g. to access the cross section? :)

include/podio/ROOTDataSource.h
```diff
@@ -0,0 +1,178 @@
...
+  std::vector<std::string> m_columnTypes = {};
+
+  /// Collections, m_Collections[columnIndex][slotIndex]
+  std::vector<std::vector<const podio::CollectionBase*>> m_Collections = {};
+
+  /// Active collections
+  std::vector<unsigned int> m_activeCollections = {};
+
+  /// Root podio readers
+  std::vector<std::unique_ptr<podio::ROOTReader>> m_podioReaders = {};
```
tmadlener, line 137:
```suggestion
  std::vector<std::unique_ptr<podio::Reader>> m_podioReaders = {};
```
This should now be possible (with `#include "podio/Reader.h"`)

src/ROOTDataSource.cc
```diff
@@ -0,0 +1,308 @@
...
+    //                            + filePath + "\" does not exist!");
+    // }
+
+    // Check if the provided file contains required metadata
+    std::unique_ptr<TFile> inFile(TFile::Open(filePath.data(), "READ"));
+    auto metadata = inFile->Get("podio_metadata");
+    if (!metadata) {
+      throw std::runtime_error("podio::ROOTDataSource: Provided file is missing podio metadata!");
+    }
+  }
```
tmadlener, line 51: If we go for the `Reader` interface, this check can go, as the `Reader` will take care of this.

src/ROOTDataSource.cc
```diff
@@ -0,0 +1,308 @@
...
+
+  // Create probing frame
+  podio::Frame frame;
+  unsigned int nEventsInFiles = 0;
+  podio::ROOTReader podioReader;
+  podioReader.openFiles(m_filePathList);
```
tmadlener, line 57:
```suggestion
  auto podioReader = podio::makeReader(m_filePathList);
```
with the reader interface.

tmadlener (commented):
A few new minor suggestions for using the reader interface, now that it's available. I think I got all the places, but I might have missed some.

include/podio/ROOTDataSource.h
```diff
@@ -0,0 +1,117 @@
...
+   * \brief Retrieve from podio::ROOTDataSource per-thread readers for the
+   *        desired columns.
+   */
+  template<typename T>
+  std::vector<T**>
+  ROOTDataSource::GetColumnReaders(std::string_view columnName) {
+    std::cout << "podio::ROOTDataSource: Getting column readers for column: "
+              << columnName << std::endl;
+
+    std::vector<T**> readers;
```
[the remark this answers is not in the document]
kjvbrt: This is not used, I believe it's old way now.

src/ROOTDataSource.cc
```diff
@@ -0,0 +1,401 @@
...
+#include <podio/ROOTReader.h>
+
+bool loadROOTDataSource() {
+  return true;
+}
+
+namespace podio {
+  /**
+   * \brief Construct the podio::ROOTDataSource from the provided file.
+   */
```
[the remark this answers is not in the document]
kjvbrt: Moved all doxygen documentation to header file.

src/ROOTDataSource.cc
```diff
@@ -0,0 +1,401 @@
...
+  }
+
+
+  /**
+   * \brief Inform podio::ROOTDataSource that a certain thread is about to start working
+   *        on a certain range of entries.
+   */
+  void
+  ROOTDataSource::InitSlot([[maybe_unused]] unsigned int slot,
+                       [[maybe_unused]] ULong64_t firstEntry) {
```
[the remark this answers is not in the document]
kjvbrt: Using the trick...

src/ROOTDataSource.cc
```diff
@@ -0,0 +1,401 @@
...
+    ROOT::RDataFrame rdf(std::make_unique<ROOTDataSource>(filePathList));
+
+    return rdf;
+  }
+
+  ROOT::RDataFrame
+  CreateDataFrame(const std::string& filePath) {
+    std::vector<std::string> filePathList;
+    filePathList.emplace_back(filePath);
+    ROOT::RDataFrame rdf(std::make_unique<ROOTDataSource>(filePathList));
```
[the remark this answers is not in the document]
kjvbrt: Now there are two constructors...

src/ROOTDataSource.cc
```diff
@@ -0,0 +1,401 @@
...
+                         columnName);
+    if (itr != m_columnNames.end()) {
+      auto i = std::distance(m_columnNames.begin(), itr);
+      // std::cout << "podio::ROOTDataSource: Found type name: "
+      //           << m_columnTypes.at(i) << std::endl;
+
+      return m_columnTypes.at(i) + "Collection";
+    }
+
+    return "float";
```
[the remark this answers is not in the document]
kjvbrt: Throwing exception...

include/podio/ROOTDataSource.h
```diff
@@ -0,0 +1,178 @@
+#ifndef PODIO_DATASOURCE_H__
+#define PODIO_DATASOURCE_H__
+
+// STL
+#include <string>
+#include <vector>
+
+// ROOT
+#include <ROOT/RDataFrame.hxx>
+#include <ROOT/RDataSource.hxx>
```
tmadlener, line 10: STL should go all the way to the bottom, ROOT above them, such that the podio headers are first.

include/podio/ROOTDataSource.h
```diff
@@ -0,0 +1,178 @@
...
+
+// Podio
+#include <podio/CollectionBase.h>
+#include <podio/Frame.h>
+#include <podio/ROOTReader.h>
+
+namespace podio {
+using Record_t = std::vector<void*>;
```
tmadlener, line 18: Does this have to be so "public", it's a rather general name and unless it's necessary from the ROOT side, this looks like we could move it into a member type of `ROOTDataSource`.

src/ROOTDataSource.cc
```diff
@@ -0,0 +1,308 @@
...
+  //           << std::endl;
+
+  // Get collections stored in the files
+  std::vector<std::string> collNames = frame.getAvailableCollections();
+  // std::cout << "podio::ROOTDataSource: Found following collections:\n";
+  for (auto& collName : collNames) {
+    const podio::CollectionBase* coll = frame.get(collName);
+    if (coll->isValid()) {
+      m_columnNames.emplace_back(collName);
+      m_columnTypes.emplace_back(coll->getValueTypeName());
```
tmadlener, line 95:
```suggestion
      m_columnTypes.emplace_back(coll->getTypeName());
```
Since the only usage below is with `+ "Collection"`, we could just store the complete type name, including the collection here.

src/CMakeLists.txt
```diff
@@ -105,7 +107,12 @@ if(ENABLE_RNTUPLE)
 endif()
 
 PODIO_ADD_LIB_AND_DICT(podioRootIO "${root_headers}" "${root_sources}" root_selection.xml)
-target_link_libraries(podioRootIO PUBLIC podio::podio ROOT::Core ROOT::RIO ROOT::Tree ROOT::ROOTVecOps)
+target_link_libraries(podioRootIO PUBLIC podio::podio
+                                         ROOT::Core
+                                         ROOT::RIO
+                                         ROOT::Tree
+                                         ROOT::ROOTVecOps
+                                         ROOT::ROOTDataFrame)
```
tmadlener, line 115: I am contemplating whether we should move the `ROOTDataSource` into a separate library instead of making it part of the `podioRootIO` one. The main argument against it is that we will have one more library. However, if we really make use of the `Reader` interface internally, then we would need to link `podioRootIO` against `podioIO` (making a cyclic lib dependency) and potentially also against `podioSioIO`. Then we would have no separation of the libraries at all any longer. 

Since it is probably easier to merge libraries down the line, than splitting them apart, I would suggest to leave the `podioRootIO` library untouched and instead make a new `podioRDataSource` library that links against `podio::podioIO` (and all the necessary things from ROOT).

include/podio/ROOTDataSource.h
kjvbrt: done

include/podio/ROOTDataSource.h
kjvbrt: using plain `std::vector<void*>` now.

src/ROOTDataSource.cc
kjvbrt: Done

src/ROOTDataSource.cc
kjvbrt: Done

src/ROOTDataSource.cc
kjvbrt: Done

src/ROOTDataSource.cc
kjvbrt: Good point

include/podio/ROOTDataSource.h
kjvbrt: Done

src/ROOTDataSource.cc
kjvbrt: Done

src/ROOTDataSource.cc
kjvbrt: Done

src/CMakeLists.txt
kjvbrt: Yes, I had to link against `podio::podioIO` in couple of places. Will create separate library...

src/ROOTDataSource.cc
kjvbrt: Removed

kjvbrt:
I continue with testing the changes made here in FCCAnalyses

src/DataSource.cc
```diff
@@ -0,0 +1,188 @@
...
+  // Initialize set of addresses needed
+  m_Collections.resize(m_columnNames.size(), std::vector<const podio::CollectionBase*>(m_nSlots, nullptr));
+
+  // Initialize podio readers
+  for (size_t i = 0; i < m_nSlots; ++i) {
+    m_podioReaders.emplace_back(std::make_unique<podio::Reader>(podio::makeReader(m_filePathList)));
+  }
+
+  for (size_t i = 0; i < m_nSlots; ++i) {
+    m_frames.emplace_back(std::make_unique<podio::Frame>(podio::Frame{}));
```
andresailer, line 92:
```suggestion
    m_frames.emplace_back(std::make_unique<podio::Frame>());
```
make_unique forwards to c'tor, so there is no need to first create a frame and then make make_unique call the copy c'tor ?

include/podio/DataSource.h
```diff
@@ -0,0 +1,179 @@
...
+  /// @brief Setup input for the podio::DataSource.
+  ///
+  /// @param[in]  Number of events.
+  /// @return     void.
+  ///
+  void SetupInput(int nEvents);
+};
+
+///
+/// Not used.
```
andresailer, line 150: If this is not used, can you explain why it is needed to be here?

include/podio/DataSource.h
```diff
@@ -0,0 +1,179 @@
...
+
+private:
+  /// Number of slots/threads
+  unsigned int m_nSlots = 1;
+
+  /// Input filename
+  std::vector<std::string> m_filePathList = {};
+
+  /// Total number of events
+  unsigned int m_nEvents = 0;
```
andresailer, line 114:
```suggestion
  ULong64_t m_nEvents = 0;
```
Shouldn't this match with the ranges defined in the next lines?

CMakeLists.txt
```diff
@@ -64,6 +64,9 @@ option(FORCE_RUN_ALL_TESTS "Run all the tests even those with known problems" OF
 option(CLANG_TIDY "Run clang-tidy after compilation." OFF)
 ADD_CLANG_TIDY()
 
+# Export compile commands --- used by the tools from clang family
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+
```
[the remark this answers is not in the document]
kjvbrt: OK, removed :)

include/podio/ROOTDataSource.h
```diff
@@ -0,0 +1,182 @@
...
+
+protected:
+  ///
+  /// @brief Type-erased vector of pointers to pointers to column
+  ///        values --- one per slot.
+  ///
+  std::vector<void*> GetColumnReadersImpl(std::string_view name, const std::type_info& typeInfo) override;
+
+  std::string AsString() override {
+    return "Podio data source";
```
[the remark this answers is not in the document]
kjvbrt: Seems to be fixed by `clang-format`

src/CMakeLists.txt
```diff
@@ -160,18 +161,57 @@ if(ENABLE_SIO)
...
+     RELATIVE ${PROJECT_SOURCE_DIR}
+     "${PROJECT_SOURCE_DIR}/install/podio/*.h")
+
+if (NOT ENABLE_SIO)
+  list(FILTER headers_necessary EXCLUDE REGEX SIO.*\\.h$)
+endif()
+if (NOT ENABLE_RNTUPLE)
+  list(FILTER headers_necessary EXCLUDE REGEX RNTuple.*\\.h$)
 endif()
+if (NOT ENABLE_RNTUPLE)
```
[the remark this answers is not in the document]
kjvbrt: Yup

src/DataSource.cc
```diff
@@ -0,0 +1,188 @@
...
+
+  // Check if the provided file(s) exists and contain required metadata is done
+  // by podio::Reader
+
+  // Create probing frame
+  podio::Frame frame;
+  unsigned int nEventsInFiles = 0;
+  auto podioReader = podio::makeReader(m_filePathList);
+  nEventsInFiles = podioReader.getEntries(podio::Category::Event);
+  frame = podio::Frame(podioReader.readFrame(podio::Category::Event, 0));
```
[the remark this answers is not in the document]
kjvbrt: Thx, removed

src/DataSource.cc
kjvbrt: Thx, removed.

include/podio/DataSource.h
kjvbrt: Not needed at all in the newer versions of ROOT, completely removed. This method was part of the old RDataSource interface, if I'm not mistaken.

include/podio/DataSource.h
kjvbrt: Changed, thx.

src/CMakeLists.txt
```diff
@@ -160,18 +161,57 @@ if(ENABLE_SIO)
...
 
 # Only install the necessary headers
-if (ENABLE_SIO)
-  install(DIRECTORY ${PROJECT_SOURCE_DIR}/include/podio DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}")
-else()
-  install(DIRECTORY ${PROJECT_SOURCE_DIR}/include/podio DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
-    REGEX SIO.*\\.h$ EXCLUDE )
+file(GLOB headers_necessary
+     RELATIVE ${PROJECT_SOURCE_DIR}
+     "${PROJECT_SOURCE_DIR}/install/podio/*.h")
```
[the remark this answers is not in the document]
kjvbrt: removed RELATIVE :)

tmadlener (commented):
As far as I can tell this is ready, right? From which ROOT version onwards should this work? Then we could add a simple check in CMake (similar to RNTuple).

I will push the minor code changes immediately. Once we have this, we can the add support for parameters in a follow-up PR, I would suggest.

src/ROOTDataSource.cc
tmadlener: Will do this in a follow up PR.

kjvbrt:
I tested this in FCCAnalysis, seems to work...

CMakeLists.txt
```diff
@@ -69,20 +69,24 @@ option(CREATE_DOC        "Whether or not to create doxygen doc target." OFF)
...
-  if(${ROOT_VERSION} VERSION_LESS 6.28.02)
+set(root_components_needed RIO Tree)
+if(ENABLE_RNTUPLE)
+  list(APPEND root_components_needed ROOTNTuple)
+endif()
+if(ENABLE_DATASOURCE)
+  list(APPEND root_components_needed ROOTDataFrame)
+endif()
+find_package(ROOT REQUIRED COMPONENTS ${root_components_needed})
+if((ENABLE_RNTUPLE) AND (${ROOT_VERSION} VERSION_LESS 6.28.02))
```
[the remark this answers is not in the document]
kjvbrt: The RDataSource is in ROOT for a long time (not experimental since ROOT 6.14), I tested this in nightlies stack (6.32)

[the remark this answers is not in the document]
tmadlener: Alright, then I think we can also skip the check here. Thanks for checking.

tmadlener (approved):

Issue #150, opened 2020-11-23
CollectionIterator does not fully satisfy iterator concept


tmadlener:
It currently seems to be impossible to use podio collections with the STL algorithms operating on iterator ranges, because the `CollectionIterator` does not provide enough information, leading the compiler complaining about (greatly abbreviated to the crucial line):

```
/usr/include/c++/9/bits/stl_iterator_base_types.h:205:5: error: no type named ‘iterator_category’ in ‘struct std::iterator_traits<edm4hep::MCRecoParticleAssociationCollectionIterator>’
```

I think having a correctly implemented const-iterator is something we should aim for, to at least enable the usage of STL algorithms that can be used on ranges of const-iterators.

whit2333:
IMO, this is critical. I have almost given up and started using the raw `*Data` classes which are not accessible otherwise.

PR #598, opened 2024-05-07
Document 'Collection' comparison with 'Container' named requirement

### LINKED ISSUE #479: Check podio Collections for C++ Container compatibility
Document which parts of the [requirements](https://en.cppreference.com/w/cpp/named_req/Container) are fulfilled in PODIO Collections and document it properly. This is related to #478 
Add whatever can be added.

### LINKED ISSUE #546: collection push_back and value_type incompatible with some std iterator adaptors
The std adaptors often rely on a typedef `value_type`. The PODIO mutable semantics complicate matters, as `Collection::value_type`  is simply defined as an immutable type.

Collection `push_back` has overloads for both mutable and immutable types. The overloads have different behaviors, for instance, for `Collection::value_type`  overload the non-subset collections throws:

> invalid_argument: Can only add immutable objects to subset collections

While itself it's correct behavior, some std adapters do implicit conversion to `value_type` and end up using the overload for immutable type, e.g copying elements with`std::back_insert_iterator`:

```cpp
// this throws because the immutable type is pushed
std::transform(std::begin(a), std::end(a), std::back_inserter(b),
[](const ExampleHit& i) -> MutableExampleHit {
      return i.clone();
});

// this doesn't throw since the mutable type is pushed
std::for_each(std::begin(a), std::end(a), [&b](const auto& i) {
      b.push_back(i.clone());
});
```

I'm not sure if there are any possible solutions that don't change the `value_type` or `push_back` semantics

m-fila:
BEGINRELEASENOTES
- Added documentation on collection compatibility with container named requirement
- Added container-like methods `cbegin`, `cend`, `max_size` and aliases `size_type`, `difference_type` to collection 

ENDRELEASENOTES

Fixes #479, Fixes #546

The goal of this PR is to document current behavior not to fix things. It'll be easier to fix later once we known what is missing. A few methods and aliases were added but they don't collide with anything else

C++ specifies *Container* as a [named requirement](https://en.cppreference.com/w/cpp/named_req/Container), that means specifies required types, functions, semantics and behaviors as well post requirements on its iterators.

The requirements were collected in a form of a markdown file. Next to each requirement it's indicated whether it's fulfilled by the collection/iterators. A short comment indicates any peculiar behaviors or why it isn't fulfilled.
The document is accompanied by a set of tests. The tests are meant to detect changes in collection semantic and interface, and fail indicating need to update the documentation

Beside *Container* a short paragraph and a few tests are added for checking *AllocationAwareContainer* (collection uses default allocator implicitly #424), iterator adapters. Another short paragraph is added on interactions with standard algorithms (algorithms require at least *LegacyInputIterator* which is not fulfilled by collection iterators, the code may compile but it's std-implementation depended #150) and ranges (which use at least *std::input_iterator* constrain that again isn't fulfilled by collection iterators #273) 

- [x] comparison in text document
- [x] test Container interface
- [x] test Container semantics
- [x] test iterator interface
- [x] test iterator semantics
- [x] test iterator concepts 
- [x] simple test for AllocationAwareContainer
- [x] explain why algorithms and ranges are broken 
- [x] test both iterator and const_iterator
- [x] tests with subcollections


include/podio/UserDataCollection.h
```diff
@@ -206,6 +217,12 @@ class UserDataCollection : public CollectionBase {
   typename std::vector<BasicType>::const_iterator end() const {
     return _vec.end();
   }
+  typename std::vector<BasicType>::const_iterator cbegin() const {
```
hegner, line 220: shouldn't we use here the definitions of line 80-84 instead of repeating?

hegner (changes requested):

include/podio/UserDataCollection.h
```diff
@@ -77,6 +77,12 @@ class UserDataCollection : public CollectionBase {
   VectorMembersInfo m_vecmem_info{};
 
 public:
+  using value_type = typename decltype(_vec)::value_type;
```
tmadlener, line 80:
```suggestion
  using value_type = typename std::vector<BasicType>::value_type;
```

Or alternatively introduce a `using VecType = std::vector<BasicType>` (private) member type, to not have the `decltype` in there. I think that would make this a bit more readable.

include/podio/UserDataCollection.h
m-fila: Thanks, `begin` , `cbegin`, `end`, `cend` use now type aliases

m-fila:
> AFAICT this is ready for review, right? One general comment this has to be included in the `doc/index.rst` file to show up in the documentation in the end.

Yes, please :slightly_smiling_face: 

I realized that we have a detection idiom implemented in utils. I could rewrite the traits in test with it to reduce the boiler plate a bit if you think it's worth it

tmadlener:
No strong opinion on the usage of the detection idiom, I leave it up to you.

doc/collections_as_container.md
```diff
@@ -0,0 +1,173 @@
...
+| `std::counted_iterator` |  ❌ no | `iterator` and `const_iterator` not `std::input_or_output_iterator` |
+
+
+## Collection and standard algorithms
+
+Most of the standard algorithms require the iterators to be at least *InputIterator*. The iterators of PODIO collection don't fulfil this requirement, therefore they are not compatible with standard algorithms according to the specification. In practice some algorithms may still compile with the collections depending on an implementation of a given algorithm.
+
+## Standard range algorithms
+
+The standard range algorithm use constrains to operate at least on `std::input_iterator`s and `std::ranges::input_range`s. The iterators of PODIO collection don't model these concepts, therefore can't be used with standard range algorithms.
```
tmadlener, line 173: I would move this further up (almost all the way to the top?), because for most users this will be the interesting information. We would then also need to specify that algorithms that mutate the collection (filtering, sorting, etc.) will not work, but things that do not mutate the collection should in practice work.

doc/collections_as_container.md
```diff
@@ -0,0 +1,173 @@
...
+| `a = rv` | `C&` | Destroys or move-assigns all elements of `a` from elements of `rv` |  ✔️ yes |
+| `a.~C()` | `void` | Destroys all elements of `a` and frees all memory|  ✔️ yes |
+| `a.begin()` | `(const_)iterator` | Iterator to the first element of `a` |  ✔️ yes |
+| `a.end()` | `(const_)iterator` | Iterator to one past the last element of `a` | ✔️ yes |
+| `a.cbegin()` | `const_iterator` | Same as `const_cast<const C&>(a).begin()` | ✔️ yes |
+| `a.cend()` | `const_iterator` | Same as `const_cast<const C&>(a).end()`|  ✔️ yes |
+| `a == b` | Convertible to `bool` | Same as `std::equal(a.begin(), a.end(), b.begin(), b.end())`| ❌ no | Not defined |
+| `a != b` | Convertible to `bool` | Same as `!(a == b)` | ❌ no | Not defined |
+| `a.swap(b)` | `void` | Exchanges the values of `a` and `b` | ❌ no | Not defined |
+| `swap(a,b)` | `void` | Same as `a.swap(b)` | ✔️ yes | `a.swap(b)` not defined |
```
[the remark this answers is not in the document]
m-fila: Right

doc/collections_as_container.md
```diff
@@ -0,0 +1,173 @@
+# PODIO Collection as a *Container*
+
+Comparison of the PODIO `Collection`s with a C++ named requirement [*Container*](https://en.cppreference.com/w/cpp/named_req/Container).
+
+The PODIO `Collection`s interface was designed to remind the standard *Container* interface, in particular `std::vector`. The perfect compliance with the *Container* is not achieved as the `Collection`s are concerned with additional semantics such as mutable/immutable element access, associations and relations, and IO which are not part of *Container*.
+
```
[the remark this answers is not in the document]
m-fila: Added

doc/collections_as_container.md
```diff
@@ -0,0 +1,173 @@
+# PODIO Collection as a *Container*
+
+Comparison of the PODIO `Collection`s with a C++ named requirement [*Container*](https://en.cppreference.com/w/cpp/named_req/Container).
+
+The PODIO `Collection`s interface was designed to remind the standard *Container* interface, in particular `std::vector`. The perfect compliance with the *Container* is not achieved as the `Collection`s are concerned with additional semantics such as mutable/immutable element access, associations and relations, and IO which are not part of *Container*.
```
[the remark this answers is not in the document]
m-fila: Fixed as in suggestion.
Added a paragraph on collection storing data layer objects instead of user layer objects

doc/collections_as_container.md
m-fila: Added comment that if they even compile then the algorithms mutating collection should give wrong results but not-mutating should work.

I'm not sure about putting it at the top as here the iterator requirements and iterators are used but they weren't introduced or checked before. My original intent  here was to test and list as many algorithms as possible like:

| Algorithm | Works with `iterator`/`const_iterator`| Comment |
|----------------|------------------------------------------------------|----------------|
| `std::fill` | ❌ no/❌ no |Why it's broken|

but then I realized that from standard point of view we can't guarantee anything since we don't fulfill basic requirements. So for instance for `std::all_of` that doesn't really do anything special and should work , if `g++` decides to check  for example `iterator_category` (because it should be present) then we get compile error, but if clang decides it doesn't care about `iterator_category` then it's fine

 I added a short sentence in bold at the top that the algorithms might be broken and link here

m-fila:
> Thanks again for putting this together. I like this quite a lot, and I have mainly smaller comments (see below).
> 
> One thing that I am not sure how to handle best yet is the fact, that you stayed entirely within the STL language of using `iterator` and `const_iterator`. This makes this document self contained and it is straight forward to compare to the STL and the named requirements. However, there is no mention yet of the `Iterator` and `MutableIterator` that we generate. Do you think there is an easy way to at least mention them somewhere. I think that could reduce the potential for confusion if people look e.g. at the doxygen API documentation for EDM4hep.

Thank you very much :blush:
I added in the collection table that `iterator` is defined as `MutableCollectionIterator` and similar for `const_iterator`. And a short paragraph at the start of iterator part that this convention is used for `iterator` and `const_iterator`

As for the detection idiom I think I'll leave it as is

tmadlener (approved):
I am happy with this as it is. In case we really want to go through all > 100 STL algorithms and document them we can do that later as well, but for now this should serve as a very good basis and we have something that we can refer to for capabilities.

@hegner do you want to have another look?

tmadlener:
I heard no complaints, so I will merge this.

hegner:
Thanks!

PR #608, opened 2024-05-24
Add python bindings for Frame::getName

tmadlener:
BEGINRELEASENOTES
- Add python bindings for the `Frame::getName` functionality introduced in #586 

ENDRELEASENOTES

### DIFF
--- a/python/podio/frame.py
+++ b/python/podio/frame.py
@@ -129,7 +129,7 @@ class Frame:
                 return f"{tok:0>8x}"
             return _get_id(tok.getID())
 
-        raise KeyError(f"No collection name can be found in Frame for {_get_id(token)}")
+        raise KeyError(f"No collection name can be found in Frame for collection id: {_get_id(token)}")
 
     def put(self, collection, name):
         """Put the collection into the frame
--- a/python/podio/frame.py
+++ b/python/podio/frame.py
@@ -2,7 +2,6 @@
 """Module for the python bindings of the podio::Frame"""
 
 import cppyy
-from copy import deepcopy
 
 import ROOT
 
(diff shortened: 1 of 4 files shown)


tmadlener:
It looks like the additional overloads are not necessary. I will remove them again and only keep the python bindings for the existing ones.

python/podio/frame.py
```diff
@@ -107,6 +107,30 @@ def get(self, name):
...
+        maybeName = self._frame.getName(token)
+        if maybeName.has_value():
+            return maybeName.value()
+
+        def _get_id(tok):
+            if isinstance(tok, int):
+                return f"{tok:0>8x}"
+            return _get_id(tok.getID())
+
+        raise KeyError(f"No collection name can be found in Frame for {_get_id(token)}")
```
tmadlener, line 132:
```suggestion
        raise KeyError(f"No collection name can be found in Frame for collection id: {_get_id(token)}")
```

python/podio/frame.py
```diff
@@ -107,6 +107,30 @@ def get(self, name):
...
+        maybeName = self._frame.getName(token)
+        if maybeName.has_value():
+            return maybeName.value()
+
+        def _get_id(tok):
+            if isinstance(tok, int):
+                return f"{tok:0>8x}"
+            return _get_id(tok.getID())
+
+        raise KeyError(f"No collection name can be found in Frame for collection id: {_get_id(token)}")
```
tmadlener, line 132:
```suggestion
        raise KeyError(
              f"No collection name can be found in Frame for collection id: {_get_id(token)}"
        )
```

PR #610, opened 2024-05-27
Drop unused comparison templates

jmcarcell:
BEGINRELEASENOTES
- Drop unused comparison templates

ENDRELEASENOTES

It has been a while since they have been introduced and are not in use so they can be deleted.


tmadlener (commented):
Does the full stack still build with this? I probably introduced for the migration, but if things build, I am fine with removing these.

hegner:
Same with me. I wouldn't mind taking it out tonight and re-adding in case. But I doubt there will be problems

jmcarcell:
At least the core EDM4hep, k4FWCore and the wrapper are fine with this, so I'm also in favor of checking with the nightlies.

Issue #590, opened 2024-04-23
GenericParameters should be stored the same way for TTree and RNTuple backends


tmadlener:
Currently the `ROOTWriter` and `ROOTReader` (TTree) based store the `GenericParameters` of a Frame directly via a custom dictionary, e.g.:

https://github.com/AIDASoft/podio/blob/02a4b9d5d704eb572ce945594b4c85303e89e216/src/ROOTReader.cc#L42-L46

On the other hand the RNTuple backend splits these up into pairs of vectors for the keys and the values:
https://github.com/AIDASoft/podio/blob/02a4b9d5d704eb572ce945594b4c85303e89e216/src/RNTupleWriter.cc#L249-L257

These should both use the same mechanism to avoid confusion. Given that RNTuple doesn't support writing map types and that pairs of vectors are easier to interpret without knowing about GenericParameters, we decided to switch the TTree based backend to that as well. See [EDM4hep meeting on Apr 23, 2024](https://indico.cern.ch/event/1409361/)

jmcarcell:
At the time of writing the `RNTupleWriter.cc` there wasn't support for `std::map` but now there is support for it: https://github.com/root-project/root/pull/13904
So the cleaner choice is to move the `RNTupleWriter` to use `std::map`. However, someone complained today that uproot may not support `std::map` but I don't know if we want to have our files be compatible with uproot...

tmadlener:
I would prefer the solution that makes the files more easy to read without root / with uproot actually. I also think that would probably play a bit nicer with RDataFrame (without having ever tested this).

Zehvogel:
> I also think that would probably play a bit nicer with RDataFrame (without having ever tested this).

For RDataFrame this more or less does not matter as the easiest way to access GenericParameters is to include the header and use the real type.

```python
ROOT.gInterpreter.Declare("#include <podio/GenericParameters.h>")
df = df.Define("xsec_fb", "PARAMETERS.getValue<float>(\"crossSection\")")
```

tmadlener:
I am not sure we can really preserve the currently possible behavior for reading the parameters as a `GenericParameter` object as is possible now using TTrees. However, we might be able to simply offer similar functionality via the RDataSource in #593.

My currently favored options at the moment would be to either
- Move the TTree implementation over to use the same approach as RNTuple, i.e. pairs of vectors containing the keys and values
- or make both the TTree and RNtuple based implementations write maps

Zehvogel:
[...]

PR #614, opened 2024-06-05
Refactor the RNTupleWriter internals and reduce some code duplication

tmadlener:
BEGINRELEASENOTES
- Refactor the `RNTupleWriter` internals a bit and reduce some code duplication between the `RNTupleWriter` and the `ROOTWriter`
- Add `getKeys` and `getValues` methods to `GenericParameters`
- Remove the `CollectionBranches.h` header and move all root related helpers into `utilities/RootHelpers.h`

ENDRELEASENOTES

In more detail:
- Rename the internal `RNTupleWriter::CollectionInfo` struct to `RNTupleWriter::CategoryInfo` and group the members that are used to store and write the GenericParameters data into that, such that all data belonging to a category are grouped together
- Introduce `ParamStorage` to more easily refer to a pair of keys and values (vectors). This is done in preparation for another PR that is in the works for addressing #590.
- Move some of the typdefs that have been present in the `ROOTWriter` and the `RNTupleWriter` into the newly created `utilities/RootHelpers.h` header


tmadlener (commented):
Merging this to unblock #615 later today unless there are comments.

include/podio/GenericParameters.h
```diff
@@ -243,5 +248,17 @@ std::vector<std::string> GenericParameters::getKeys() const {
   return keys;
 }
 
+template <typename T, typename>
+std::vector<std::vector<T>> GenericParameters::getValues() const {
+  std::vector<std::vector<T>> values;
+  {
+    auto& mtx = getMutex<T>();
+    const auto& map = getMap<T>();
+    std::lock_guard lock{mtx};
```
hegner, line 257: Whenever locks are being used one should document where why and when they are needed

include/podio/utilities/RootHelpers.h
```diff
@@ -0,0 +1,44 @@
...
+  using StoreCollection = std::tuple<const std::string&, podio::CollectionBase*>;
+
+  /// Small helper struct to collect all branches that are necessary to read or
+  /// write a collection. Needed to cache the branch pointers and avoid having to
+  /// get them from a TTree/TChain for every event.
+  struct CollectionBranches {
+    TBranch* data{nullptr};
+    std::vector<TBranch*> refs{};
+    std::vector<TBranch*> vecs{};
+    std::vector<std::string> refNames{}; ///< The names of the relation branches
```
hegner, line 32: in principle one could do indices instead of strings here, but for the sake of simplicity we stay with this.

hegner (changes requested):
Thanks. Looks like boring, but important refactoring.
I would like to see some more documentation as mentioned in a comment

PR #615, opened 2024-06-05
[WIP] Make the TTree based ROOT backend store the GenericParameters the same way as the RNTuple based one

### LINKED ISSUE #590: GenericParameters should be stored the same way for TTree and RNTuple backends

tmadlener:
BEGINRELEASENOTES
- Make the TTree based backend also write the GenericParameters as pairs of vectors (keys and values) to make the format the same for the RNTuple and the TTree based backends.

ENDRELEASENOTES

- Fixes #590 
- [ ] Builds on #614 
- [x] Includes #623 
- [x] Needs #580 in order to make the edm4hep tests pass again
- [x] Add an additional "legacy test file" for v0.99


include/podio/GenericParameters.h
```diff
@@ -30,6 +31,10 @@ class RNTupleWriter;
 
 namespace podio {
 
+#if !defined(__CLING__)
```
hegner, line 34: one line of documentation why that is the case - if possible to explain

hegner (approved):

hegner (approved):

PR #616, opened 2024-06-06
Switch to Alma9 for Key4hep based workflow

tmadlener:
BEGINRELEASENOTES
- Switch to Alma9 for the Key4hep based workflow in CI, since CentOS7 is no longer built regularly

ENDRELEASENOTES


jmcarcell (approved):

jmcarcell:
At some point I'll delete the CentOS 7 nightlies and everything using them will start to fail :smile:

PR #620, opened 2024-06-07
Add a c++ implementation for `podio-dump`

tmadlener:
BEGINRELEASENOTES
- Replace the `podio-dump` python implementation by a `podio-dump` wrapper script around the new `podio-dump-tool` c++ implementation. **This is an (almost) drop-in replacement**, only legacy files (i.e. files written before the introduction of the `Frame`) cannot be read by this.
  - The existing python implementation is still available as `podio-dump-legacy`. 
- **podio now depends on the [*fmtlib*](https://github.com/fmtlib/fmt) formatting library** (version 9 or greater) as the `podio-dump-tool` uses that for formatting.
  - This might change again in the future, but compilers don't have enough support for `std::format` yet to easily replace this.

ENDRELEASENOTES

This has taken a bit to develop (see below for the original description). The way it is now is a combination of two tools that are wrapped in a thin bash script that acts as the entry point and is called `podio-dump`. This is effectively only necessary because the datamodel definitions are stored as JSON inside the file and it's just way easier to transform that back to YAML in python than it is in c++. Hence, the bash script only checks for the corresponding flag and if that is present pipes the output of the c++ `podio-dump-tool` to a python script (`json-to-yaml`). Otherwise it simply calls the `podio-dump-tool` passing on all the arguments.

The dependency on *fmtlib* is necessary because it supports formatting of ranges and easy opt-in for implementations via `operator<<`. Additionally, `std::print` (and `println`) are only scheduled for c++23. Once there is more / all necessary support in `<format>`, adjusting to that and dropping the fmtlib dependency should be possible.

### Original description

This is an attempt at making `podio-dump` quicker after several complaints (e.g. https://github.com/key4hep/EDM4hep/issues/312). After some "profiling" it turns out that the slowest part in the python implementation is the loop over all the collections which can be significantly sped up by going to c++. In my local timings the current (python based) `podio-dump` is almost ten times slower than this (c++ based) `podio-dump-cpp`) for dumping the `example_frame.root` file from the tests (times via `time`)

| | `podio-dump` | `podio-dump-cpp` |
|--|--|--| 
| real | 12.393s |1.513s |
| user |8.522s | 1.251s|
| sys | 3.823s| 0.296s |

The main disadvantages of  the c++ implementation are that we need quite a bit of boilerplate for things that are trivial in python, e.g.:
- We have to manually implement argument parsing and (parts of) the `tabulate` functionality
  - Since formatting with `iostream` and `iomanip` is bordering on masochism, I have decided to pull in [`fmt`](https://fmt.dev/latest/index.html) for now. In principle c++20 has similar functionality in `std::format` (but no `fmt::print` that only comes with c++26). However, that requires gcc >=13 and clang >= 16.
- Dumping datamodel definitions in YAML is missing entirely at the moment, since that would require dumping the internal json format as YAML. In python this is literally these 2 lines:
 https://github.com/AIDASoft/podio/blob/d275460be9c098f2a1d83e67ec63398538673d05/tools/podio-dump#L99-L100
 in c++ this would require at least one other library to be pulled in
 
Since dumping the datamodel would require quite a bit of work in c++, I would be in favor of keeping that in python in a separate tool, while the other functionality could be covered by the c++ implementation.

TODO:
- [x] Implement detailed dumping mode
- [x] Builds on top of #618 
- [x] Includes #621


tools/CMakeLists.txt
```diff
@@ -1,3 +1,7 @@
+add_executable(podio-dump-cpp src/podio-dump.cpp)
+target_link_libraries(podio-dump-cpp PRIVATE podio::podio podio::podioIO fmt::fmt)
+
```
tmadlener, line 3: Executable needs to be installed.

tools/src/tabulate.h
```diff
@@ -0,0 +1,71 @@
...
+std::string rowFormatString(const std::vector<size_t>& colWidths) {
+  std::string rowFmt = "";
+  rowFmt.reserve(colWidths.size() * 6); // heuristic working for columns that are 2 digit numbers wide
+
+  for (const auto w : colWidths) {
+    rowFmt.append(fmt::format("{{:<{}}}", w));
+  }
+
+  return rowFmt;
+}
```
tmadlener, line 21:
```suggestion
```
Unused

Zehvogel:
I wonder how an RDataFrame-based python version (with pre-compiled functions) would fare on a performance vs. comfort scale

tmadlener:
I have rebased this onto the latest version of podio. The main difference to before is that `podio-dump` is now a thin bash script that simply wraps the `podio-dump-tool` c++ executable. In case the user wants to dump an EDM via `--dump-edm` we invoke a thin python tool `json-to-yaml` that does the conversion on the fly. Now this is a drop-in replacement for the previous version of `podio-dump` implemented in python. The one thing the c++ implementation does not handle is reading pre-release legacy files. (I don't think that is an actual requirement any longer).

There are some formatting differences wrt to the python implmentation, which are pretty much all related to the alignment of numerical outputs in the tables.

Using ROOT 6.34.04 and python 3.11.11 I get the following times (simply using the test times here):

**Now**
```
1/7 Test #185: podio-dump-help ..................   Passed    0.03 sec
2/7 Test #186: podio-dump-root ..................   Passed    3.58 sec
3/7 Test #187: podio-dump-detailed-root .........   Passed    3.51 sec
4/7 Test #190: podio-dump-sio ...................   Passed    0.04 sec
5/7 Test #191: podio-dump-detailed-sio ..........   Passed    0.04 sec
6/7 Test #194: podio-dump-rntuple ...............   Passed    0.24 sec
7/7 Test #195: podio-dump-detailed-rntuple ......   Passed    0.23 sec

```
and
```
1/6 Test #160: datamodel_def_store_roundtrip_root ................   Passed    0.90 sec
2/6 Test #161: datamodel_def_store_roundtrip_root_extension ......   Passed    0.81 sec
3/6 Test #162: datamodel_def_store_roundtrip_sio .................   Passed    0.68 sec
4/6 Test #163: datamodel_def_store_roundtrip_sio_extension .......   Passed    0.60 sec
5/6 Test #164: datamodel_def_store_roundtrip_rntuple .............   Passed    0.81 sec
6/6 Test #165: datamodel_def_store_roundtrip_rntuple_extension ...   Passed    0.71 sec
```

**Before**
```
1/7 Test #185: podio-dump-help ..................   Passed    0.17 sec
2/7 Test #186: podio-dump-root ..................   Passed    8.50 sec
3/7 Test #187: podio-dump-detailed-root .........   Passed    9.35 sec
4/7 Test #190: podio-dump-sio ...................   Passed    8.08 sec
5/7 Test #191: podio-dump-detailed-sio ..........   Passed    7.71 sec
6/7 Test #194: podio-dump-rntuple ...............   Passed    8.36 sec
7/7 Test #195: podio-dump-detailed-rntuple ......   Passed    8.00 sec
```
and
```
1/6 Test #160: datamodel_def_store_roundtrip_root ................   Passed    2.62 sec
2/6 Test #161: datamodel_def_store_roundtrip_root_extension ......   Passed    2.52 sec
3/6 Test #162: datamodel_def_store_roundtrip_sio .................   Passed    2.32 sec
4/6 Test #163: datamodel_def_store_roundtrip_sio_extension .......   Passed    2.30 sec
5/6 Test #164: datamodel_def_store_roundtrip_rntuple .............   Passed    2.63 sec
6/6 Test #165: datamodel_def_store_roundtrip_rntuple_extension ...   Passed    2.53 sec
```

So, depending on which ROOT backend is used the speedup is somewhere between 2x and 30x. For SIO it is another factor 10 on top of that (for the test files). Even if the speedup is "only" a factor two for the most prevalent file format at the moment, I still think this would be worth it, also considering the increased implementation side, since many things that came packaged for python are now handrolled here.

----

I have checked `fmt`-lib down to version 8 which didn't compile, so minimal fmt version is 9. Things work unchanged up to version 11 (aka the latest at the moment).

tools/src/podio-dump-tool.cpp
```diff
@@ -0,0 +1,267 @@
...
+  if (argv.size() != 1) {
+    printUsageAndExit();
+  }
+  args.inputFile = argv[0];
+
+  return args;
+}
+
+template <typename T>
+std::string getTypeString() {
```
jmcarcell, line 139:
```suggestion
constexpr std::string getTypeString() {
```
I'm not sure this will change anything...

tools/src/podio-dump-tool.cpp
```diff
@@ -0,0 +1,267 @@
...
+  getParameterOverview<int>(frame, paramRows);
+  getParameterOverview<float>(frame, paramRows);
+  getParameterOverview<double>(frame, paramRows);
+  getParameterOverview<std::string>(frame, paramRows);
+
+  printTable(paramRows, {"Name", "Type", "Elements"});
+}
+
+template <typename... Args>
+void print_flush(fmt::format_string<Args...> fmtstr, Args&&... args) {
```
jmcarcell, line 185:
```suggestion
void print_flush(const fmt::format_string<Args...>& fmtstr, Args&&... args) {
```

tools/src/podio-dump-tool.cpp
```diff
@@ -0,0 +1,267 @@
...
+  fmt::print(fmtstr, std::forward<Args>(args)...);
+  std::fflush(stdout);
+}
+
+void printFrameDetailed(const podio::Frame& frame) {
+  fmt::print("Collections:\n");
+  const auto collNames = frame.getAvailableCollections();
+  for (const auto& name : podio::utils::sortAlphabeticaly(collNames)) {
+    const auto coll = frame.get(name);
+    print_flush("{}\n", name);
```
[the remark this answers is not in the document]
tmadlener: I have to check. This might also just be a remnant from porting the python implementation, where flushing on the python side was necessary, because otherwise python and c++ streams (from `coll.print()`) would not be synchronized properly.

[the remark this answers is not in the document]
tmadlener: Seems to work without.

tools/podio-dump.py
```diff
@@ -0,0 +1,213 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
tmadlener: Renamed to `podio-dump-legacy`

tools/src/podio-dump-tool.cpp
tmadlener: The whole function has been removed, as it is no longer necessary.

tools/src/podio-dump-tool.cpp
tmadlener: Made it a `consteval std::string_view`, so it's definitely evaluated at compile time.

tools/src/podio-dump-tool.cpp
```diff
@@ -0,0 +1,263 @@
...
+#include <string>
+#include <tuple>
+
+template <>
+struct fmt::formatter<podio::version::Version> : ostream_formatter {};
+
+struct ParsedArgs {
+  std::string inputFile{};
+  std::string category{"events"};
+  std::vector<size_t> events = std::vector<size_t>(1, 0);
```
jmcarcell, line 25:
```suggestion
  std::vector<size_t> events{0};
```

tools/src/argparseUtils.h
```diff
@@ -0,0 +1,54 @@
...
+    if (pos != str.size()) {
+      throw std::invalid_argument("");
+    }
+    return number;
+  } catch (const std::invalid_argument& err) {
+    parseError();
+  }
+
+  return -1;
+}
```
jmcarcell, line 52:
```suggestion
inline size_t parseSizeOrExit(const std::string& str) {
  try {
    std::size_t pos{};
    const auto number = std::stoull(str, &pos);
    if (pos != str.size()) {
      throw std::invalid_argument("");
    }
    return number;
  } catch (const std::invalid_argument&) {
    std::cerr << "'" << str << "' cannot be parsed into an integer number" << std::endl;
    std::exit(1);
  }
}
```
Address a few things:
- There is no need to create a lambda that is used once
- `return -1` can never be reached
- Avoid a change in signedness when returning `number` by using `std::stoull`

tools/src/podio-dump-tool.cpp
```diff
@@ -0,0 +1,263 @@
...
+  -e ENTRIES, --entries ENTRIES
+                        Which entries to print. A single number, comma separated list of numbers or "first:last" for an inclusive range of entries. Defaults to the first entry.
+  -d, --detailed        Dump the full contents not just the collection info
+  --dump-edm DUMP_EDM   Dump the specified EDM definition from the file in yaml format
+  --version             show program's version number and exit
+)";
+
+void printUsageAndExit() {
+  fmt::println(stderr, "{}", usageMsg);
+  std::exit(1);
```
andresailer, line 51: Can we exit with 0 when we ask for help, and with 1 if we gave bad arguments?

andresailer: Never mind, I see that is how it is implemented below!

tools/src/argparseUtils.h
```diff
@@ -0,0 +1,54 @@
...
+      }
+    } else {
+      token += ch;
+    }
+  }
+  if (!token.empty()) {
+    tokens.push_back(token);
+  }
+  return tokens;
+}
```
jmcarcell, line 32:
```suggestion
inline std::vector<std::string> splitString(const std::string& str, const char delimiter) {
    std::vector<std::string> result;

    for (const auto part : std::ranges::views::split(str, delimiter)) {
      result.emplace_back(part.begin(), part.end());
    }

    return result;
}
```
There is `std::ranges::views::split` to do exactly what this function does. In principle it should be possible to replace all the occurrences of `splitString` with `std::ranges::views::split` but more code would have to be modified to use views so this is a partial solution.

tools/src/podio-dump-tool.cpp
tmadlener: Done

tools/src/argparseUtils.h
tmadlener: I have made this return a `vector<string_view>` and adapted the `parseSizeOrExit` to take a `string_view`.

tools/src/podio-dump-tool.cpp
```diff
@@ -0,0 +1,263 @@
...
+    events.reserve(splitRange.size());
+    std::ranges::transform(splitRange, std::back_inserter(events),
+                           [](const auto& elem) { return parseSizeOrExit(elem); });
+
+    return events;
+  }
+
+  parseError();
+  return {};
+}
```
jmcarcell, line 95:
```suggestion
std::vector<size_t> parseEventRange(const std::string_view evtRange) {
  auto parseError = [&evtRange]() {
    fmt::println(stderr, "'{}' cannot be parsed into a list of entries", evtRange);
    std::exit(1);
  };

  // Split by ',' and transform into a range of string views
  auto splitRange = evtRange | std::views::split(',') |
      std::views::transform([](auto&& subrange) { return std::string_view(subrange.begin(), subrange.end()); });

  if (std::ranges::distance(splitRange) == 1) {
    // Only one entry, check if it's a range (start:end)
    auto colonSplitRange = evtRange | std::views::split(':') |
        std::views::transform([](auto&& subrange) { return std::string_view(subrange.begin(), subrange.end()); });

    const auto it = std::ranges::begin(colonSplitRange);

    if (std::ranges::distance(colonSplitRange) == 1) {
      return {parseSizeOrExit(*it)};
    } else if (std::ranges::distance(colonSplitRange) == 2) {
      size_t start = parseSizeOrExit(*it);
      size_t stop = parseSizeOrExit(*std::ranges::next(it));
      std::vector<size_t> events(stop - start + 1);
      std::iota(events.begin(), events.end(), start);
      return events;
    } else {
      parseError();
    }
  } else {
    std::vector<size_t> events;
    events.reserve(std::ranges::distance(splitRange));
    std::ranges::transform(splitRange, std::back_inserter(events), parseSizeOrExit);
    return events;
  }

  parseError();
  return {};
}
```
I propose this range/views-based implementation that does not use the `splitString` function.

tools/src/argparseUtils.h
```diff
@@ -0,0 +1,42 @@
...
+inline std::vector<std::string_view> splitString(const std::string& str, const char delimiter) {
+  std::vector<std::string_view> tokens;
+
+  for (const auto& token : std::ranges::views::split(str, delimiter)) {
+    tokens.emplace_back(token.begin(), token.end());
+  }
+
+  return tokens;
+}
+
```
jmcarcell, line 27:
```suggestion
```
Not needed if the proposed implementation for `parseEventRange` is used.

tools/src/argparseUtils.h
```diff
@@ -0,0 +1,42 @@
...
+inline size_t parseSizeOrExit(std::string_view str) {
+  try {
+    std::size_t number{};
+    const auto [ptr, err] = std::from_chars(str.data(), str.data() + str.size(), number);
+    if (err != std::errc{} || ptr != str.data() + str.size()) {
+      throw std::invalid_argument("");
+    }
+    return number;
+  } catch (const std::invalid_argument&) {
+    std::cerr << "'" << str << "' cannot be parsed into an integer number" << std::endl;
```
jmcarcell, line 37: I've noticed that now passing a bad range like `-e 1:` fails with `'' cannot be parsed into an integer number` while before this would fail with: 
```
usage: podio-dump [-h] [-c CATEGORY] [-e ENTRIES] [-d] [--dump-edm DUMP_EDM] [--version]
                  inputfile
podio-dump: error: argument -e/--entries: '0:' cannot be parsed into a list of entries
```
Not a huge problem since it is stated in the help how to correctly pass a range, but the error message is a bit useless in that case.

tools/src/argparseUtils.h
```diff
@@ -0,0 +1,42 @@
+#ifndef PODIO_TOOLS_ARGPARSEUTILS_H // NOLINT(llvm-header-guard): folder structure not suitable
+#define PODIO_TOOLS_ARGPARSEUTILS_H // NOLINT(llvm-header-guard): folder structure not suitable
+
+#include <algorithm>
+#include <charconv>
+#include <iostream>
+#include <ranges>
```
jmcarcell, line 7:
```suggestion
```
If `splitString` is removed

tools/src/argparseUtils.h
```diff
@@ -0,0 +1,42 @@
...
+    const auto [ptr, err] = std::from_chars(str.data(), str.data() + str.size(), number);
+    if (err != std::errc{} || ptr != str.data() + str.size()) {
+      throw std::invalid_argument("");
+    }
+    return number;
+  } catch (const std::invalid_argument&) {
+    std::cerr << "'" << str << "' cannot be parsed into an integer number" << std::endl;
+    std::exit(1);
+  }
+}
```
jmcarcell, line 40:
```suggestion
inline size_t parseSizeOrExit(std::string_view str) {
  std::cout << "Parsing '" << str << "' into a size_t" << std::endl;
  std::size_t number{};
  const auto [ptr, err] = std::from_chars(str.data(), str.end(), number);
  if (err != std::errc{} || ptr != str.end()) {
    std::cerr << "'" << str << "' cannot be parsed into an integer number" << std::endl;
    std::exit(1);
  }
  return number;
}
```
The function does not need to throw anymore. In the examples they use `str.data() + str.size()` but `str.end()` also works fine. I'm not sure if there is any difference.

tools/src/argparseUtils.h
tmadlener: Should be back to (almost) what it was before now. I have put in an explicit check for an empty string after the colon. Ideally `parseSizeOrExit` would return an optional, but that would require quite some other work as well.

tools/src/tabulate.h
```diff
@@ -0,0 +1,62 @@
+#ifndef PODIO_TOOLS_TABULATE_H // NOLINT(llvm-header-guard): folder structure not suitable
+#define PODIO_TOOLS_TABULATE_H // NOLINT(llvm-header-guard): folder structure not suitable
+
+#include <fmt/core.h>
+
+#include <algorithm>
+#include <iostream>
```
jmcarcell, line 7:
```suggestion
```
Marked by clangd

jmcarcell:
I don't have any more comments and probably won't have another look in detail unless I find something. I have been using this and it feels soooo much better to use, certainly way faster even in debug builds it is very fast. It also crashes faster when reading an incompatible file :smile: 
I wonder if having a better python implementation of the readers (I'm quite sure that's where most of the time was spent) would have helped. In any case this is good to have for a tool that is (should be) used a lot.

jmcarcell:
This is now in the Key4hep nightlies on Alma 9 and Ubuntu 24. On Ubuntu 22, it doesn't build with GCC 11 anymore with multiple failures:
```
/podio/tools/src/podio-dump-tool.cpp:85:35: error: could not convert '{<expression error>}' from '<brace-enclosed initializer list>' to 'std::vector<long unsigned int>'
   85 |       return {parseSizeOrExit(*it)};
```

```
/podio/tools/src/podio-dump-tool.cpp:78:54: error: no match for 'operator|' (operand types are 'std::ranges::split_view<std::basic_string_view<char>, std::ranges::single_view<char> >' and 'std::ranges::views::__adaptor::_Partial<std::ranges::views::_Transform, parseEventRange(std::string_view)::<lambda(auto:32&&)> >')
   78 |     auto colonSplitRange = evtRange | rv::split(':') |
      |                            ~~~~~~~~~~~~~~~~~~~~~~~~~ ^
      |                                     |
      |                                     std::ranges::split_view<std::basic_string_view<char>, std::ranges::single_view<char> >
   79 |         rv::transform([](auto&& subrange) { return std::string_view(subrange.begin(), subrange.end()); });
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                      |
      |                      std::ranges::views::__adaptor::_Partial<std::ranges::views::_Transform, parseEventRange(std::string_view)::<lambda(auto:32&&)> >
```

```
/podio/tools/src/podio-dump-tool.cpp:79:57: error: no matching function for call to 'std::basic_string_view<char>::basic_string_view(std::ranges::split_view<std::basic_string_view<char>, std::ranges::single_view<char> >::_InnerIter<false>, std::default_sentinel_t)'
   79 |         rv::transform([](auto&& subrange) { return std::string_view(subrange.begin(), subrange.end()); });
```

I wasn't aware of this, but actually building with GCC 11 (with tests) hasn't been possible since #626. I think probably the way to go forward with this is to request GCC > 11 rather than trying to make the errors go away.

tmadlener:
I am not entirely sure how to best deal with this, but gcc11 is still the system compiler on Alma9, so if we can keep support for it alive, I would be all for it.

jmcarcell:
On the other side, GCC 15 is going to be released in ~ 1 month, so by now GCC 11 is already 4 years old. In addition no one seems to be building podio with tests since https://github.com/AIDASoft/podio/pull/626 is from last year's summer.

tmadlener:
Well, it got merged in December and the v01-02 tag does not yet include it, so if people only build tagged versions they will not have encountered it yet. The main question is whether we consider gcc11 as c++20 compatible (enough) or not. Since we switched to c++20 after v01-02 that would also give us a clean cut and a reason to abandon gcc11.

Issue #563, opened 2024-02-22
Schema evolution script does not flag some evolutions as impossible


tmadlener:
The `podio_schema_evolution.py` script should catch problematic schema evolutions before they are actually commited. The recent swtich from `Vector3f` to `Vector3d` in EDM4hep was one of these changes as it is not supported by default by ROOT. However, the output of the current version of the script simply states the following:

```console
$ python3 ~/work/AIDASoft/podio/python/podio_schema_evolution.py edm4hep.yaml edm4hep_float_mom.yaml 
Comparing datamodel versions 1 and 1
Detected 2 schema changes:
 - 'edm4hep::MCParticle.momentum' changed type from '+                    'edm4hep::Vector3f to edm4hep::Vector3d
 - 'edm4hep::MCParticle.momentumAtEndpoint' changed type from '+                    'edm4hep::Vector3f to edm4hep::Vector3d
{}
```

This evolution should be flagged as impossible.

Related PRs / issues:
- https://github.com/key4hep/EDM4hep/pull/267 (plus PRs therein)
- https://github.com/key4hep/k4MarlinWrapper/issues/170 (what was broken by not realizing the issue early enough)

PR #622, opened 2024-06-10
Marking move from one component type to another one as impossible for schema evolution

hegner:
BEGINRELEASENOTES
- Detect more impossible schema evolutions
- Improving printout of schema evolution checker
ENDRELEASENOTES

This addresses #563


python/podio_schema_evolution.py
```diff
@@ -105,8 +105,8 @@ def __init__(self, name, member_name, old_member, new_member):
         self.new_member = new_member
         self.klassname = name
         super().__init__(
-            f"'{self.name}.{self.member_name}' changed type from '+\
-                    '{self.old_member.full_type} to {self.new_member.full_type}"
+            f"'{self.name}.{self.member_name}' changed type from "+\
+                    f"{self.old_member.full_type} to {self.new_member.full_type}"
```
tmadlener, line 109:
```suggestion
            f"'{self.name}.{self.member_name}' changed type from "
            f"{self.old_member.full_type} to {self.new_member.full_type}"
```

python/podio_schema_evolution.py
```diff
@@ -469,4 +475,4 @@ def read_evolution_file(self) -> None:
     comparator.read()
     comparator.compare()
     comparator.print_comparison()
-    print(comparator.get_changed_schemata(schema_filter=root_filter))
+    #print(comparator.get_changed_schemata(schema_filter=root_filter))
```
tmadlener, line 478: I suppose this was some debug output at some point, which is why it is removed?

tmadlener (commented):
This turned out to be a slightly simpler change than what I would have imagined :).

python/podio_schema_evolution.py
hegner: sort of. still useful for the future to include again

hegner:
> This turned out to be a slightly simpler change than what I would have imagined :).

Probably speaks for the design for once :-D

tmadlener (commented):
There seems to be an unrelated error that currently makes the script fail when trying to run it on the edm4hep.yaml file that triggered the original issue:

```console
  File "/home/tmadlener/work/AIDASoft/podio/python/podio_schema_evolution.py", line 476, in <module>
    comparator.compare()
  File "/home/tmadlener/work/AIDASoft/podio/python/podio_schema_evolution.py", line 184, in compare
    self._compare_components()
  File "/home/tmadlener/work/AIDASoft/podio/python/podio_schema_evolution.py", line 196, in _compare_components
    [
  File "/home/tmadlener/work/AIDASoft/podio/python/podio_schema_evolution.py", line 197, in <listcomp>
    AddedComponent(self.datamodel_new.components[name], name)
  File "/home/tmadlener/work/AIDASoft/podio/python/podio_schema_evolution.py", line 34, in __init__
    super().__init__(f"'{self.component.name}' has been added")
AttributeError: 'dict' object has no attribute 'name'
```

The issue is here:
https://github.com/AIDASoft/podio/blob/ab6b2a4374709d98407102e841edb9690b04d039/python/podio_schema_evolution.py#L34

it should be `self.name` instead of `self.component.name` to make things work again. With that fix the script detects invalid schema changes in the EDM4hep yaml file.

Can we add a simple test that at least runs the script on the example datamodel(s) that we have?

hegner:
> Can we add a simple test that at least runs the script on the example datamodel(s) that we have?
Surprised we didn't have. This PR or a new one?

tmadlener:
Yeah, I also thought we had one, but at least I couldn't find one. Or if we have one at least it didn't trigger this issue. I would prefer a new PR.

hegner:
We only had for the C++ part it seems. OK. new ticket, new PR

tmadlener:
PR to fix the bug in the schema evolution generation: #627

hegner:
other PR merged now

tmadlener:
Can you rebase this quickly to pick up those changes?

hegner:
obviously that fails now ;-) need to change the test definition

tmadlener (commented):
Works for me. Ideally we could condense down the test for a failing schema evolution a bit rather than having the third copy of the datamodel effectively, but we can also postpone that to after v1.0

hegner:
for post v1 we could do that a bit separated without a file, but in pure python tests. I will merge now.

PR #623, opened 2024-06-11
Move ROOTFrameData implementation into .cc file

tmadlener:
BEGINRELEASENOTES
- Move the implementation of `ROOTFrameData` member functions into a separate cc file.

ENDRELEASENOTES

PR #625, opened 2024-06-11
Make the TTree and RNTuple based backends write the GenericParameters the same way

### LINKED ISSUE #590: GenericParameters should be stored the same way for TTree and RNTuple backends

tmadlener:
BEGINRELEASENOTES
- Make the TTree based backend also write the GenericParameters as pairs of vectors (keys and values) to make the format the same for the RNTuple and the TTree based backends.

ENDRELEASENOTES

- Fixes #590 
- [x] Builds on #614 
- [x] Includes #623 
- [x] Needs #580 in order to make the edm4hep tests pass again
- [x] Add an additional "legacy test file" for v0.99

This was originally done in #615. See there for a bit more discussion.


include/podio/utilities/RootHelpers.h
```diff
@@ -36,7 +46,34 @@ namespace root_utils {
...
+
+    auto keysPtr() {
+      m_keysPtr = &keys;
+      return &m_keysPtr;
+    }
+
+    auto valuesPtr() {
+      m_valuesPtr = &values;
+      return &m_valuesPtr;
+    }
```
tmadlener, line 68: Document why these are necessary

src/ROOTWriter.cc
```diff
@@ -175,4 +195,11 @@ ROOTWriter::checkConsistency(const std::vector<std::string>& collsToWrite, const
   return {std::vector<std::string>{}, collsToWrite};
 }
 
+void ROOTWriter::fillParams(CategoryInfo& catInfo, const GenericParameters& params) {
+  catInfo.intParams = {params.getKeys<int>(), params.getValues<int>()};
+  catInfo.floatParams = {params.getKeys<float>(), params.getValues<float>()};
+  catInfo.doubleParams = {params.getKeys<double>(), params.getValues<double>()};
+  catInfo.stringParams = {params.getKeys<std::string>(), params.getValues<std::string>()};
```
tmadlener, line 202: Technically we should only take one lock here per type, rather than doing it separately for the keys and values.

tests/root_io/param_reading_rdataframe.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+ROOT.gInterpreter.Declare("#include <podio/utilities/RootHelpers.h>")
+df = ROOT.RDataFrame("events", sys.argv[1])
+
+evtWeights = (
+    df.Define(
+        "params",
+        "podio::root_utils::loadParamsFrom(GPIntKeys, GPIntValues, GPFloatKeys, GPFloatValues, GPDoubleKeys, GPDoubleValues, GPStringKeys, GPStringValues)",
+    )
```
tmadlener, line 14: @Zehvogel, does this look like an acceptable compromise for using GenericParameters in RDataFrame until we have a proper RDataSource?

Zehvogel: yes thx!

src/ROOTWriter.cc
```diff
@@ -117,28 +120,45 @@ void ROOTWriter::initBranches(CategoryInfo& catInfo, const std::vector<root_util
...
+  categoryInfo.branches[iColl + 1].data->SetAddress(categoryInfo.intParams.valuesPtr());
+
+  categoryInfo.branches[iColl + 2].data->SetAddress(categoryInfo.floatParams.keysPtr());
+  categoryInfo.branches[iColl + 3].data->SetAddress(categoryInfo.floatParams.valuesPtr());
+
+  categoryInfo.branches[iColl + 4].data->SetAddress(categoryInfo.doubleParams.keysPtr());
+  categoryInfo.branches[iColl + 5].data->SetAddress(categoryInfo.doubleParams.valuesPtr());
+
+  categoryInfo.branches[iColl + 6].data->SetAddress(categoryInfo.stringParams.keysPtr());
+  categoryInfo.branches[iColl + 7].data->SetAddress(categoryInfo.stringParams.valuesPtr());
```
tmadlener, line 161: Make sure to use consistent offsets here wrt reading instead of hardcoding them here.

hegner (changes requested):
Thanks. That seems to do things in a very thorough way now.

What I really would like to see for the future is some documentation about the chosen file structure in the files created.  Some `file_structure_details.doc` or similar

tmadlener:
I agree, more documentation on this is necessary. Since, we also don't really have documentation on how we store collections, I would prefer a separate PR for the documentation. I can prepare that in parallel.

tmadlener:
Documentation can be seen in #629, rendered by markdown [here](https://github.com/AIDASoft/podio/blob/44bd2afd4ab9e6ddffc1b71500e507b9f8284eba/doc/storage_details.md)

hegner (approved):

jmcarcell:
Something I found with this PR is that it fixes an issue I had only when making Debug builds with Clang 17. For some reason, the GenericParameters were not being saved in the file and all the tests for reading would fail. I picked up a working build of podio and read the files produced with this debug build and indeed the GenericParameters were not there. I never spent more time to figure out what was going on but out of all the build types and compilers (GCC and Clang) I only managed to reproduce it with Clang and `-DCMAKE_BUILD_TYPE=Debug`.

PR #626, opened 2024-06-11
Make collection iterators fulfill LegacyInputIterator and input_iterator concept

### LINKED ISSUE #272: Deleted iterator copy constructor/assignment and `find`ing

m-fila:
BEGINRELEASENOTES
- Added missing operations and type aliases so the collection iterators fulfill *LegacyInputIterator* requirement and `std::input_iterator` concept, and the collections fulfill  `std::input_range` concept. Thanks to these the collections are compatible with standard algorithms and standard range algorithms such as `std::find`, `std::count`, `std::all_of`

ENDRELEASENOTES

Fixes: #150, fixes #272
Conflicts: #273

Implement missing features so `CollectionIterators` and `MutableCollectionIterators` fulfill the requirements imposed on the C++ iterators (#598). It's expected than not all requirements can be fulfilled as some are in direct conflict with podio design:
- *LegacyForwardIterator*: dereference should return a reference, but in podio iterators are  'proxy iterators' returning user layer objects as a proxy for reference
- *LegacyForwardItertor*:  dereferencing  two iterators that are equal should give two references bound to the same object, but in podio the returned 'proxies' are different objects
-  `std::forward_iterator`: pointers and references obtained from an iterator should be valid as range the range is valid, but in podio the iterators behave like 'stashing iterators' -> pointers point to iterator member variable limiting their validity to the lifetime of the iterator.
- *LegacyOutputIterator*/`output_iterator`: datatype assignment replaces the internal object handle instead of modifying the handle properties.

As a consequence the iterators can the podio iterators can be at most *LegacyInputIterator* and `std::input_iterator`, and collections be `std::ranges::input_range`. This PR implements missing features to achieve this.

#### Named requirements:

- [x] *LegacyIterator*:
  - [x] copy constructor and assignment
  - [x] type aliases
- [x] *LegacyInputIterator*:
  - [x] postfix increment
- :x: *LegacForwardIterator*:
  - [x] value-initialization 
  - :x: dereference return reference not proxy
  - :x: dereferencing equal iterators returns references to the same object
- :x:  *OutputIterator*:
  - :x: assignment should modify collection
- [x] `iterator_category` is `std::input_iterator_tag` 
- [x] check selected algorithms operating on *LegacyInputIterators*

#### Concepts:

- [x] semantics checks for concepts
- [x] `std::input_or_output_iterator`
  - [x] same as *LegacyIterator* but  copyability not required
- [x] `std::input_iterator`
  - [x] `std::indirectly_readable` and `std::readable` (can dereference both `iterator` and `const iterator`)
- :x: `std::forward_iterator`
  - [x] syntax: constrains (except `iterator_concept` to be derived from`std::forward_iterator_tag`) 
  - :x: semantic: pointers obtained from iterator should remain valid as long as range is valid (no-stashing)
- [x] `iterator_concept` is `std::input_iterator_tag`
- [x] check range concepts
- [x] check selected range algorithms constrained to `std::input_iterator`

### DIFF
--- a/doc/collections_as_container.md
+++ b/doc/collections_as_container.md
@@ -7,7 +7,7 @@ The PODIO `Collection`s interface was designed to mimic the standard *Container*
 On the implementation level most of the differences with respect to the *Container* comes from the fact that in order to satisfy the additional semantics a `Collection` doesn't directly store [user layer objects](design.md#the-user-layer). Instead, [data layer objects](design.md#the-internal-data-layer) are stored and user layer objects are constructed and returned when needed. Similarly, the `Collection` iterators operate on the user layer objects but don't expose `Collection`'s storage directly to the users. Instead, they construct and return user layer objects when needed.
 In other words, a `Collection` utilizes the user layer type as a reference type instead of using plain references (`&` or `&&`) to stored data layer types.
 
-As a consequence some of the **standard algorithms may not work** with PODIO `Collection` iterators.  See [standard algorithm documentation](#collection-and-standard-algorithms) below.
+As a consequence some of the **standard algorithms may not work** with PODIO `Collection` iterators. See [standard algorithm documentation](#collection-and-standard-algorithms) below.
 
 The following tables list the compliance of a PODIO generated collection with the *Container* named requirement, stating which member types, interfaces, or concepts are fulfilled and which are not. Additionally, there are some comments explaining missing parts or pointing out differences in behaviour.
 
@@ -16,12 +16,12 @@ The following tables list the compliance of a PODIO generated collection with th
 | Name | Type | Requirements | Fulfilled by Collection? | Comment |
 |------|------|--------------|--------------------------|---------|
 | `value_type` | `T` | *[Erasable](https://en.cppreference.com/w/cpp/named_req/Erasable)* | ✔️ yes | Defined as an immutable user layer object type |
-| `reference` | `T&` |  | ❌ no | Not defined |
+| `reference` | `T&` | | ❌ no | Not defined |
 | `const_reference` | `const T&` | | ❌ no | Not defined |
 | `iterator` | Iterator whose `value_type` is `T` | [*LegacyForwardIterator*](https://en.cppreference.com/w/cpp/named_req/ForwardIterator) convertible to `const_iterator` | ❌ no | Defined as podio `MutableCollectionIterator`. `iterator::value_type` not defined, not [*LegacyForwardIterator*](https://en.cppreference.com/w/cpp/named_req/ForwardIterator) ([see below](#legacyforwarditerator)), not convertible to `const_iterator`|
 | `const_iterator` | Constant iterator whose `value_type` is `T` | [*LegacyForwardIterator*](https://en.cppreference.com/w/cpp/named_req/ForwardIterator) | ❌ no | Defined as podio `CollectionIterator`. `const_iterator::value_type` not defined, not [*LegacyForwardIterator*](https://en.cppreference.com/w/cpp/named_req/ForwardIterator) ([see below](#legacyforwarditerator))
 | `difference_type`| Signed integer | Must be the same as `std::iterator_traits::difference_type` for `iterator` and `const_iterator` | ❌ no | `std::iterator_traits::difference_type` not defined |
-| `size_type` | Unsigned integer | Large enough to represent all positive values of `difference_type` | ✔️ yes |  |
+| `size_type` | Unsigned integer | Large enough to represent all positive values of `difference_type` | ✔️ yes | |
 
 ### Container member functions and operators
 
@@ -31,18 +31,18 @@ The following tables list the compliance of a PODIO generated collection with th
 | `C(a)` | `C` | Creates a copy of `a` | ❌ no | Not defined, non-copyable by design |
 | `C(rv)` | `C` | Moves `rv` | ✔️ yes | |
 | `a = b` | `C&` | Destroys or copy-assigns all elements of `a` from elements of `b` | ❌ no | Not defined, non-copyable by design |
-| `a = rv` | `C&` | Destroys or move-assigns all elements of `a` from elements of `rv` |  ✔️ yes | |
-| `a.~C()` | `void` | Destroys all elements of `a` and frees all memory|  ✔️ yes | Invalidates all handles retrieved from this collection |
-| `a.begin()` | `(const_)iterator` | Iterator to the first element of `a` |  ✔️ yes | |
+| `a = rv` | `C&` | Destroys or move-assigns all elements of `a` from elements of `rv` | ✔️ yes | |
+| `a.~C()` | `void` | Destroys all elements of `a` and frees all memory| ✔️ yes | Invalidates all handles retrieved from this collection |
+| `a.begin()` | `(const_)iterator` | Iterator to the first element of `a` | ✔️ yes | |
 | `a.end()` | `(const_)iterator` | Iterator to one past the last element of `a` | ✔️ yes | |
 | `a.cbegin()` | `const_iterator` | Same as `const_cast<const C&>(a).begin()` | ✔️ yes | |
-| `a.cend()` | `const_iterator` | Same as `const_cast<const C&>(a).end()`|  ✔️ yes | |
+| `a.cend()` | `const_iterator` | Same as `const_cast<const C&>(a).end()`| ✔️ yes | |
 | `a == b` | Convertible to `bool` | Same as `std::equal(a.begin(), a.end(), b.begin(), b.end())`| ❌ no | Not defined |
 | `a != b` | Convertible to `bool` | Same as `!(a == b)` | ❌ no | Not defined |
 | `a.swap(b)` | `void` | Exchanges the values of `a` and `b` | ❌ no | Not defined |
 | `swap(a,b)` | `void` | Same as `a.swap(b)` | ❌ no | `a.swap(b)` not defined |
 | `a.size()` | `size_type` | Same as `std::distance(a.begin(), a.end())` | ✔️ yes | |
-| `a.max_size()` | `size_type` | `b.size()` where b is the largest possible container | ✔️ yes | |
+| `a.max_size()` | `size_type` | `b.size()` where `b` is the largest possible container | ✔️ yes | |
 | `a.empty()` | Convertible to `bool` | Same as `a.begin() == a.end()` | ✔️ yes | |
 
 ## Collection as an *AllocatorAwareContainer*
@@ -53,9 +53,9 @@ PODIO collections don't provide a customization point for allocators and use onl
 
 ### AllocatorAwareContainer types
 
-| Name |  Requirements | Fulfilled by Collection? | Comment |
+| Name | Requirements | Fulfilled by Collection? | Comment |
 |------|--------------|--------------------------|---------|
-| `allocator_type`  | `allocator_type::value_type` same as `value_type` | ❌ no | `allocator_type` not defined |
+| `allocator_type` | `allocator_type::value_type` same as `value_type` | ❌ no | `allocator_type` not defined |
 
 ### *AllocatorAwareContainer* expression and statements
 
@@ -120,7 +120,7 @@ In the following tables a convention from `Collection` is used: `iterator` stand
 
 | Expression | Return type | Semantics | Fulfilled by `iterator`/`const_iterator`? | Comment |
 |------------|-------------|-----------|-------------------------------------------|---------|
-| `i != j` |  Contextually convertible to `bool` | Same as `!(i==j)` | ✔️ yes / ✔️ yes | |
+| `i != j` | Contextually convertible to `bool` | Same as `!(i==j)` | ✔️ yes / ✔️ yes | |
 | `*i` | `reference`, convertible to `value_type` | | ❌ no / ❌ no | `reference` and `value_type` not defined |
 | `i->m` | | Same as `(*i).m` | ✔️ yes / ✔️ yes | |
 | `++r` | `It&` | | ✔️ yes / ✔️ yes | |
@@ -155,7 +155,7 @@ In addition to the *LegacyForwardIterator* the C++ standard specifies also the *
 | Expression | Return type | Semantics | Fulfilled by `iterator`/`const_iterator`? | Comment |
 |------------|-------------|-----------|-------------------------------------------|---------|
 | `*r = o` | | | ❗ attention / ❗ attention | Defined but an assignment doesn't modify objects inside collection |
-| `++r` | `It&` | | ✔️ yes / ✔️ yes  | |
+| `++r` | `It&` | | ✔️ yes / ✔️ yes | |
 | `r++` | Convertible to `const It&` | Same as `It temp = r; ++r; return temp;` | ❌ no / ❌ no | Post-increment not defined |
 | `*r++ = o` | | Same as `*r = o; ++r;`| ❌ no / ❌ no | Post-increment not defined |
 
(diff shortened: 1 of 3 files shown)


python/templates/macros/iterator.jinja2
```diff
@@ -3,10 +3,20 @@
...
+
   {{ iterator_type }}(size_t index, const {{ class.bare_type }}ObjPointerContainer* collection) : m_index(index), m_object({{ ptr_init }}), m_collection(collection) {}
+  {{ iterator_type }}(): {{ iterator_type }}(0, nullptr) {}
 
-  {{ iterator_type }}(const {{ iterator_type }}&) = delete;
-  {{ iterator_type }}& operator=(const {{ iterator_type }}&) = delete;
+  {{ iterator_type }}(const {{ iterator_type }}&) = default;
+  {{ iterator_type }}({{ iterator_type }}&&) = default;
+  {{ iterator_type }}& operator=(const {{ iterator_type }}&) = default;
+  {{ iterator_type }}& operator=({{iterator_type}}&&) = default;
```
tmadlener, line 18: I suppose all of these are working "as expected" if they are `defaulted`? In the end all the members are either pointers into some other collection or values that behave as expected, so I think it should be OK, just to make sure

python/templates/macros/iterator.jinja2
```diff
@@ -3,10 +3,20 @@
...
 class {{ iterator_type }} {
 public:
+  using value_type = {{ class.bare_type }};
+  using difference_type = ptrdiff_t;
+  using reference = {{ prefix }}{{ class.bare_type }};
+  using pointer = {{ prefix }}{{ class.bare_type }}*;
+  using iterator_category = std::input_iterator_tag;
+
   {{ iterator_type }}(size_t index, const {{ class.bare_type }}ObjPointerContainer* collection) : m_index(index), m_object({{ ptr_init }}), m_collection(collection) {}
+  {{ iterator_type }}(): {{ iterator_type }}(0, nullptr) {}
```
tmadlener, line 13: We could default initialize the members in their definitions, then this could also be `= default`, I think.

tmadlener (commented):
I know this is not fully done yet, but I think it would be nice to also document which (classes of) algorithms will be enabled by this (if any).

python/templates/macros/iterator.jinja2
m-fila: Normally we would go with rule 0 and not define anything, but there is a pointer member so this time we should be explicit about the semantics we want. Our pointer is not-owning so we want just the simplest assignment which is exactly what `default` does

python/templates/macros/iterator.jinja2
```diff
@@ -3,27 +3,39 @@
...
   {{ iterator_type }}& operator++();
+  {{ iterator_type }} operator++(int);
 
 private:
-  size_t m_index;
-  {{ prefix }}{{ class.bare_type }} m_object;
-  const {{ class.bare_type }}ObjPointerContainer* m_collection;
+  size_t m_index = 0;
+  {{ prefix }}{{ class.bare_type }} m_object { {{ ptr_init }} };
+  const {{ class.bare_type }}ObjPointerContainer* m_collection = nullptr;
```
tmadlener, line 38:
```suggestion
  size_t m_index{0};
  {{ prefix }}{{ class.bare_type }} m_object { {{ ptr_init }} };
  const {{ class.bare_type }}ObjPointerContainer* m_collection{nullptr};
```
To stay consistent with how we do it throughout the rest of podio and the generated EDMs.

tests/unittests/std_interoperability.cpp
```diff
@@ -895,32 +1090,56 @@ TEST_CASE("Collection and std iterator adaptors", "[collection][container][adapt
...
+      auto coll = CollectionType();
+      coll.create().cellID(42);
+      auto counted = std::counted_iterator(coll.cbegin(), coll.size());
+      REQUIRE(counted != std::default_sentinel);
+      REQUIRE(counted.count() == 1);
+      REQUIRE((*counted).cellID() == 42);
+      DOCUMENTED_STATIC_FAILURE(std::contiguous_iterator<iterator>); // counted-> requires std::contiguous_iterator
+      // REQUIRE(counted->cellID() == 42)
+      DOCUMENTED_STATIC_FAILURE(traits::has_member_of_pointer_v<const const_iterator>); // can't base()->() because
+                                                                                        // base() returns const object
```
tmadlener, line 1139: I don't know enough about the details here, so this is probably a stupid question. Is this something where we are currently missing some syntax, or is this something that is conceptually not possible?

m-fila: That's a very good question :smile: 

It's conceptual. A collection is storing data objects but we want iterators to give user objects. For dereference it's easy becasue a user object can be returned by value, no problem. But `->` operator should return a raw pointer so its pointee must exist somewhere. It can't point to the collection storage because there are no user objects there. So as a workaround the iterators have an additional user object type member to which we can point.
The question is when do we update that member? During `++` would break `end()` (out-of-range access to collection storage to get data object). So this must be during `->` hence it can't be `const`

Strictly speaking there is no requirement for `operator->()` to be `const` in general. It's just the standard adaptors using`const` for their `->` and `base()` for whatever reason. But `*it` and `(*it).member` still works so it's not impossible to use them, just a little confusing

Zehvogel:
can you also take a look at `podio::RelationRange`? :)

m-fila:
Thanks for pointing out, I'll check it out later

tmadlener:
Stupid question:

> *LegacyForwardItertor*: dereferencing two iterators that are equal should give two references bound to the same object, but in podio the returned 'proxies' are different objects

What are the semantic requirements for this? `operator==` returns true when comparing them? Or do they literally have to be a reference to the same object, i.e. the address of the referenced object has to be the same?

If the former is the case, than this should be doable, I think. Because our `operator==` compares the handled `Obj`.

m-fila:
> What are the semantic requirements for this? `operator==` returns true when comparing them? Or do they literally have to be a reference to the same object, i.e. the address of the referenced object has to be the same?
> 
> If the former is the case, than this should be doable, I think. Because our `operator==` compares the handled `Obj`.

Unfortunately it's the second. One of the previous points already states that `*it` must be a reference (as in `&` or `&&`), then 
> If i and j are both dereferenceable, then `i == j` if and only if `*i` and `*j` are bound to the same object.  

---

I think this is ready.
We can have *LegacyInputIterator* and `std::input_iterator` without big changes.
- *LegacyForwardIterator* seems to be fundamentally not compatible with the semantics we have,
- *LegacyOutputIterator* and `std::output_iterator` would require to change the nature of assignment in datatypes - which is a big change that I don't really want to force here.
- `std::forward_iterator` is almost there. If this gets merged then I could separate discussion whether we can get it with some disclaimers.

doc/collections_as_container.md
```diff
@@ -167,17 +167,57 @@ In addition to the *LegacyForwardIterator* the C++ standard specifies also the *
 | `std::back_insert_iterator` | ❗ attention | Compatible only with SubsetCollections, otherwise throws `std::invalid_argument` |
 | `std::front_insert_iterator` | ❌ no | `push_front` not defined |
 | `std::insert_iterator` | ❌ no | `insert` not defined |
-| `std::const_iterator` | ❌ no | `iterator` and `const_iterator` not *LegacyInputIterator* or `std::input_iterator` |
-| `std::move_iterator` | ❌ no | `iterator` and `const_iterator` not *LegacyInputIterator* or `std::input_iterator` |
-| `std::counted_iterator` |  ❌ no | `iterator` and `const_iterator` not `std::input_or_output_iterator` |
-
+| `std::const_iterator` (C++23) | ❌ no | C++23 not supported yet |
```
tmadlener, line 170: Is this only an issue of enabling c++23 in our cmake config, here?

https://github.com/AIDASoft/podio/blob/f6662ad9bf5c202f487b13d70914b5f0b6f29e2a/CMakeLists.txt#L39-L41

doc/collections_as_container.md
```diff
@@ -167,17 +167,57 @@ In addition to the *LegacyForwardIterator* the C++ standard specifies also the *
...
+Algorithms requiring a different iterator category may compile but do not guarantee correct results. An important case are mutating algorithms requiring the iterators to be [*writable*](https://en.cppreference.com/w/cpp/iterator), or [*LegacyOutputIterator*](https://en.cppreference.com/w/cpp/named_req/OutputIterator) or *mutable*, which are not compatible and will produce incorrect results.
+
+For example:
+```c++
+std::find_if(std::begin(collection), std::end(collection), predicate ); // requires InputIterator -> OK
+std::adjacent_find(std::begin(collection), std::end(collection), predicate ); // requires ForwardIterator -> might compile, unspecified result
+std::fill(std::begin(collection), std::end(collection), value ); // requires ForwardIterator and writable -> might compile, wrong result
+std::::sort(std::begin(collection), std::end(collection)); // requires RandomAccessIterator and Swappable -> might compile, wrong result
```
tmadlener, line 204:
```suggestion
std::sort(std::begin(collection), std::end(collection)); // requires RandomAccessIterator and Swappable -> might compile, wrong result
```

tests/unittests/std_interoperability.cpp
```diff
@@ -895,35 +1092,177 @@ TEST_CASE("Collection and std iterator adaptors", "[collection][container][adapt
...
+  subcoll.setSubsetCollection();
+  std::ranges::copy_if(coll, std::back_inserter(subcoll), [](const auto& x) { return x.cellID() > 2; });
+  REQUIRE(subcoll.size() == 2);
+  REQUIRE(subcoll[0].cellID() == 5);
+  REQUIRE(subcoll[1].cellID() == 3);
+
+  // check that algorithms requiring unsupported iterator concepts won't compile
+  DOCUMENTED_STATIC_FAILURE(is_range_adjacent_findable<CollectionType>);
+  DOCUMENTED_STATIC_FAILURE(is_range_sortable<CollectionType>);
+  DOCUMENTED_STATIC_FAILURE(is_range_fillable<CollectionType>);
```
tmadlener, line 1263: Can these be pulled out into a separate test case? two main reasons from my side:
- The concepts could be defined closer to the test case
- We would have an easier way to filter for things that are expected to fail and things that are expected to work

tests/unittests/std_interoperability.cpp
```diff
@@ -783,58 +963,74 @@ TEST_CASE("Collection iterators", "[collection][container][iterator][std]") {
...
-    DOCUMENTED_STATIC_FAILURE(
+    STATIC_REQUIRE(traits::has_dereference_assignment_increment_v<const_iterator, CollectionType::value_type>);
+    STATIC_REQUIRE(
         traits::has_dereference_assignment_increment_v<const_iterator, CollectionType::value_type::mutable_type>);
-    // TODO add runtime check for assignment validity like in '*r = o' case
+    {
+      auto coll = CollectionType{};
+      auto item = coll.create(13ull, 0., 0., 0., 0.);
+      REQUIRE(coll.cbegin()->cellID() == 13ull);
+      auto new_item = CollectionType::value_type::mutable_type{42ull, 0., 0., 0., 0.};
```
tmadlener, line 992: This looks like we should consider adding `mutable_type` directly to the collection as well (I don't think we have yet)?

tests/unittests/std_interoperability.cpp
```diff
@@ -403,33 +411,205 @@ TEST_CASE("Collection AllocatorAwareContainer types", "[collection][container][t
...
+    // TODO check semantic requirements
+  }
+
+  SECTION("contiguous_iterator_iterator") {
+    // iterator
+    DOCUMENTED_STATIC_FAILURE(std::contiguous_iterator<iterator>);
+    // TODO check semantic requirements
+    // const_iterator
     DOCUMENTED_STATIC_FAILURE(std::contiguous_iterator<const_iterator>);
+    // TODO check semantic requirements
```
tmadlener, line 612: Can we group the ones where we already know that we have no chance of getting them with the current design into one section and label that accordingly?

tests/unittests/std_interoperability.cpp
```diff
@@ -1,10 +1,13 @@
...
 #include "datamodel/MutableExampleHit.h"
+#include <algorithm>
 #include <catch2/catch_test_macros.hpp>
+#include <cstddef>
 #include <iterator>
 #include <limits>
 #include <memory>
+#include <ranges>
 #include <stdexcept>
 #include <type_traits>
```
tmadlener, line 12: Can we separate STL, Catch2 and datamodel headers here?

tmadlener (commented):
Thanks a lot for all the work. I mainly have a few minor questions, resp. some (hopefully small) requests for some potential re-organization.

tests/unittests/std_interoperability.cpp
m-fila: We have it in link collections but not in the datatype collections. I guess then we should also think if there should be one in user data collection to keep the  api somehow similar.

But I think this could be done in a separate PR as it's not touching our compliance to iterator specification

doc/collections_as_container.md
m-fila: I was a bit lazy and wanted to skip that part :sweat_smile:
Added checks. It seems that `std::const_iterator` should be fine except that `->` can't be used as `->` in our iterators can't be `const`

tests/unittests/std_interoperability.cpp
tmadlener: Yes, definitely a separate PR.

doc/collections_as_container.md
tmadlener: Ah sorry, I didn't want to imply more work. I was mainly wondering because of the comment and if that would have been the only reason, it would have been an easy fix.

m-fila: no problem, thanks to this I learned some more new C++23 stuff :slightly_smiling_face:

Issue #631, opened 2024-06-22
OneToOne relations are not preserved on `clone()` since v01-00


veprbl:
!!!IMPORTANT!!!: To facilitate faster and easier response to your issue please provide in addition to the description of the issue also the following information

 - OS version: e.g. SLC6
 - Compiler version: e.g GCC 6.2
 - PODIO version: 01-00
 - Reproduced by: 

sorry, no exact steps, but the idea is we start with:
```yaml
  edm4eic::MCRecoTrackParticleAssociation:
    Description: "Association between a Track and a MCParticle"
    Author : "S. Joosten"
    OneToOneRelations:
      - edm4hep::MCParticle sim             // reference to the Monte-Carlo particle
```
then we fill that with both `rec` and `sim` set.
```c++
MCRecoTrackParticleAssociationCollection *partassocs_in;
for (auto assoc_in : *partassocs_in) {
  // assoc_in.getSim().isAvailable() == 1
  auto assoc_out = assoc_in.clone();
  // assoc_out.getSim().isAvailable() == 0 <- BUG
}
```

 - Input: link to input files if applicable
 - Output: full build and run log output
 - Goal: running existing eicrecon code

Bisect shows that 02a4b9d5d704eb572ce945594b4c85303e89e216 is the first bad commit

jmcarcell:
Hi @veprbl, feel free to test #632, it should be fixed there.

PR #632, opened 2024-06-23
Fix one to one relations for cloned objects

jmcarcell:
Fixes https://github.com/AIDASoft/podio/issues/631. In https://github.com/AIDASoft/podio/pull/583, one to one relations were not added together with the one to many relations.

BEGINRELEASENOTES
- Fix one to one relations for cloned objects by copying the one to one relations too.
- Add a test that checks values for cloned objects from a non mutable object.

ENDRELEASENOTES

### DIFF
--- a/python/templates/Object.cc.jinja2
+++ b/python/templates/Object.cc.jinja2
@@ -16,7 +16,7 @@
 
 {{ utils.namespace_open(class.namespace) }}
 
-{{ macros.constructors_destructors(class.bare_type, Members, OneToManyRelations + VectorMembers) }}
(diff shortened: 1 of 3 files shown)


veprbl (approved):
This works, thanks.

tmadlener:
@veprbl I will make a new v1.0.1 tag with these fixes in, unless there is more that is breaking on your side.

veprbl:
@tmadlener That would be nice to have. There are no other regressions that I'm aware of.

veprbl:
Ah, I see 01-00-01 is a thing already. Thanks.

PR #633, opened 2024-06-23
Pass an argument by const reference instead of value and other minor fixes

jmcarcell:
BEGINRELEASENOTES
- Minor cleanups for generated EDMs
  - Pass the argument by const reference in the relation setter
  - Remove a few includes that are not needed
  - Don't check for `nullptr` before deleting a (possibly `nullptr`) pointer
  - Remove `if - else` check and use implicit conversion of pointer to bool
 
ENDRELEASENOTES

### DIFF
--- a/python/templates/macros/implementations.jinja2
+++ b/python/templates/macros/implementations.jinja2
@@ -123,7 +123,7 @@ const {{ relation.full_type }} {{ class_type }}::{{ relation.getter_name(get_syn
 {% macro single_relation_setters(class, relations, get_syntax, prefix='') %}
 {% set class_type = prefix + class.bare_type %}
 {% for relation in relations %}
-void {{ class_type }}::{{ relation.setter_name(get_syntax) }}({{ relation.full_type }} value) {
+void {{ class_type }}::{{ relation.setter_name(get_syntax) }}(const {{ relation.full_type }}& value) {
   if (m_obj->m_{{ relation.name }}) {
     delete m_obj->m_{{ relation.name }};
   }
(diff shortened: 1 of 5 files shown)


python/templates/macros/implementations.jinja2
```diff
@@ -118,7 +118,7 @@ const {{ relation.full_type }} {{ class_type }}::{{ relation.getter_name(get_syn
 {% macro single_relation_setters(class, relations, get_syntax, prefix='') %}
 {% set class_type = prefix + class.bare_type %}
 {% for relation in relations %}
-void {{ class_type }}::{{ relation.setter_name(get_syntax) }}({{ relation.full_type }} value) {
+void {{ class_type }}::{{ relation.setter_name(get_syntax) }}(const {{ relation.full_type }}& value) {
   if (m_obj->m_{{ relation.name }}) {
     delete m_obj->m_{{ relation.name }};
   }
```
m-fila, line 124: Another case of unnecessary `nullptr` check before delete
```suggestion
  delete m_obj->m_{{ relation.name }};
```

tmadlener:
Looks like CERNs infrastructure is slightly overwhelmed, CI is failing because we cannot get cvmfs or because fetching test data is failing.

m-fila:
Another minor fix that could be added:
this could be simiplified to just return `m_obj`
https://github.com/AIDASoft/podio/blob/d12f4df81591e54e7750500f5d6ae70bad7ecc39/python/templates/macros/implementations.jinja2#L184-L189

tmadlener:
Can you pick up the changes from #634 (i.e. merge / rebase onto latest master)? I think that should fix the remaining CI issues that we see.

tmadlener (approved):

jmcarcell:
> Can you pick up the changes from #634 (i.e. merge / rebase onto latest master)? I think that should fix the remaining CI issues that we see.

Done

tmadlener (approved):

PR #638, opened 2024-07-10
Make ROOT load libraries without the .so extension

jmcarcell:
BEGINRELEASENOTES
- Make ROOT load libraries without the .so extension

ENDRELEASENOTES

See https://github.com/key4hep/EDM4hep/pull/339 and https://github.com/key4hep/EDM4hep/issues/338. This means no one is running the tests on MacOS :smile:

### DIFF
--- a/doc/python.md
+++ b/doc/python.md
@@ -10,7 +10,7 @@ To load the Python bindings from a generated C++ model dictionary, first make su
 ```python
 import ROOT
 
-res = ROOT.gSystem.Load('libGeneratedModelDict.so')
+res = ROOT.gSystem.Load('libGeneratedModelDict')
 if res < 0:
     raise RuntimeError('Failed to load libGeneratedModelDict.so')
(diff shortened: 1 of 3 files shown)


tmadlener:
Well, maybe someone has, but they ignored the failing tests ;)

There is another error message with an explicit `.so` here:

https://github.com/AIDASoft/podio/blob/1fc075df881f1b4c58f7e85b53aecb14fcba1e24/python/podio/__init__.py#L10

jmcarcell:
I don't have a mac handy to check but maybe it doesn't matter because of https://github.com/AIDASoft/podio/blob/master/CMakeLists.txt#L52

tmadlener:
Ah yes, that probably also affects mac OS builds. Not sure if mac OS still allows shared libs with .so ending though. Anyhow, for this PR it should not matter too much as long as ROOT does the "right thing".

PR #640, opened 2024-07-12
Make sure subset collections do not duplicate information when dumped to JSON

tmadlener:
BEGINRELEASENOTES
- Make JSON dumping of subset collections only store the `ObjectID`s to the elements they are pointing to instead of duplicating the information.
- Add JSON conversion support to `podio::ObjectID`
  - Use this to simplify JSON conversion code for relations in objects

ENDRELEASENOTES

### DIFF
--- a/include/podio/ObjectID.h
+++ b/include/podio/ObjectID.h
@@ -5,6 +5,10 @@
 #include <iomanip>
 #include <ostream>
 
+#if defined(PODIO_JSON_OUTPUT) && !defined(__CLING__)
+  #include "nlohmann/json.hpp"
+#endif
+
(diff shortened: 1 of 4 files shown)


include/podio/ObjectID.h
```diff
@@ -5,6 +5,10 @@
 #include <iomanip>
 #include <ostream>
 
+#if defined(PODIO_JSON_OUTPUT) && !defined(__CLING__)
```
hegner, line 8: we have to watch out for this combination of flags. If we move it more often, we should replace it with a single combined one.

hegner (approved):

PR #646, opened 2024-07-24
Add a LinkNavigator utility

tmadlener:
BEGINRELEASENOTES
- Add a `LinkNavigator` utility class that facilitates the lookup of linked objects

ENDRELEASENOTES

The main working principle is:
```cpp
// From MC to reco particles
auto& mc2recoLinks = event.get<RecoMCParticleLinkCollection>("MCRecoTruthLink");
auto mc2reco = LinkNavigator{mc2recoLinks};

auto& mcps = event.get<MCParticle>("MCParticles");
for (const auto p : mcps) {
  const auto recos = mc2reco.getLinked(p);
}
```
`recos` in this case is a `std::vector<WeightedObject<ReconstructedParticle>>`, which effectively just couples an object and a relation weight together. The foreseen usage is something like this:
```cpp
for (const auto p : mcps) {
  for (const auto& [reco, weight] : mc2reco.getLinked(p)) {
    // do something with reco and the weight
  }
}
```

- [x] Needs (includes) #255 
- [x] Documentation
- [x] Handling of `std::is_same_v<FromT, ToT>`


tmadlener:
This is now also ready for review and I would like to merge it before the next tag.

doc/links.md
```diff
@@ -136,6 +136,40 @@ and have that compiled into the library. This is necessary if you want to use
...
+
+```cpp
+const auto& recoMcLinks = event.get<edm4hep::RecoMCParticleLinkCollection>("mcrecolinks");
+const auto linkNavigator = podio::LinkNavigator(recoMcLinks);
+
+// For podio::LinkCollections with disparate types just use getLinked
+const auto linkedRecs = linkNavigator.getLinked(mcParticle);
+// If you want to make a point about the direction use getLinked{From,To}
+// This is also necessary if the From and To type in the link collection are the same
+const auto linkedMCs = linkNavigator.getLinkedFrom(recoParticle);
```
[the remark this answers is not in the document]
tmadlener: Yes, exactly. I started the naming for these from the arguments. Maybe we can drop that detail from the documentation as well, as for the majority of links, I assume the templated version will do correct thing in any case. 

Alternatively, we could also introduce a two argument version that takes the direction as some enum?

include/podio/LinkNavigator.h
```diff
@@ -0,0 +1,147 @@
...
+  /// @param object The object that is labeled *to* in the link
+  ///
+  /// @returns A vector of all objects and their weights that are linked to
+  ///          the passed object
+  template <typename ToU = ToT>
+  std::enable_if_t<!std::is_same_v<FromT, ToU>, std::vector<WeightedObject<FromT>>> getLinked(const ToT& object) const {
+    return getLinkedTo(object);
+  }
+
+  /// Get all the objects and weights that are linked to the passed object
```
tmadlener, line 96:
```suggestion
  /// Get all the objects and weights that are linked from the passed object
```
(?)

doc/links.md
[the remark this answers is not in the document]
andresailer: I think a two argument version of `getLinked` with an enum sounds less confusing and more explicit than `getLinkedFrom`.

include/podio/LinkNavigator.h
```diff
@@ -0,0 +1,147 @@
...
+
+  /// We do only construct from a collection
+  LinkNavigator() = delete;
+  LinkNavigator(const LinkNavigator&) = default;
+  LinkNavigator& operator=(const LinkNavigator&) = default;
+  LinkNavigator(LinkNavigator&&) = default;
+  LinkNavigator& operator=(LinkNavigator&&) = default;
+  ~LinkNavigator() = default;
+
+  /// Get all the objects and weights that are linked to the passed object
```
andresailer, line 60:
```suggestion
  /// Get all the *From* objects and weights that have links with the passed object
```

include/podio/LinkNavigator.h
```diff
@@ -0,0 +1,167 @@
...
+  };
+
+  /// Simple struct tag for overload selection in LinkNavigator below
+  struct LookupFromTag {};
+  /// Simple struct tag for overload selection in LinkNavigator below
+  struct LookupToTag {};
+} // namespace detail::links
+
+/// Tag variable to select the lookup of *From* objects that are linked from a
+/// *To* object in podio::LinkNavigator::getLinked
```
andresailer, line 37:
```suggestion
/// Tag variable to select the lookup of *From* objects that have links with the given
/// *To* object in podio::LinkNavigator::getLinked
```

include/podio/LinkNavigator.h
```diff
@@ -0,0 +1,167 @@
...
+static constexpr detail::links::LookupFromTag LookupFrom;
+/// Tag variable to select the lookup of *To* objects that are linked from a
+/// *From* object in podio::LinkNavigator::getLinked
```
andresailer, line 40:
```suggestion
/// Tag variable to select the lookup of *To* objects that have links with the given
/// *From* object in podio::LinkNavigator::getLinked
```

doc/links.md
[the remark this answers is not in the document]
andresailer: I proposed some word changes to avoid re-using to and from in the documentation strings.

I think instead of `Lookup` `Return` might be more explicit.

`getLinked(recoParticle, ReturnFrom);`

makes it more obvious that we return `from` objects?

tmadlener:
Thanks for the wording suggestions. Unless I missed some instances (which is not unlikely), I think it should now all be more or less self-consistent.

include/podio/LinkNavigator.h
```diff
@@ -0,0 +1,171 @@
...
+  LinkNavigator(const LinkNavigator&) = default;
+  LinkNavigator& operator=(const LinkNavigator&) = default;
+  LinkNavigator(LinkNavigator&&) = default;
+  LinkNavigator& operator=(LinkNavigator&&) = default;
+  ~LinkNavigator() = default;
+
+  /// Get all the *From* objects and weights that have links with the passed
+  /// object
+  ///
+  /// You will get this overload if you pass the podio::LookupFrom tag as second
```
andresailer, line 75:
```suggestion
  /// You will get this overload if you pass the podio::ReturnFrom tag as second
```

include/podio/LinkNavigator.h
```diff
@@ -0,0 +1,171 @@
...
+  ///          the passed object
+  template <typename ToU = ToT>
+  std::enable_if_t<!std::is_same_v<FromT, ToU>, std::vector<WeightedObject<FromT>>> getLinked(const ToT& object) const {
+    return getLinked(object, podio::ReturnFrom);
+  }
+
+  /// Get all the *To* objects and weights that have links with the passed
+  /// object
+  ///
+  /// You will get this overload if you pass the podio::LookupTo tag as second
```
andresailer, line 117:
```suggestion
  /// You will get this overload if you pass the podio::ReturnTo tag as second
```

doc/links.md
```diff
@@ -136,6 +136,44 @@ and have that compiled into the library. This is necessary if you want to use
...
+```
+
+If you want to be explicit about the lookup direction, e.g. in case you have a
+link that has the same `From` and `To` type, you can use the overloads that take
+a second *tag argument*:
+```cpp
+const auto linkedMCs = linkNavigator.getLinked(recoParticle, podio::ReturnTo);
+```
+
+The return type of all methods is a `std::vector<WeightedObject>`, where the
```
tmadlener, line 164:
```suggestion
The return type of all methods is a `std::vector<podio::detail::links::WeightedObject>`, where the
```
With the implicit question whether we should lift this object out of the (implicitly private) `detail` namespace, since it is in principle user facing.

tmadlener: Actually we pull the `WeightedObject` into the `podio` top level namespace, so it is in fact actually public.

Issue #647, opened 2024-07-24
Support for reading multiple files with SIO


tmadlener:
Currently the `SIOReader` does not support automatic file switching.

SaxenaAnushka102:
Hello @tmadlener! I'd love to work on this. Could you please describe the issue a in a bit more detail? As per my understanding, any new functions in this context will be defined in `SIOReader.h` and the implementation would be done in the `SIOReader.cc`.  Does this work?

SaxenaAnushka102:
Hi @tmadlener,
Thanks so much for the detailed explanation! While it sounds like there's a bit more complexity than I initially anticipated, I'm still very interested in working on this and getting some hands-on experience with the `SIOReader` code.
To get started, I had a couple of questions for clarification:
In the new structure `EntryFileMap`, how will it handle situations where an entry might span multiple files?
For the `openFiles` method, is there a specific way we want to handle potential errors during the TOC record reading process for each file?
I'm happy to jump in and start working on the structure that keeps track of entries and files (potentially `EntryFileMap`).  Do you have any specific guidance on the implementation approach for this part? Also, are there any project community calls that I can be a part of to know more & discuss this further? 
Thanks,
Anushka.

tmadlener:
Glad to hear it :) And obviously very happy to help along the way. To answer your specific questions:

> In the new structure `EntryFileMap`, how will it handle situations where an entry might span multiple files?

Entries will always be fully within one file, unless there is a problem with the file. However, in that case there is not too much we can do in any case.

> For the openFiles method, is there a specific way we want to handle potential errors during the TOC record reading process for each file?

For the first version we simply go the easiest way; If we find an issue in any of the files during TOC reading, we will just throw an exception, as there is again not too much we can do with faulty input files.

> Do you have any specific guidance on the implementation approach for this part?

I would start with something a long the lines in my original comment. I am fairly certain it will need some refinement that we will discover throughout the development.

> Also, are there any project community calls that I can be a part of to know more & discuss this further?

No, not really at this point.

PR #648, opened 2024-07-24
Document I/O interfaces

tmadlener:
BEGINRELEASENOTES
- Add docstrings to the `Reader` and `Writer` interface classes
- Make sure that `makeReader` throws an exception if multiple SIO files are passed (see [#647](https://github.com/AIDASoft/podio/issues/647))

ENDRELEASENOTES

### DIFF
--- a/include/podio/Reader.h
+++ b/include/podio/Reader.h
@@ -12,7 +12,8 @@ namespace podio {
 /// Offers some more high level functionality compared to the lower level
 /// backend specific readers that this class wraps. In contrast to the lower
 /// level readers that usually return arbitrary FrameData, this interface class
-/// will return fully constructed Frames.
+/// will return fully constructed Frames. In addition, it provides convenience
+/// methods to deal specifically with the "events" frame category.
 ///
 /// @note The recommended way to construct is to use the makeReader() functions
 /// since they handle the instantiation of the correct low level readers
@@ -216,7 +217,7 @@ Reader makeReader(const std::string& filename);
 ///
 /// @throws std::runtime_error in case the file extensions differ or in case
 ///         support for the necessary I/O backend has not been built
-Reader makeReader(const std::vector<std::string>& filename);
+Reader makeReader(const std::vector<std::string>& filenames);
 
 } // namespace podio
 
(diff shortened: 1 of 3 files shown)


include/podio/Reader.h
```diff
@@ -73,50 +83,139 @@ class Reader {
...
+/// @note For SIO files this will only read the first file!
+///
+/// @param filenames The (paths to the) files to read from
+///
+/// @returns A Reader that has been initialized and that can be used for reading
+///          data from the passed files
+///
+/// @throws std::runtime_error in case the file extensions differ or in case
+///         support for the necessary I/O backend has not been built
 Reader makeReader(const std::vector<std::string>& filename);
```
hegner, line 219: actually it should be `filenames`, not `filename`

include/podio/Reader.h
```diff
@@ -73,50 +83,139 @@ class Reader {
...
 Reader makeReader(const std::string& filename);
+
+/// Create a Reader that is able to read the files
+///
+/// This will inspect the filenames as well as peek into the **first file only**
+/// to decide based on the contents which low level reader to instantiate for
+/// reading. All files are assumed to be of the same I/O format, no switching
+/// between formats is possible.
+///
+/// @note For SIO files this will only read the first file!
```
hegner, line 210: we should put in some exception handling or something of that kind. That is too burried in the docs.

include/podio/Reader.h
```diff
@@ -6,6 +6,16 @@
...
 namespace podio {
 
+/// Generic (type erased) reader class that can handle different I/O backends
+/// transparently
+///
+/// Offers some more high level functionality compared to the lower level
+/// backend specific readers that this class wraps. In contrast to the lower
+/// level readers that usually return arbitrary FrameData, this interface class
+/// will return fully constructed Frames.
+///
```
hegner, line 16: In addition it provides convenience methods to deal specifically with the event frame category...

hegner (changes requested):
Thanks a lot!
Just a few minor remarks. Except for the exception handling they could be covered in the same PR

include/podio/Reader.h
tmadlener: You mean, throw an exception when more than one SIO file is passed?

hegner: we should not silently ignore faulty states.

hegner: so yes

tmadlener:
Thanks for checking. I think I addressed all of them.

hegner (approved):
That was quick! thanks

PR #651, opened 2024-07-31
Add a hook to inject datamodel version information into podio internals

### LINKED ISSUE #630: Make it possible to store more version information of generated EDMs
We currently store the schema version of all datatypes that we encounter during writing, but it should also be possible to add the "package version" of an EDM to the file. E.g. EDM4hep has an `edm4hepVersion.h` header similar to what we have in podio, which defines a version (not entirely) independent of the schema version. There should be some hook that makes it possible to easily pick up those versions as well and add it to the metadata.

tmadlener:
BEGINRELEASENOTES
- Make it possible to pass the datamodel version (which does not have to be the schema version!) to the class generator to inject it into the podio internals. This will automatically be stored in all output files, such that this information can also be obtained from the files.

ENDRELEASENOTES

@peremato this would make e.g. the EDM4hep version appear in the output files (once the few lines are added to the `edm4hep.yaml` file and its version header).

Conceptually, this relies on the fact that a datamodel will have a version header of some form and will just try to reuse that. The main requirements are that the version class of the datamodel is convertible to a`podio::version::Version`.

Fixes #630 

- [x] Includes #659


m-fila:
Wouldn't it be more straightforward to just add a simple field for version in the yaml and then relay on a build system to populate it somehow (e.g. `configure_file` in cmake)? A version header file will be most likely generated by build system too

tmadlener:
I would like to avoid having to `configure_file` the full datamodel yaml file. Then I went with something that would allow sort of minimal work in EDM4hep, where we already have a version header that we configure via the build system. In the end this is a first proposal, so I am very open for suggestions.

Another possibility I thought about was to configure a smaller YAML file with the version information (but then we might be configuring quite a few files in the end).

In the end how the information is injected exactly is rather flexible in this PR, so it should be fairly straight forward to do what we want to do.

tmadlener:
I am not sure if it's an argument in any direction, but I would at least like to point it out (and write it down for posteriority): If the version information is not part of the YAML configuration, the "roundtripped generated code" might be different because the generated code now depends on some pretty much ephemeral version input. On the other hand if we pull in a c++ variable (like we did in the first approach) we have the same code, but potentially different contents, because the value of that variable can change without visible changes in the generated code.

tmadlener:
As discussed during [todays EDM4hep meeting](https://indico.cern.ch/event/1455240/), this PR effectively imposes that packages that use podio follow SemVer as we use that format to store the information. There is no real technical blocker to store the version as e.g. a string. However, we will defer implementing that until someone actually requests that feature.

python/podio_gen/cpp_generator.py
```diff
@@ -505,12 +514,14 @@ def _write_all_collections_header(self):
     def _write_edm_def_file(self):
         """Write the edm definition to a compile time string"""
         model_encoder = DataModelJSONEncoder()
+        print(f"{self.datamodel_version=}")
```
jmcarcell, line 517:
```suggestion
```
Debugging print? Otherwise it's going to be a bit weird like that whenever it appears with `self.`

tmadlener: Ah, few seconds too late. This is indeed a debug print that I missed while having another look.

peremato:
@tmadlener I am a bit confused. What is the "datamodel version"? I can see in the generated files the PODIO version and the EDM4hep schema version but not datamodel. See what I print using the Julia interface: 
```
┌────────────────┬─────────────────────────┐
│ Atribute       │ Value                   │
├────────────────┼─────────────────────────┤
│ File Name(s)   │ test/EDM4hep_TTree.root │
│ # of events    │ 3                       │
│ IO Format      │ TTree                   │
│ EDM4hep schema │ 2.0.0                   │
│ PODIO version  │ 1.0.1                   │
│ ROOT version   │ 6.32.4                  │
└────────────────┴─────────────────────────┘ 
```

tmadlener:
The datamodel version would be the tag for EDM4hep. With the current `main` branch of EDM4hep you would get `0.99.0`. However, you will also need https://github.com/key4hep/EDM4hep/pull/362 first, because that is necessary to pass that information to podio in the first place.

PR #567, opened 2024-03-10
Clean up headers with include-what-you-use

jmcarcell:
BEGINRELEASENOTES
- Clean up headers with include-what-you-use, second part after https://github.com/AIDASoft/podio/pull/652

ENDRELEASENOTES

PR #652, opened 2024-08-07
Remove unnecessary headers

jmcarcell:
#567 is a mess so let's divide it in the deletions and additions and then it will be easier to see what's going on. In addition, the deletions should be quite uncontroversial.

BEGINRELEASENOTES
- Remove unnecessary headers

ENDRELEASENOTES

### DIFF
--- a/tests/unittests/interface_types.cpp
+++ b/tests/unittests/interface_types.cpp
@@ -2,9 +2,7 @@
 
 #include "podio/ObjectID.h"
 
-#include "datamodel/AnotherTypeWithEnergy.h"
-#include "datamodel/EnergyInNamespaceCollection.h"
-#include "datamodel/ExampleClusterCollection.h"
+#include "datamodel/MutableExampleCluster.h"
 #include "datamodel/ExampleHitCollection.h"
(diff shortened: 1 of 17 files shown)


andresailer:
Too much removed, breaking the build against LCG_105?

jmcarcell:
Now fixed

tmadlener:
pre-commit workflow should be fixed with #654. Can you pick that up?

tests/unittests/interface_types.cpp
```diff
@@ -2,9 +2,7 @@
 
 #include "podio/ObjectID.h"
 
-#include "datamodel/AnotherTypeWithEnergy.h"
-#include "datamodel/EnergyInNamespaceCollection.h"
-#include "datamodel/ExampleClusterCollection.h"
+#include "datamodel/MutableExampleCluster.h"
 #include "datamodel/ExampleHitCollection.h"
```
tmadlener, line 6:
```suggestion
#include "datamodel/ExampleHitCollection.h"
#include "datamodel/MutableExampleCluster.h"
```
clang-format wants them ordered alphabetically

PR #657, opened 2024-08-23
[WIP] doc: Make automatic deployment of versioned documentation possible

tmadlener:
BEGINRELEASENOTES
- Add the necessary template to enable version navigation to the generated documentation
  - Dynamically determine available versions from the already deployed tags
- Adapt the gitlab ci to push documentation to a dedicated folder for each tag

ENDRELEASENOTES

This seems to be a bit convoluted, but I could not get any other version that claims to do automatic generation of versioned docs to work properly for podio, because they usually simply check out the repository for a given version and then do the doc generation from the code. However, we would need to run the build step again for each tag, which these tools don't do, plus it would get more expensive to do with every version. Hence, I went for a solution where we can simply add new content / tags and it will by picked up automatically.

It is possible to add older versions manually to this structure and new tags will simply be picked up automatically.

Overall the process is working and my playing around / testing can be tried [here](https://key4hep.web.cern.ch/podio/_contents/versioned-docs/) (the documentation is the same in both cases, it just shows how the version navigation looks like in the end).

There are a few assumptions in the dynamic determination of available versions, see the README for the gist of it.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -73,7 +73,9 @@ deployment:
 
   # see see https://gitlab.cern.ch/ci-tools/ci-web-deployer
   image: gitlab-registry.cern.ch/ci-tools/ci-web-deployer:latest
+  before_script:
+    - if [ -z "$CI_COMMIT_BRANCH" ]; then TARGET_DEPLOY_DIR=$CI_COMMIT_TAG; else TARGET_DEPLOY_DIR=$CI_COMMIT_BRANCH; fi
+    - export EOS_PATH=$EOS_DEPLOY_BASE_PATH/test_deploy/$TARGET_DEPLOY_DIR
   script:
     - deploy-eos
-  before_script: []
   after_script: []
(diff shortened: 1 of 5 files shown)


.gitlab-ci.yml
```diff
@@ -62,13 +62,18 @@ deployment:
   only:
     refs:
       - master
+      - versioned-docs
```
tmadlener, line 65:
```suggestion
```
Leftover from testing / developing

doc/web-deploy/README.md
```diff
@@ -0,0 +1,28 @@
...
+│   ├── vXX-YY
+│   └── master
+├── index.html
+```
+  - Each version / tag gets a new folder and the `master` / develop version is
+    continually updated
+  - The `_contents` folder has to configured to have *Directory Browsing* enabled
+    since that will be used to determine the available versions (each top-level
+    folder in here will be considered a version). This is achieved via `echo
+    "Options +Indexes > _contents/.htaccess"`
```
tmadlener, line 16:
```suggestion
    "Options +Indexes" > _contents/.htaccess"`
```

PR #662, opened 2024-09-05
Remove the check for different filesystem libs and fallback to Boost

tmadlener:
BEGINRELEASENOTES
- Remove the cmake checks to figure out against which filesystem library to link, since all major compilers have had it properly integrated into their STLs for quite some time now. Also remove the fallback to Boost.

ENDRELEASENOTES


cmake/podioMacros.cmake
```diff
@@ -371,37 +371,3 @@ function(PODIO_ADD_SIO_IO_BLOCKS CORE_LIB HEADERS SOURCES)
...
-    ENDIF()
-    MESSAGE(STATUS "Compiler not compatible when linking against '${FS_LIB_NAME}'")
-  ENDFOREACH()
-
-  MESSAGE(STATUS "Compiler does not have filesystem support, falling back to boost::filesystem")
-  find_package(Boost REQUIRED COMPONENTS filesystem system)
-  SET(${FS_LIBS} Boost::filesystem Boost::system PARENT_SCOPE)
-  SET_TARGET_PROPERTIES(Boost::filesystem
-    PROPERTIES
-    INTERFACE_COMPILE_DEFINITIONS USE_BOOST_FILESYSTEM
```
andresailer, line 405: That means we can also drop
https://github.com/AIDASoft/podio/blob/98a786a47443eb324cae27b1b24c013c315989e1/src/SIOBlock.cc#L8
and
https://github.com/AIDASoft/podio/blob/98a786a47443eb324cae27b1b24c013c315989e1/src/SIOBlock.cc#L136

tmadlener: See #664

PR #663, opened 2024-09-06
Add pythonization for freezing class attribute and preventing creating new attrubes by mistake

m-fila:
BEGINRELEASENOTES
- Added pythonization for "freezing" class disallowing setting non-existent attributes

ENDRELEASENOTES

Pythonization proposed in key4hep/EDM4hep#357. Adding new attributes to existing objects is a feature in python. Unfortunately in the context of podio this is a very error prone behaviour that already lead to some bugs in the past.

For instance a typo in attribute name will silently create new attribute. With this pythonization an exception will be thrown
The error message is consistent with `__slots__` which is conventionally used in standard python to construct classes with frozen attributes

### DIFF
--- a/python/podio/pythonizations/freeze_class.py
+++ b/python/podio/pythonizations/freeze_class.py
@@ -1,0 +1,31 @@
+"""Prevent creating new attributes for existing objects
+The new attributes created in Python won't be visible for podio IO
+therefore preventing the addition of new attributes for podio objects
+might be desirable and help detecting miss-assignments"""
+
+from .utils.pythonizer import Pythonizer
+
+
+class FreezeClassPythonizer(Pythonizer):
+    """Prevent setting new attributes"""
+
+    @classmethod
+    def priority(cls):
+        """This most likely should be the last pythonization loaded
+        otherwise it may interfere with creating attributes during other pythonizations"""
+        return 99
+
+    @classmethod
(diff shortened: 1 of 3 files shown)


hegner (approved):

hegner:
@m-fila - always on the wish list and you did it - great! :-)

tmadlener (commented):
Just for my understanding. This is still opt-in for generated EDMs, right? And I only don't see the "opting in" in the diff of this PR, because for podio tests we already have opted in with the other pythonizations?

Is it possible to selectively opt in to the pythonizations?

m-fila:
Yes, it's opt-in. That test opted for all the pythonizations in podio for namespace/module `nsp`

https://github.com/AIDASoft/podio/blob/27e92d5b086554247ca84fac6557873c712735c5/python/podio/test_CodeGen.py#L8-L12

Loading specific pythonization is also supported

```python
from  podio.pythonizations import freeze_class

freeze_class.FreezeClassPythonizer.register("edm4hep")
```

The docs are already there https://key4hep.web.cern.ch/podio/python.html#pythonizations :wink:

tmadlener:
> The docs are already there https://key4hep.web.cern.ch/podio/python.html#pythonizations 😉

:joy::joy: you would think people look there first before asking stupid questions ;)

Thanks for the clarifications.

PR #665, opened 2024-09-06
add editorconfig

m-fila:
BEGINRELEASENOTES
- Added .editorconfig for basic cross-editor configuration for generic files

ENDRELEASENOTES

Editorconfig is a popular tool for portable configuration of editors. It's automatically recognized by some editors (vim, neovim, github editor) or has a plugin for the others (emacs, vscode, clion, sublime). The tool takes lower precedence than any language specific config so there shouldn't be conflicts

Here I propose a very basic config that will automatically add newline at the end of any file so hopefully there will be less small corrections like in https://github.com/AIDASoft/podio/pull/660#discussion_r1743364648

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+# https://editorconfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true


hegner:
I would not like adding it for similar reasons.

m-fila:
ok, closing

jmcarcell:
I'm not opposed to adding it, I just don't think it's going to be very useful for people that make those mistakes

tmadlener:
I think it could be useful given the VSCode support and that that is what many people nowadays use. For those it would take care of a few otherwise annoying things. The rest probably has their editor configured in some other way in any case, so for them it doesn't do any harm either.

PR #666, opened 2024-09-06
Add two checks not to crash when a file doesn't exist or can't be opened

jmcarcell:
BEGINRELEASENOTES
- Add two checks not to crash when a file doesn't exist or can't be opened and when an empty list is passed to makeReader.

ENDRELEASENOTES


jmcarcell:
Yes, right now it crashes if the file doesn't exist because `file` is a `nullptr`.

tmadlener (approved):
:+1: Thanks.

PR #667, opened 2024-09-06
fix importing submodules in pythonizations

m-fila:
BEGINRELEASENOTES
- Fix pythonizations trying to import local modules as pythonizations submodules

ENDRELEASENOTES

`load_pythonizations` function was by accident sometimes trying to import some local files of downstream projects as submodules of `pythonization` module (https://github.com/key4hep/EDM4hep/pull/290#issuecomment-2333857243)

The culprit is the function `pkgutil.walk_modules` that tries to recursively find submodules on a given path. `utils` is a legit submodule of `pythonizations` found by `walk_modules`,  but then it will try to recursively find submodules somehow including `utils` also among the local modules if present. I'm not sure whether this is intended or actual bug in `pkgutils`

Anyway I replace the `walk_packages` with `iter_modules` that doesn't have this weird recursive behaviour. The "downside" is that now the pythonizations must be put directly in pythonizations directory with no nesting (which was the case anyway)

### DIFF
--- a/python/podio/pythonizations/__init__.py
+++ b/python/podio/pythonizations/__init__.py
@@ -1,13 +1,16 @@
 """cppyy pythonizations for podio"""
 
 from importlib import import_module
-from pkgutil import walk_packages
+from pkgutil import iter_modules
+from os import path
 from .utils.pythonizer import Pythonizer
 
 
 def load_pythonizations(namespace):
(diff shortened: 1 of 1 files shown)


tmadlener (approved):
Merging this. Then https://github.com/key4hep/EDM4hep/pull/290 should pick it up in the CI workflows

PR #668, opened 2024-09-08
Change C-style casts to static_cast and remove unnecessary semicolons

jmcarcell:
I have been playing a bit with some warnings in Key4hep and I get some from included files from podio. The C-style cast may be useful. I guess these don't hurt anyway.

BEGINRELEASENOTES
- Change C-style casts to static_cast and remove unnecessary semicolons

ENDRELEASENOTES

### DIFF
--- a/tests/read_test.h
+++ b/tests/read_test.h
@@ -39,8 +39,7 @@ bool check_fixed_width_value(FixedWidthT actual, FixedWidthT expected, const std
 void processEvent(const podio::Frame& event, int eventNum, podio::version::Version fileVersion) {
   const float evtWeight = event.getParameter<float>("UserEventWeight").value();
   if (evtWeight != 100.f * eventNum) {
-    std::cout << " read UserEventWeight: " << evtWeight << " - expected : " << 100.f * eventNum
-              << std::endl;
+    std::cout << " read UserEventWeight: " << evtWeight << " - expected : " << 100.f * eventNum << std::endl;
     throw std::runtime_error("Couldn't read event meta data parameters 'UserEventWeight'");
   }
 
--- a/tests/read_test.h
+++ b/tests/read_test.h
@@ -20,12 +20,10 @@
 
 // STL
 #include <cassert>
-#include <exception>
 #include <iostream>
 #include <limits>
 #include <sstream>
 #include <stdexcept>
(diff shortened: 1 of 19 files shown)


tests/read_test.h
```diff
@@ -40,8 +38,9 @@ bool check_fixed_width_value(FixedWidthT actual, FixedWidthT expected, const std
 
 void processEvent(const podio::Frame& event, int eventNum, podio::version::Version fileVersion) {
   const float evtWeight = event.getParameter<float>("UserEventWeight").value();
-  if (evtWeight != (float)100. * eventNum) {
-    std::cout << " read UserEventWeight: " << evtWeight << " - expected : " << (float)100. * eventNum << std::endl;
+  if (evtWeight != static_cast<float>(100. * eventNum)) {
+    std::cout << " read UserEventWeight: " << evtWeight << " - expected : " << static_cast<float>(100. * eventNum)
```
tmadlener, line 42:
```suggestion
  if (evtWeight != 100.f * eventNum) {
    std::cout << " read UserEventWeight: " << evtWeight << " - expected : " << 100.f * eventNum)
```
A bit less noisy with the same effect

tests/root_io/read_and_write_associated.cpp
```diff
@@ -24,7 +24,7 @@ void writeCollection() {
     item1.Number(i);
     info.push_back(item1);
 
-    event.putParameter("UserEventWeight", (float)100. * i);
+    event.putParameter("UserEventWeight", static_cast<float>(100. * i));
```
tmadlener, line 27:
```suggestion
    event.putParameter("UserEventWeight", 100.f * i);
```

tmadlener (commented):
Looks like there were quite a few. Thanks for cleaning up. Minor comment for the ones where we could also use a `.f` possibly.

tmadlener (approved):
Thanks.

jmcarcell:
For casts there is `-Wold-style-cast`. I added a pragma for Murmurhash3 so that it doesn't complain.

m-fila:
Are these flags and pragmas supported by all the compilers we have?

jmcarcell:
With all the compilers we have being GCC yes, but they also work fine with Clang.

PR #670, opened 2024-09-10
Strip empty lines from ExtraCode includes to avoid some warnings during code generation

tmadlener:
BEGINRELEASENOTES
- Strip strings to avoid having empty includes from ExtraCode that would otherwise generate a warning during code generation.
- Remove a debug printout from code generation

ENDRELEASENOTES

"Proper" fix for https://github.com/key4hep/EDM4hep/pull/352


jmcarcell (approved):
Independently of this, some consistency with newlines (`\n`) in EDM4hep would be nice.

PR #674, opened 2024-09-11
make `CreateDataFrame` accessible from python

### LINKED ISSUE #672: Make podio.CreateDataFrame() available
Reading a file as a dataframe with the following code

```
import podio

podio.CreateDataFrame("input-file.root")
```
is not yet available as of today. 

Tested with latest `key4hep` nightly (`-r 2024-09-11`).

m-fila:
BEGINRELEASENOTES
- Added `podio.data_source.CreateDataFrame` for creating RDataFrames with DataSource in python.

ENDRELEASENOTES

`CreateDataFrame` should be now accessible from python. It was put in a submodule `data_source` that will be accessible only when corresponding files are found (like `sio_io` submodule)

Closes #672


python/podio/data_source.py
```diff
@@ -0,0 +1,13 @@
+"""Python module for creating ROOT RDataFrame with files containing podio Frames"""
+
+from ROOT import gSystem, gInterpreter
+
+if not gSystem.DynamicPathName("libpodioDataSource", True):
+    raise ImportError("Error finding libpodioDataSource")
+
+if gInterpreter.LoadFile("podio/DataSource.h") != 0:
+    raise ImportError("Error when loading file podio/DataSource.h")
```
tmadlener, line 9:
```suggestion
if gSystem.Load("libpodioDataSource") < 0:
    raise ImportError("Cannot load the podio DataSource Library")
```
Is this not enough to make the function available? I.e. do we need to load the header file?

m-fila: That's what I tried originally but that doesn't seem to bring `CreateDataFrame` under `ROOT.podio`
```python
>>> import ROOT
>>> ROOT.gSystem.Load("libpodioDataSource")
0
>>> ROOT.podio.CreateDataFrame
AttributeError: <namespace cppyy.gbl.podio at 0x6159120> has no attribute 'CreateDataFrame'. Full details:
  type object 'podio' has no attribute 'CreateDataFrame'
  'podio::CreateDataFrame' is not a known C++ class
  'CreateDataFrame' is not a known C++ template
  'CreateDataFrame' is not a known C++ enum
>>> dir(ROOT.podio)
['CollectionBase', 'CollectionIDTable', 'DataSource', 'GenericParameter', 'GenericParameters', 'ObjectID', 'ROOTLegacyReader', 'ROOTReader', 'ROOTWriter', '__add__', '__bool__', '__class__', '__delattr__', '__destruct__', '__dict__', '__dir__', '__dispatch__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__getitem__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__invert__', '__le__', '__lt__', '__module__', '__mul__', '__ne__', '__neg__', '__new__', '__pos__', '__python_owns__', '__radd__', '__reduce__', '__reduce_ex__', '__repr__', '__reshape__', '__rmul__', '__rsub__', '__rtruediv__', '__setattr__', '__sizeof__', '__smartptr__', '__str__', '__sub__', '__subclasshook__', '__truediv__', '__weakref__', 'versio']
>>> 'DataSource' in dir(ROOT.podio)
True
>>> 'CreateDataFrame' in dir(ROOT.podio)
False
```

tmadlener: OK. I thought that would be the case, just wanted to make sure. I suppose it's also not possible to simply add a function declaration to `src/rds_selection.xml` in some way? I haven't quickly found a way on how to declare functions there. I am fine with loading the header here in fact, but I also think we should make sure to load the library not just detecting it, so a `gSystem.Load("...")` should be there in any case.

src/rds_selection.xml
```diff
@@ -1,5 +1,6 @@
 <lcgdict>
   <selection>
     <class name="podio::DataSource"/>
+    <function name="podio::CreateDataFrame"/>
```
tmadlener, line 4: This is good to know :)

tests/root_io/read_datasource.py
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env python3
+"""Small test case for checking DataSource based creating RDataFrames is accessible from python"""
+
+import ROOT
+from podio.data_source import CreateDataFrame
+
+if ROOT.gSystem.Load("libTestDataModelDict") < 0:  # noqa: E402
+    raise RuntimeError("Could not load TestDataModel dictionary")
+
+input_file = "example_frame.root"  # pylint: disable-msg=C0103
```
tmadlener, line 10: What is `C0103`? Can we avoid the pylint directive if we simply do this?
```python
rdf = CreateDataFrame("example_frame.root")
```

m-fila: invalid name for a constant. yeah that will work

tmadlener (approved):
Thanks.

PR #675, opened 2024-09-16
Fix a few wrong @params

jmcarcell:
BEGINRELEASENOTES
- Fix a few wrong @params in the documentation of some headers

ENDRELEASENOTES


jmcarcell:
I don't have any more. I had a quick look at other header files but didn't find anything.

PR #676, opened 2024-09-17
Add an static_assert for the GenericParameters and remove the EnableIf

jmcarcell:
This was discussed in https://github.com/key4hep/k4FWCore/pull/223#issuecomment-2354758322.

BEGINRELEASENOTES
- Remove EnableIf for GenericParameters and add `static_assert` in its place to make errors easier to read and debug.

ENDRELEASENOTES

### DIFF
--- a/include/podio/Frame.h
+++ b/include/podio/Frame.h
@@ -236,7 +236,7 @@ public:
   /// @param value The value of the parameter. A copy will be put into the Frame
   template <typename T>
   inline void putParameter(const std::string& key, T value) {
-    static_assert(podio::isSupportedGenericDataType<T>, "The type is not supported by the GenericParameters");
+    static_assert(podio::isSupportedGenericDataType<T>, "T is not a supported parameter type");
     m_self->parameters().set(key, std::move(value));
   }
 
@@ -285,7 +285,7 @@ public:
   /// @returns   An optional holding the value if it is present
   template <typename T>
   inline auto getParameter(const std::string& key) const {
-    static_assert(podio::isSupportedGenericDataType<T>, "The type is not supported by the GenericParameters");
+    static_assert(podio::isSupportedGenericDataType<T>, "T is not a supported parameter type");
     return m_self->parameters().get<T>(key);
   }
 
@@ -306,7 +306,7 @@ public:
   /// @returns  A vector of keys for this parameter type
   template <typename T>
   inline std::vector<std::string> getParameterKeys() const {
-    static_assert(podio::isSupportedGenericDataType<T>, "The type is not supported by the GenericParameters");
+    static_assert(podio::isSupportedGenericDataType<T>, "T is not a supported parameter type");
     return m_self->parameters().getKeys<T>();
   }
 
--- a/include/podio/Frame.h
+++ b/include/podio/Frame.h
@@ -304,8 +304,9 @@ public:
   /// @tparam T The desired parameter type
   ///
   /// @returns  A vector of keys for this parameter type
-  template <typename T, typename = podio::EnableIfValidGenericDataType<T>>
+  template <typename T>
   inline std::vector<std::string> getParameterKeys() const {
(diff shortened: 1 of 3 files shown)


tmadlener (commented):
I think the overload with `std::vector<std::string>` also needs some changes, because without SFINAE, `putParameter("somekey", {"an", "initializer_list", "of", "string", "literals"), will hit the one with `std::initializer_list<T>`, which will trigger a compilation error once it gets to the underlying `GenericParameters`.

tmadlener:
Also similar changes should be put in place for `getParameter`. However, there it should be much more straight forward.

tmadlener:
CI failure seems to be due to an issue in EDM4hep.

tmadlener:
CI for edm4hep workflows should start to work again with #677

python/podio/frame.py
```diff
@@ -264,7 +264,7 @@ def put_parameter(self, key, value, as_type=None):
                 cpp_types = _get_cpp_types(as_type)
                 if len(cpp_types) == 0:
                     raise ValueError(f"Cannot put a parameter of type {as_type} into a Frame")
-                self._frame.putParameter[cpp_types[0]](key, value)
+                self._frame.putParameter[cpp_types[0][0]](key, value)
```
tmadlener, line 267: Why is this change necessary? Was this a pre-existing bug?

tmadlener (commented):
I found another one at `getParameterKeys`. And the `set` with the `initializer_list` could also be changed to a `static_assert`, I think.

jmcarcell:
I think I got all of them now.

> I found another one at `getParameterKeys`. And the `set` with the `initializer_list` could also be changed to a `static_assert`, I think.

This one actually has to remain, because otherwise the `initializer_list<const char*>` will give an error since the elements won't be converted to `std::string`.

python/podio/frame.py
jmcarcell: `cpp_types` will be a list of tuples so the first `[0]` gets the first tuple. Before since there were two template arguments (one of them being the EnableIf) it was possible to pass this tuple, now it is not, and the second `[0]` gets the actual C++ type. So not exactly a bug. While having a look at this I also tried to simplify this part a bit but in the end I left it as it is with only a couple of lines in a more pythonic way.

include/podio/Frame.h
```diff
@@ -234,8 +234,9 @@ class Frame {
   ///              is supported by GenericParameters
   /// @param key   The name under which this parameter should be stored
   /// @param value The value of the parameter. A copy will be put into the Frame
-  template <typename T, typename = podio::EnableIfValidGenericDataType<T>>
+  template <typename T>
   inline void putParameter(const std::string& key, T value) {
+    static_assert(podio::isSupportedGenericDataType<T>, "The type is not supported by the GenericParameters");
```
tmadlener, line 239:
```suggestion
    static_assert(podio::isSupportedGenericDataType<T>, "T is not a supported parameter type");
```
I think `GenericParameters` is too much of an implementation detail here, especially if users get this message from their algorithms, they will not know what it is and where to look. I would change this message at least for the cases in `Frame`. Not sure if we should also put the supported types in, because this message is likely to become outdated if we have to add more types.

jmcarcell: For the same reason `T` doesn't tell you much where to look or what to do. But showing the type isn't completely trivial :shrug:

tmadlener: We need to have documentation where we state the supported datatypes (not sure if we already have). I think we could also drop the `T` from the message. In the end the compiler error will contain the type they tried to use in some form.

tmadlener (approved):

include/podio/Frame.h
m-fila: The docs mention int, float, strong and vectors of those https://github.com/AIDASoft/podio/blob/master/doc/frame.md#usage-for-parameters

tmadlener: Ah right, I knew it was somewhere :D. But it's missing a `double` :see_no_evil:

PR #679, opened 2024-09-18
fix printout on import when not built with data source

m-fila:
BEGINRELEASENOTES
- Fixed missing libpodioDataSourceDict printout on import podio when built without data source

ENDRELEASENOTES

Missed from #674. When library is not found root will make a long printout. This is fixed by first checking if the file exists which can be actually silenced to just return value

### DIFF
--- a/python/podio/data_source.py
+++ b/python/podio/data_source.py
@@ -2,7 +2,10 @@
 
 from ROOT import gSystem
 
-if gSystem.Load("libpodioDataSourceDict") < 0:
+if (
+    not gSystem.DynamicPathName("libpodioDataSourceDict", True)
+    or gSystem.Load("libpodioDataSourceDict") < 0
+):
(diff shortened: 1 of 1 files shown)


tmadlener (approved):

jmcarcell:
But the printout is useful because it shows you where ROOT looked so even if you are inexperienced it's possible to infer that `LD_LIBRARY_PATH` is contained there in case there is some issue with that, now it's an error that doesn't tell you anything.

tmadlener:
We are a bit all over the place at the moment here. In some modules we let it fail with printouts, in others we check first like in this PR. Sometimes we make it an `ImportError`, sometimes we don't (which will probably lead to runtime errors later). I think we should decide what we actually want to do here and then make all of it consistent.

m-fila:
Here the library looked for is optional and building is enabled with a cmake flag. I wanted to do the same thing as is done with bindings for `sio_io`

PR #681, opened 2024-09-19
Add a tool to merge several podio files into a single one

jmcarcell:
BEGINRELEASENOTES
- Add a tool to merge several podio files into a single one:
  - Metadata for every event can be saved or not
  - The same format as the first input file will be used (TTree or RNTuple)
  - Metadata about the input file names will be saved

ENDRELEASENOTES

Useful not to have to deal with many small files, even though the readers can read them fine.

### DIFF
--- a/tools/podio-merge-files
+++ b/tools/podio-merge-files
@@ -2,6 +2,8 @@
 """podio-merge-files tool to merge any number of podio files into one"""
 
 import argparse
+import sys
+import podio
 import podio.root_io
 from podio import reading
 
@@ -41,13 +43,29 @@ elif root_format == reading.RootFileFormat.RNTUPLE:
 else:
     raise ValueError(f"Input file {args.files[0]} is not a TTree or RNTuple file")
 
+categories = list(reader.categories)
+metadata_available = True
+try:
+    categories.remove(args.metadata_category_name)
+except ValueError:
+    metadata_available = False
 
 for category in reader.categories:
-    if category == args.metadata_category_name and args.metadata == "none":
-        continue
-    if category == args.metadata_category_name and args.metadata == "first":
-        all_frames = [reader.get(category)[0]]
-    else:
-        all_frames = reader.get(category)
+    all_frames = reader.get(category)
     for frame in all_frames:
         writer.write_frame(frame, category)
+
+if args.metadata == "none":
+    sys.exit(0)
+
+if not metadata_available:
+    print(f"Warning: metadata category '{args.metadata_category_name}' not found in input files, it will be created")
+    all_frames = [podio.Frame()]
+else:
+    if args.metadata == "first":
+        all_frames = [reader.get(args.metadata_category_name)[0]]
+    else:
+        all_frames = reader.get(args.metadata_category_name)
+for frame in all_frames:
+    frame.put_parameter("MergeInputFiles", args.files)
+    writer.write_frame(frame, args.metadata_category_name)
--- a/tools/podio-merge-files
+++ b/tools/podio-merge-files
@@ -17,9 +17,11 @@ parser.add_argument(
     default="first",
     help="metadata to include in the output file, default: only the one from the first event, other options: all events, none",
 )
-parser.add_argument("--metadata-category-name",
-                    default="metadata",
-                    help="name of the metadata category in the output file, default: metadata")
+parser.add_argument(
+    "--metadata-category-name",
+    default="metadata",
+    help="name of the metadata category in the output file, default: metadata",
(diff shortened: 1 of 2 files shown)


tools/podio-merge-files
```diff
@@ -0,0 +1,52 @@
...
+    description="Merge any number of podio files into one, can merge TTree and RNTuple files"
+)
+
+parser.add_argument("--output-file", help="name of the output file", required=True)
+parser.add_argument("files", nargs="+", help="which files to merge")
+parser.add_argument(
+    "--metadata",
+    choices=["none", "all", "first"],
+    default="first",
+    help="metadata to include in the output file, default: only the one from the first event, other options: all events, none",
```
tmadlener, line 18:
```suggestion
    help="metadata to include in the output file, default: only the one from the first file, other options: all files, none",
```
I assume it should be files not events?

tmadlener (commented):
This does a lot of heavy lifting that might not be strictly necessary, as it unpacks and than repacks every collection in every frame, whereas a simple `hadd` would almost be enough for merging `TTree` based files. Additionally, it will do schema evolution (if applicable), so that there could be subtle changes to the produced output file. On the other hand, this will correctly handle input files with different schema versions, so that might be not too bad.

How much slower is this than a c++ implementation? There is #620 after all to solve a quite significant performance issue with `podio-dump`, and the difference is not only the long startup time in that case.

jmcarcell:
Ah I didn't think about `hadd`, this is a tool I have used a couple of times (well only the for loop) so I thought it would be nice to have. On small files (1 MB TTree files) it's much faster than in python, but that's the case with the worst relative overhead I would imagine. For RNTuples it crashes with the ones I tried.

tmadlener:
>  For RNTuples it crashes with the ones I tried.

Can you specify what crashes in this case? `hadd` or the c++ version of this script?

jmcarcell:
`hadd` is crashing

tools/podio-merge-files
jmcarcell: Yes

tmadlener:
Maybe add a bit of metadata that keeps track of which files the the merged files comes from.

jmcarcell:
This should be ready and now write some metadata with the names of the files as passed in the arguments (full paths being saved as full paths):
```
Key                           Value
--------------------------------------------------------------------------------
MergeInputFiles               [example_frame.root, example_frame2.root]
```

tools/podio-merge-files
```diff
@@ -0,0 +1,80 @@
...
+parser.add_argument(
+    "--metadata-category-name",
+    default="metadata",
+    help="name of the metadata category in the output file, default: metadata",
+)
+parser.add_argument(
+    "--metadata-parameter-name",
+    default="MergeInputFiles",
+    help="name of the metadata category in the output file, default: metadata",
+)
```
tmadlener, line 32: Does this have to be configurable?

jmcarcell: I don't know, we could just hardcode it having that string hardcoded there felt out of place when I had another look :shrug:

tools/podio-merge-files
```diff
@@ -0,0 +1,80 @@
...
+    "only the one from the first file, other options: all files, none",
+)
+parser.add_argument(
+    "--metadata-category-name",
+    default="metadata",
+    help="name of the metadata category in the output file, default: metadata",
+)
+parser.add_argument(
+    "--metadata-parameter-name",
+    default="MergeInputFiles",
```
tmadlener, line 30:
```suggestion
    default="MergedInputFiles",
```

tools/podio-merge-files
tmadlener: As discussed during the meeting just now. We will hardcode this and the `metadata-category-name` instead of making them configurable.

jmcarcell: Now they are hardcoded

tools/podio-merge-files
```diff
@@ -0,0 +1,70 @@
...
+    help="metadata to include in the output file, default: "
+    "only the one from the first file, other options: all files, none",
+)
+args = parser.parse_args()
+
+all_files = set()
+for f in args.files:
+    if f in all_files:
+        raise ValueError(f"File {f} is present more than once in the input list")
+    all_files.add(f)
```
andresailer, line 29: Maybe would be better to keep the order of the files as given in the argument list? `set()` is not insertion ordered

andresailer: Ahh, you don't use all_files later on.

PR #682, opened 2024-09-19
DataSource: implement missing GetLabel method

Zehvogel:
BEGINRELEASENOTES
- DataSource: implement missing GetLabel method

ENDRELEASENOTES

also updated a comment in the generic parameters

### DIFF
--- a/include/podio/DataSource.h
+++ b/include/podio/DataSource.h
@@ -85,7 +85,7 @@ public:
   ///
   std::string GetTypeName(std::string_view columnName) const override;
 
-  std::string GetLabel() override { 
+  std::string GetLabel() override {
     return "PODIO Datasource";
   };
 
--- a/include/podio/DataSource.h
+++ b/include/podio/DataSource.h
@@ -85,7 +85,9 @@ public:
   ///
   std::string GetTypeName(std::string_view columnName) const override;
 
-  std::string GetLabel() override { return "PODIO Datasource"; };
+  std::string GetLabel() override { 
+    return "PODIO Datasource";
+  };
 
 protected:
   ///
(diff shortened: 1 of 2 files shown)


src/GenericParameters.cc
```diff
@@ -17,7 +17,7 @@ GenericParameters::GenericParameters(const GenericParameters& other) :
     m_stringMtx(std::make_unique<std::mutex>()),
     m_doubleMtx(std::make_unique<std::mutex>()) {
   {
-    // acquire all three locks at once to make sure all three internal maps are
+    // acquire all four locks at once to make sure all four internal maps are
```
jmcarcell, line 20:
```suggestion
    // acquire all locks at once to make sure all internal maps are
```
What about this because it's likely if the number is changed the comments will not be changed?

Zehvogel: I would also be fine with that yes :)

tmadlener (approved):

include/podio/DataSource.h
```diff
@@ -85,6 +85,8 @@ class DataSource : public ROOT::RDF::RDataSource {
   ///
   std::string GetTypeName(std::string_view columnName) const override;
 
+  std::string GetLabel() override { return "PODIO Datasource"; };
```
tmadlener, line 88:
```suggestion
  std::string GetLabel() override { 
    return "PODIO Datasource";
  };
```

include/podio/DataSource.h
```diff
@@ -85,6 +85,10 @@ class DataSource : public ROOT::RDF::RDataSource {
   ///
   std::string GetTypeName(std::string_view columnName) const override;
 
+  std::string GetLabel() override {
```
tmadlener, line 88:
```suggestion
  std::string GetLabel() override {
```

PR #683, opened 2024-09-20
use enable_if to allow using containers of interfaces with object wrappers

m-fila:
BEGINRELEASENOTES
- Using enable if instead of static_assert in interface type construct to avoid errors with direct initialization of contianers of interface types

ENDRELEASENOTES

key4hep/k4FWCore#238 I spotted that the containers of interface types can't be used in stock `Gaudi::Functional` algorithms (unlike containers of data types) due to instantiating wrong constructor from the template.
The problem was vector's direct initialization with `{std::vector<interface_type>&&}`  would call the interface type constructor with `std::vector<interface_type>&&` (failing late due to the assert) rather than vector's constructor. With `enable_if` the correct overload is chosen due to SFINAE

### DIFF
--- a/python/templates/Interface.h.jinja2
+++ b/python/templates/Interface.h.jinja2
@@ -92,7 +92,8 @@ private:
  std::unique_ptr<Concept> m_self{nullptr};
 
 public:
-  template<typename ValueT, typename = std::enable_if_t<isInitializableFrom<ValueT>>> // "{{ class.bare_type }} can only be initialized with one of the following types (and their Mutable counter parts): {{ Types | join(", ") }}");
+  // {{ class.bare_type }} can only be initialized with one of the following types (and their Mutable counter parts): {{ Types | join(", ") }}
+  template<typename ValueT, typename = std::enable_if_t<isInitializableFrom<ValueT>>>
   {{ class.bare_type }}(ValueT value) :
     m_self(std::make_unique<Model<podio::detail::GetDefaultHandleType<ValueT>>>(value)) {
   }


tests/unittests/interface_types.cpp
```diff
@@ -62,6 +64,10 @@ TEST_CASE("InterfaceTypes STL usage", "[interface-types][basics]") {
   REQUIRE(counterMap[empty] == 1);
   REQUIRE(counterMap[hit] == 2);
   REQUIRE(counterMap[wrapper] == 2);
+
+  // check container of interfaces move constructor and direct initialization
+  auto interfaces = std::vector<TypeWithEnergy>{};
+  [[maybe_unused]] auto interfaces2 = std::vector<TypeWithEnergy>{std::move(interfaces)};
```
tmadlener, line 70: Can we add an element or two to this vector and check that it moves correctly?

python/templates/Interface.h.jinja2
```diff
@@ -89,10 +89,9 @@ class {{ class.bare_type }} {
  std::unique_ptr<Concept> m_self{nullptr};
 
 public:
-  template<typename ValueT>
+  template<typename ValueT, typename = std::enable_if_t<isInitializableFrom<ValueT>>>
   {{ class.bare_type }}(ValueT value) :
     m_self(std::make_unique<Model<podio::detail::GetDefaultHandleType<ValueT>>>(value)) {
-    static_assert(isInitializableFrom<ValueT>, "{{ class.bare_type }} can only be initialized with one of the following types (and their Mutable counter parts): {{ Types | join(", ") }}");
```
tmadlener, line 95: Wondering whether we should leave this in, but I don't think it's visible in the compiler error message in any case, right?

m-fila: I'm not sure if there is any scenario in which that would get printed- even if some would try to explicitly instantiate wrong overload the compiler will print some substitution error in the `enable_if`.

tmadlener: Yeah, they are rather terse with context around where things happen. It would only help if people look into the header file then and see the message to the `static_assert`. In principle `isInitializableFrom<ValueT>` is rather obvious I think, but you have to find it (and be able to interpret it), in all the compiler output.

m-fila: Hmm, the comment could be moved to line with enable_if so it appears somewhere in the compilation error
```
/cvmfs/sw.hsf.org/contrib/x86_64-almalinux9-gcc11.4.1-opt/gcc/14.2.0-yuyjov/lib/gcc/x86_64-pc-linux-gnu/14.2.0/../../../../include/c++/14.2.0/type_traits: In substitution of 'template<bool _Cond, class _Tp> using std::enable_if_t = typename std::enable_if::type [with bool _Cond = false; _Tp = void]':
/home/mafila/podio/tests/datamodel/TypeWithEnergy.h:96:29:   required from here
   96 |   template<typename ValueT, typename = std::enable_if_t<isInitializableFrom<ValueT>>> // TypeWithEnergy can only be initialized with one of the following types (and their Mutable counter parts): ExampleHit, ExampleMC, ExampleCluster");
      |                             ^~~~~~~~
/cvmfs/sw.hsf.org/contrib/x86_64-almalinux9-gcc11.4.1-opt/gcc/14.2.0-yuyjov/lib/gcc/x86_64-pc-linux-gnu/14.2.0/../../../../include/c++/14.2.0/type_traits:2696:11: error: no type named 'type' in 'struct std::enable_if<false, void>'
 2696 |     using enable_if_t = typename enable_if<_Cond, _Tp>::type;
```
Would it be helpful?

tmadlener: It's definitely better than nothing, and it will appear pretty much at the top somewhere, right?

tmadlener: I think, it's pretty much the best we can do at this point without concepts.

tmadlener (approved):

python/templates/Interface.h.jinja2
```diff
@@ -92,10 +92,9 @@ private:
  std::unique_ptr<Concept> m_self{nullptr};
 
 public:
-  template<typename ValueT>
+  template<typename ValueT, typename = std::enable_if_t<isInitializableFrom<ValueT>>> // "{{ class.bare_type }} can only be initialized with one of the following types (and their Mutable counter parts): {{ Types | join(", ") }}");
```
jmcarcell, line 95: Since the comment is quite long, can it be moved to another line?

m-fila: If we move it then it won't be displayed in the compilation error. I put some example in the other suggestion

tmadlener: I just realized that anyone who has `clang-format` will probably not see the comment in this form in any case, because that will just split it into several lines, and the error will only contain the first one.

m-fila: Oh, right. I think I'd rather give up and put in another line then

tmadlener:
EDM4hep CI failures should be fixed with #690

tmadlener:
I am going to merge this as soon as CI passes. Just to be sure: This one is not immediately needed in a tag, right? Because the discussion that triggered this is still open / not finally settled in k4FWCore either.

m-fila:
Yes, I think it should be fine to add this later

PR #687, opened 2024-09-23
Make sure test environment points to the correct place

tmadlener:
BEGINRELEASENOTES
- Make sure the `ROOT_INCLUDE_PATH` in the test environment points to the correct place.

ENDRELEASENOTES

Should point into the source tree not the build tree (where the pointed to place does not exist).


hegner (approved):
If tests run through, I'll merge. thanks!

tmadlener:
The key4hep based one will likely fail: https://github.com/key4hep/key4hep-spack/issues/651

hegner:
The pre-commit looks very strange, but unrelated to this change. Key4hep worked though

PR #689, opened 2024-09-25
Use unique_ptr to manage one-to-one-relations in the Obj classes

tmadlener:
BEGINRELEASENOTES
- Use a `unique_ptr` to manage the OneToOneRelation pointers in the `Obj` classes.

ENDRELEASENOTES

I am not entirely sure why they were not managed by smart pointers before. It's possible that this was not possible before we introduced the `MaybeSharedPtr`s.


hegner:
@tmadlener - ROOT got hickups previously with them. But that's plenty of years ago. Thanks for the improvement!

PR #692, opened 2024-09-27
GenericParameters.h: add missing clear and empty check for double map

Zehvogel:
just saw this while scrolling through. I have not checked if it is also missing in more places

BEGINRELEASENOTES
- GenericParameters.h: added missing clear and empty check for double map

ENDRELEASENOTES


m-fila:
It doesn't seem to be missing anywhere else.

tmadlener (approved):
Thanks. Seems like I missed this when we added the `double`s.

PR #693, opened 2024-10-04
Make accessing single elements in empty parameter vectors return an empty optional

tmadlener:
BEGINRELEASENOTES
- Fix small issue in `GenericParameters` where trying to access a single element of a parameter that was stored as an empty vector resulted in a crash. Instead make this return an empty optional now.

ENDRELEASENOTES

@dirkzerwas this is the fix for the issue in our email discussion.

This seems to be the most reasonable thing to do. The other option would be to return an empty / default constructed value in this case. However, since we introduced the `std::optional` return type for avoiding exactly that (see #580), I don't think that makes a lot of sense and would make the API inconsistent again.

### DIFF
--- a/include/podio/GenericParameters.h
+++ b/include/podio/GenericParameters.h
@@ -219,6 +219,9 @@ std::optional<T> GenericParameters::get(const std::string& key) const {
     return it->second;
   } else {
     const auto& iv = it->second;
+    if (iv.empty()) {
+      return std::nullopt;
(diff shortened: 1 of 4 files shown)


tmadlener:
As discussed during [the meeting on Nov 5](https://indico.cern.ch/event/1465606/), we will be using an empty optional (as is the current state of this PR at the moment).

PR #694, opened 2024-10-09
Remove redundant inline

jmcarcell:
BEGINRELEASENOTES
- Remove redundant inline, `constexpr` implies `inline`
ENDRELEASENOTES

### DIFF
--- a/tests/unittests/std_interoperability.cpp
+++ b/tests/unittests/std_interoperability.cpp
@@ -30,7 +30,7 @@ struct has_value_type : std::false_type {};
 template <typename T>
 struct has_value_type<T, std::void_t<typename T::value_type>> : std::true_type {};
 template <typename T>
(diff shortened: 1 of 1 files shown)


m-fila:
> A constexpr specifier used in an object declaration or non-static member function(until C++14) implies const. A constexpr specifier used in a function or [static](https://en.cppreference.com/w/cpp/language/static) data member(since C++17) declaration implies inline.

Those are variable declarations so `inline` isn't implied

tmadlener:
I agree with @m-fila here. On the other hand it shouldn't matter too much in this case since we are not relying on anything that would require them to be `inline` (i.e. we are not explicitly linking against any of this code, nor are we expecting the variables to have the same address in all translation units).

jmcarcell:
Yeah `inline` is not doing anything here, the same object code is produced with and without so it may as well be removed. I'll change the title.

PR #697, opened 2024-10-17
[temp-fix] Set JULIA_DEPOT_PATH to a writable directory

tmadlener:
BEGINRELEASENOTES
- Temporarily set the `JULIA_DEPOT_PATH` to a writable directory to make CI pass.

ENDRELEASENOTES

At least until LCG stacks have been fixed.


tmadlener:
Not our issue to fix, I think. Opened https://its.cern.ch/jira/browse/SPI-2666 for an upstream fix in LCG stacks

andresailer:
Might still be worth changing this here until there is a fix in the stacks?

PR #699, opened 2024-10-23
add compatibility with pathlib.Path to root_io

Victor-Schwan:
BEGINRELEASENOTES
- add compatibility with pathlib.Path objects in root_io and sio_io

ENDRELEASENOTES

### DIFF
--- a/python/podio/root_io.py
+++ b/python/podio/root_io.py
@@ -8,12 +8,8 @@ from utils import convert_to_str_paths
 gSystem.Load("libpodioRootIO")  # noqa: E402
 from ROOT import podio  # noqa: E402 # pylint: disable=wrong-import-position
 
-from podio.base_reader import (  # pylint: disable=wrong-import-position # noqa: E402
-    BaseReaderMixin,
-)
-from podio.base_writer import (  # pylint: disable=wrong-import-position # noqa: E402
-    BaseWriterMixin,
-)
+from podio.base_reader import BaseReaderMixin  # pylint: disable=wrong-import-position # noqa: E402
+from podio.base_writer import BaseWriterMixin  # pylint: disable=wrong-import-position # noqa: E402
 
 
 class Reader(BaseReaderMixin):
--- a/python/podio/root_io.py
+++ b/python/podio/root_io.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 """Python module for reading root files containing podio Frames"""
 
+from pathlib import Path
 from ROOT import gSystem
 
 gSystem.Load("libpodioRootIO")  # noqa: E402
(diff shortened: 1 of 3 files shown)


tmadlener (commented):
I suppose you stumbled over this when trying to instantiate a `Reader` with a `pathlib.Path` object and it failed?

The same issue will be present in `sio_io`, so it should also be fixed there.

python/podio/root_io.py
```diff
@@ -10,18 +11,39 @@
...
+    if isinstance(filenames, list):
+        str_filenames = []
+        for f in filenames:
+            if isinstance(f, (str, Path)):
+                str_filenames.append(str(f))
+            else:
+                raise TypeError(f"Invalid filename type: {f} (type: {type(f)})")
+        return str_filenames
+
+    raise TypeError(f"Invalid filenames argument: {filenames} (type: {type(filenames)})")
```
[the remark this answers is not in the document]
Victor-Schwan: Ok. Do you have any objections to using `os.fspath` instead of `str`? It is better for catching wrong types, e.g. if `None` is passed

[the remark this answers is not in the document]
tmadlener: If `reader.openFiles` swallows a list of `os.fspath`, I have no objections.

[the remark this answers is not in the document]
tmadlener: If `reader.openFiles` swallows a list of `os.fspath`, I have no objections.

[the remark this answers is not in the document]
Victor-Schwan: `os.fspath()` only returns `str` or `bytes`. I am unsure what happens if a `bytes` path is given to `reader.openFiles`. `fspath` is just better at throwing `TypeError` if the argument is not `pathlike` compared to `str()`. https://docs.python.org/3/library/os.html#os.fspath

python/podio/sio_io.py
```diff
@@ -20,8 +21,9 @@ def __init__(self, filename):
         """Create a reader that reads from the passed file.
 
         Args:
-            filename (str): File to open and read data from
+            filename (str or Path): File to open and read data from.
         """
+        filename = convert_to_str_paths(filename)
```
tmadlener, line 26:
```suggestion
        filename = convert_to_str_paths(filename)[0]
```
As `openFile` takes a single string.

python/podio/sio_io.py
```diff
@@ -49,13 +52,15 @@ def __init__(self, filename):
...
 
     def __init__(self, filename):
-        """Create a writer for writing files
+        """Create a writer for writing files.
 
         Args:
-            filename (str): The name of the output file
+            filename (str or Path): The name of the output file.
         """
+        filename = convert_to_str_paths(filename)
```
tmadlener, line 63:
```suggestion
        filename = convert_to_str_paths(filename)[0]
```

python/podio/root_io.py
```diff
@@ -10,18 +13,34 @@
...
+    Returns:
+        list[str]: A list of filepaths as strings.
+    """
+
+    if isinstance(filenames, Iterable) and not isinstance(filenames, (str, Path)):
+        return [os.fspath(fn) for fn in filenames]
+
+    return [os.fspath(filenames)]
+
+
```
tmadlener, line 33: I would move this into a separate module that can then be imported by `root_io` and `sio_io` instead of cross-importing this from `root_io` to `sio_io`. Altough I generally don't like `utils` too much, I can't think of a better name at the moment, so I would be OK with that.

python/podio/root_io.py
```diff
@@ -3,11 +3,17 @@
...
 from ROOT import podio  # noqa: E402 # pylint: disable=wrong-import-position
 
-from podio.base_reader import BaseReaderMixin  # pylint: disable=wrong-import-position # noqa: E402
-from podio.base_writer import BaseWriterMixin  # pylint: disable=wrong-import-position # noqa: E402
+from podio.base_reader import (  # pylint: disable=wrong-import-position # noqa: E402
+    BaseReaderMixin,
+)
+from podio.base_writer import (  # pylint: disable=wrong-import-position # noqa: E402
+    BaseWriterMixin,
+)
```
tmadlener, line 16:
```suggestion
from podio.base_reader import BaseReaderMixin  # pylint: disable=wrong-import-position # noqa: E402
from podio.base_writer import BaseWriterMixin  # pylint: disable=wrong-import-position # noqa: E402
```
No unrelated format changes please

python/podio/utils.py
```diff
@@ -0,0 +1,21 @@
+import os
```
tmadlener, line 1:
```suggestion
#!/usr/bin/env python3
"""utility functionality for podio"""

import os
```

PR #700, opened 2024-10-30
Add a deprecation warning when building with C++17

jmcarcell:
BEGINRELEASENOTES
- Add a deprecation warning when building with C++17 before removing support for it in https://github.com/AIDASoft/podio/pull/698

ENDRELEASENOTES

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,6 +40,10 @@ if(NOT CMAKE_CXX_STANDARD MATCHES "17|20")
   message(FATAL_ERROR "Unsupported C++ standard: ${CMAKE_CXX_STANDARD}")
 endif()
 
+if(CMAKE_CXX_STANDARD EQUAL 17)
+  message(WARNING "C++17 is deprecated and support for it will be removed in the future. Please use at least C++20.")
+endif()
+
 # Prevent CMake falls back to the latest standard the compiler does support
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -40,6 +40,10 @@ if(NOT CMAKE_CXX_STANDARD MATCHES "17|20")
   message(FATAL_ERROR "Unsupported C++ standard: ${CMAKE_CXX_STANDARD}")
 endif()
 
+if(CMAKE_CXX_STANDARD EQUAL 17)
+  message(WARNING "C++17 is deprecated and support for it will be removed in the future. Please use at least C++20.")
```
tmadlener, line 44: Should we put a version here when we are going to remove it? If we want to have a tag in between where this message should be visible that would make it v01-03.

jmcarcell: Added as discussed today in the podio and EDM4hep meeting.

andresailer:
Should we also set the default standard to 20 then, and not print the warning if the user doesn't do anything?

(cannot comment on the line directly)
https://github.com/AIDASoft/podio/pull/700/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR37

tmadlener:
> Should we also set the default standard to 20 then, and not print the warning if the user doesn't do anything?

Yes, definitely.

jmcarcell:
Added, I thought leaving all the changes to #698.

tmadlener:
The `dev3` based workflows are failing for several PRs and I think is unrelated to the changes in any of those. From what I can tell from just looking at the outputs, it looks like we no longer recognize whether a file has a TTree or an RNTuple inside and it then instantiates the wrong reader in the interface:

https://github.com/AIDASoft/podio/actions/runs/11681697887/job/32527272599?pr=700#step:4:855

PR #702, opened 2024-11-07
Make sure that reader still works with newer ROOT versions

tmadlener:
BEGINRELEASENOTES
- Make sure that `makeReader` still recognizes files with RNTuple inside after they have been marked non-experimental in ROOT.

ENDRELEASENOTES

This should fix the issues in CI we see in some other PRs that build against the root nightlies.


tmadlener:
Backwards compatibility tests involving RNTuples will probably be a source for some headaches until it's fully finalized in ROOT.

PR #704, opened 2024-11-08
Fix tests when SIO is OFF

jmcarcell:
BEGINRELEASENOTES
- Fix tests when SIO is OFF

ENDRELEASENOTES

In #257 `DIFF_EXTRA_ARGS` is being set above so when SIO is OFF it gets overwritten


tmadlener (approved):

tmadlener:
Good catch. Thanks.

PR #707, opened 2024-11-19
Run clang-tidy at the end of pre-commit

jmcarcell:
BEGINRELEASENOTES
- Run clang-tidy at the end of pre-commit since other failures are more likely and clang-tidy takes a long time.

ENDRELEASENOTES

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,12 +7,6 @@ repos:
         exclude: (doc/ReleaseNotes.md)
   - repo: local
     hooks:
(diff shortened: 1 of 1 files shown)


tmadlener:
Does it make a difference from a runtime point of view? I thought pre-commit runs all the checks regardless of the ones that fail. For looking at the logs while they are produced this makes sense, just curious about other effects.

jmcarcell:
No, runtime will be the same.

tmadlener (approved):

PR #708, opened 2024-11-19
Add support for reading several rntuple files

jmcarcell:
BEGINRELEASENOTES
- Add support for reading several RNtuple files

ENDRELEASENOTES

Now it will work correctly for multiple files. For TTrees reading multiple files with random access (needed because we can read any entry) is provided through `TChain` and `TBranch`. For `RNTuples` there is not such thing, there is a `RNTupleProcessor` but it is intended for iterative access and not random access:
https://root.cern.ch/doc/master/classROOT_1_1Experimental_1_1RNTupleProcessor.html

So the idea is to save a vector with how many entries each file has, and then every time an entry is read do a lookup in the vector to find out which file we are at and get the corresponding reader for that file. Scales as log N in the number of files, which is probably OK for most cases.

Requires C++20 for `std::ranges::upper_bound`.

### DIFF
--- a/include/podio/RNTupleReader.h
+++ b/include/podio/RNTupleReader.h
@@ -150,7 +150,8 @@ private:
   std::unordered_map<std::string, unsigned> m_entries{};
   // Map category to a vector that contains at how many entries each reader starts
   // For example, if we have 3 readers and the first one has 10 entries, the second one 20 and the third one 30
-  // then the vector will be {0, 10, 30, 60}
+  // then the vector will be {0, 10, 30}
+  // 60 is not needed because anything after 30 will be in the last reader
   std::unordered_map<std::string, std::vector<unsigned>> m_readerEntries{};
   std::unordered_map<std::string, unsigned> m_totalEntries{};
 
--- a/include/podio/RNTupleReader.h
+++ b/include/podio/RNTupleReader.h
@@ -151,7 +151,7 @@ private:
   // Map category to a vector that contains at how many entries each reader starts
   // For example, if we have 3 readers and the first one has 10 entries, the second one 20 and the third one 30
   // then the vector will be {0, 10, 30, 60}
-  std::unordered_map<std::string, std::vector<unsigned>> m_readerEntries;
+  std::unordered_map<std::string, std::vector<unsigned>> m_readerEntries{};
   std::unordered_map<std::string, unsigned> m_totalEntries{};
 
   struct CollectionInfo {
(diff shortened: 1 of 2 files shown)


include/podio/RNTupleReader.h
```diff
@@ -147,7 +147,11 @@ class RNTupleReader {
   std::unordered_map<std::string, std::unique_ptr<ROOT::Experimental::RNTupleReader>> m_metadata_readers{};
   std::vector<std::string> m_filenames{};
 
-  std::unordered_map<std::string, int> m_entries{};
+  std::unordered_map<std::string, unsigned> m_entries{};
+  // Map category to a vector that contains at how many entries each reader starts
+  // For example, if we have 3 readers and the first one has 10 entries, the second one 20 and the third one 30
+  // then the vector will be {0, 10, 30, 60}
```
tmadlener, line 153: I am not sure I fully understand why we need to store the final sum of entries here? Is it because of the `upper_bound` below?

jmcarcell: It doesn't have to be stored (and it isn't) but since it has to be computed anyway I think the code looks cleaner like it is (computing it and storing it in the vector and later removing it) than having to check if it's the last entry every time and not add it in that case. For the comment I can remove the last number

src/RNTupleReader.cc
```diff
@@ -143,15 +150,22 @@ std::unique_ptr<ROOTFrameData> RNTupleReader::readEntry(const std::string& categ
 
   m_entries[category] = entNum + 1;
 
+  // m_readerEntries contains the accumulated entries for all the readers
+  // therefore, the first number that is lower or equal to the entry number
+  // is at the index of the reader that contains the entry
+  auto upper = std::ranges::upper_bound(m_readerEntries[category], entNum);
+  auto localEntry = entNum - *(upper - 1);
+  auto readerIndex = upper - 1 - m_readerEntries[category].begin();
```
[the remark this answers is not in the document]
jmcarcell: No, it's not the same, `upper_bound` gives the first one that is greater and `lower_bound` gives the first that is equal or greater so in most cases the lookup will return the same number for both and therefore `lower_bound != upper_bound -1`.

tmadlener (commented):
Ubuntu CI is failing because the use of `std::ranges` effectively requires c++20 and gcc11 that is used there is not yet enough. That should resolve itself with #698 I think.

tmadlener:
Rebased this after merging #698. Merging after CI passes.

PR #691, opened 2024-09-26
Introduce a new `links` category for the YAML definitions

tmadlener:
BEGINRELEASENOTES
- Introduce a new `links` category into the YAML grammar to automate the declaration of Links.

ENDRELEASENOTES

- [x] Includes #257 
- [x] Documentation
- [x] Includes #710 

[Diff wrt to #257](https://github.com/tmadlener/podio/compare/associations...tmadlener:podio:links-in-yaml?expand=1)

As can be seen from the diff above, I needed to change the definition of the `PODIO_DECLARE_LINK` macro, and it now only deals with the registration to the `CollectionBufferFactory` and there is a separate macro to deal with the registration to the `SIOBlockFactory`. Otherwise, it becomes hard to only build the "central" part of a datamodel without any dependencies on an I/O backend. This could be lifted to #257 if we wanted.

PR #710, opened 2024-11-20
Make sure that link weights are stored in a struct

tmadlener:
Otherwise ROOT does something clever and simply stores the float without the struct. Which makes the switch over a bit less transparent.

BEGINRELEASENOTES
- Make sure that the weights of the links end up as a `vector<LinkData>` in the ROOT files instead of a `vector<float>`

ENDRELEASENOTES

Pulled this out of #691 as it is really an unrelated fix that has nothing to do with that.


tmadlener:
Merging this now and #691 on Monday next week, unless someone speaks up against it.

Issue #611, opened 2024-05-30
Use types of heterogeneous origins in interfaces


BrieucF:
- OS version: Alma9
 - Environment: source /cvmfs/sw-nightlies.hsf.org/key4hep/setup.sh -r 2024-05-30
 - Goal: add an interface which includes types of heterogeneous origins (edm4hep and an extension):

This:
```
interfaces:
  extension::TrackerHit:
    Description: "Tracker hit interface class"
    Author: "Thomas Madlener, DESY"
    Members:
      - uint64_t cellID // ID of the sensor that created this hit
      - int32_t type // type of the raw data hit
      - int32_t quality // quality bit flag of the hit
      - float time [ns] // time of the hit
      - float eDep [GeV] // energy deposited on the hit
      - float eDepError [GeV] // error measured on eDep
      - edm4hep::Vector3d position [mm] // hit position
    Types:
      - edm4hep::TrackerHitPlane
      - edm4hep::TrackerHit3D
      - extension::DriftChamberDigi
```

Crashes with 
```
/afs/cern.ch/user/b/brfranco/work/public/tracketHitInterface_drift_chamber/k4RecTracker/build/DCHdigi/src/TrackCollection.cc:12:10: fatal error: extension/TrackerHit3DCollection.h: No such file or directory
   12 | #include "extension/TrackerHit3DCollection.h"
```

There is a workaround which is to redefine those types in the extensions and use 
```
interfaces:
  extension::TrackerHit:
    Description: "Tracker hit interface class"
    Author: "Thomas Madlener, DESY"
    Members:
      - uint64_t cellID // ID of the sensor that created this hit
      - int32_t type // type of the raw data hit
      - int32_t quality // quality bit flag of the hit
      - float time [ns] // time of the hit
      - float eDep [GeV] // energy deposited on the hit
      - float eDepError [GeV] // error measured on eDep
      - edm4hep::Vector3d position [mm] // hit position
    Types:
      - extension::TrackerHitPlane
      - extension::TrackerHit3D
      - extension::DriftChamberDigi
```

tmadlener:
I can't reproduce the exact error you see (i.e. missing include). However, there is indeed a somewhat conceptual issue here. The main issue is that types that are used in interface declare that interface as `friend`, such that the interface can access its internals.

Since we can't change the headers of the upstream models, we will have to find another solution to this.

m-fila:
I think it would be another problem if a model and upstream used different `getSyntax`

[...]

PR #714, opened 2024-11-29
Allow types from different data models in interfaces

m-fila:
BEGINRELEASENOTES
- Added possibility to use in the interfaces the datatypes from different podio-based datamodels

ENDRELEASENOTES

The data types had to be aware of interfaces in which they were used in order to make them `friend`. For instance the types from base model couldn't be made aware of types in model extension so it was impossible to define a model extension with an interface that uses types from base model. #611

It seems the `friend` internal access was only used to get pointers to internal object in order to define `operator<` (for usage with maps) by comparing addresses.

In this PR I propose to remove the `friend` coupling of interfaces and data types. Instead a `podio::detail::getRank`free function is added as a friend to each data type. The functions returns unsigned numeric type that can be used for comparisons. According to the standard the most appropriate  type here - `uintptr_t`- is optional, if it's not present then `uintmax_t` will be used as it's the biggest unsigned type so a pointer will also fit in it.

I had hard time choosing a name for numeric value, that can be used for comparisons and I'm not entirely happy with calling it `rank`.

Another extension model is used as the extension we had had different `getSyntax` than the base model which the interfaces can't handle.

Renamed `Rank` to `OrderKey`


tmadlener:
I think this is a nice solution to the problem. We will potentially need to put some comments into the code as well to explain why we have this (or mention it in some of the documentation).

One potential issue I see with this approach is that we are technically leaking the `Obj*` addresses to the outside, right? For example, it would be possible to do the following:

```cpp
MutableHit hit;
auto rank = detail::getRank(hit);
auto hitObjPtr = reinterpret_cast<HitObj*>(rank);
// now we have access to the hitObjPtr!
```

I don't really see a way to avoid this though at the moment. We would need to have a way to have access to both ranks simultaneously without leaking the actual values to the outside.  In the end we probably have to weigh whether this potential leak is worth the additional feature.

What other names did you have on your short list instead of `rank`?

m-fila:
Thanks

> One potential issue I see with this approach is that we are technically leaking the `Obj*` addresses to the outside, right?

Yes, unfortunately this is possible. We could hide it in a proxy like this:
```c++
class Rank {
public:
  Rank(void* rank) : m_rank(reinterpret_cast<value_type>(rank)) {
  }
  friend bool operator<(const Rank& lhs, const Rank& rhs) {
    return lhs.m_rank < rhs.m_rank;
  }

private:
#ifdef UINTPTR_MAX
  using value_type = uintptr_t;
#else
  using value_type = uintmax_t;
#endif
  value_type m_rank;
};
```
> What other names did you have on your short list instead of `rank`?
- rankValue
- ordinal, ordinalValue,
- order, orderKey
- priority

tmadlener:
How hard is it to implement the whole thing with the `Rank` class? If it allows for hiding the `Obj*` from outside access, I would be in favor of that.

include/podio/detail/OrderKey.h
```diff
@@ -0,0 +1,34 @@
...
+// The friend free function design is used in order to reduce the coupling between interfaces and datatypes. Interfaces
+// do not need to be friends of datatypes to define the less-than comparison operator, which allows using datatypes from
+// different datamodels in an interface type.
+class OrderKey {
+public:
+  OrderKey(void* orderKey) : m_orderKey(reinterpret_cast<value_type>(orderKey)) {
+  }
+  friend bool operator<(const OrderKey& lhs, const OrderKey& rhs) {
+    return lhs.m_orderKey < rhs.m_orderKey;
+  }
```
tmadlener, line 22: Depending on when we merge this (i.e. before or after #698). Should we consider simply doing `operator<=>() = default;`? Or should we keep it simple for now and do that only if it becomes necessary?

python/templates/macros/declarations.jinja2
```diff
@@ -97,7 +97,7 @@
   bool operator!=(const {{ inverse_type }}& other) const { return !(*this == other); }
 
   // less comparison operator, so that objects can be e.g. stored in sets.
-  bool operator<(const {{ full_type }}& other) const { return m_obj < other.m_obj; }
+  bool operator<(const {{ full_type }}& other) const { return podio::detail::getOrderKey(*this) < podio::detail::getOrderKey(other); }
```
tmadlener, line 100: Is this strictly necessary? Wondering whether the compiler sees through all of this wrt optimizations.

tmadlener (commented):
Nice, thanks. From my point of view there are currently only two smallish questions. But in general, I like this and the feature it enables.

python/templates/macros/declarations.jinja2
m-fila: For datatypes this is not necessary, I just thought it could be easier to maintain if the same mechanism was used for both datatypes and interfaces. I'll have to take a look at the cost

include/podio/detail/OrderKey.h
m-fila: I think we don't really need the extra operators here unless we want to express the remaining datatypes and interface types comparisons with `OrderKey` too

include/podio/detail/OrderKey.h
```diff
@@ -0,0 +1,34 @@
...
+
+private:
+#ifdef UINTPTR_MAX
+  using value_type = uintptr_t;
+#else
+  using value_type = uintmax_t;
+#endif
+  value_type m_orderKey;
```
tmadlener, line 30:
```suggestion
  void* m_orderKey;
```

Just to have a papertrail of the [EDM4hep meeting today](https://indico.cern.ch/event/1484070/)

m-fila: Thanks, I'm having second thoughts on this now

It's super pedantic, but the standard guarantees implementation defined strict-ordering of all the pointers for comparisons done with `std::less` - comparing two pointers to unrelated objects with plain `<` yields 'unspecified' result. Taking look at the `std::less` for pointers from libstdc++ it does cast the pointers to internal numeric type then compare them with `<`.

So I think it's marginally better to either keep the casting or go with `less` rather than do plain comparison for pointers. On the other hand so far we had plain comparisons and haven't seen any problems yet

tmadlener: What are "unrelated objects" in this case? One way to exclude undefined behavior would be to mark the `operator<` as `constexpr`. In that case the compiler should refuse to compile it, I think. (Also we do have builds with *UBSan* and either that isn't able to flag this, or it's fine?)

If `std::less` solves all of this (with our without `void*`), I would just go for that here.

m-fila: If I understood [cppreference](https://en.cppreference.com/w/cpp/language/operator_comparison) ('Built-in pointer relational comparison' section) correctly, unspecified occurs unless the pointers are from the same array, or a pointer is from an array and another is past the array, or they are pointers to members of the same object

Yeah, let's do `std::less` and don't dwell on this

include/podio/detail/OrderKey.h
```diff
@@ -0,0 +1,28 @@
...
+// Internal class allowing datatype objects to be placed in data structures like maps and sets by providing a way to
+// compare them. The comparison is based on addresses of their internal data objects.
+//
+// This class is intended to be used as the return value of internal `podio::detail::getOrderKey` free functions. These
+// functions are friends of each datatype, allowing them to access the internal data objects and define less-than
+// comparison operators for both datatypes and interface types.
+//
+// The friend free function design is used in order to reduce the coupling between interfaces and datatypes. Interfaces
+// do not need to be friends of datatypes to define the less-than comparison operator, which allows using datatypes from
+// different datamodels in an interface type.
```
tmadlener, line 14:
```suggestion
/// Internal class allowing datatype objects to be placed in data structures like maps and sets by providing a way to
/// compare them. The comparison is based on addresses of their internal data objects.
///
/// This class is intended to be used as the return value of internal `podio::detail::getOrderKey` free functions. These
/// functions are friends of each datatype, allowing them to access the internal data objects and define less-than
/// comparison operators for both datatypes and interface types.
///
/// The friend free function design is used in order to reduce the coupling between interfaces and datatypes. Interfaces
/// do not need to be friends of datatypes to define the less-than comparison operator, which allows using datatypes from
/// different datamodels in an interface type.
```
Just to be consistent with other docstrings

tmadlener (commented):
Stupid question that didn't occur to me before: I suppose I/O works as expected, but I don't see a test for that?

m-fila:
Ups, I forgot about IO tests. Is it only checking relations and links?

tmadlener:
Yeah, basically the one place where it could break that is not covered by other tests is during relation resolving. So a minimal test would write some objects from the upstream model and some from the downstream model and use them to fill a relation where an interface is used. 

I don't suspect for anything to go wrong, but I would sleep better if there was a test telling me that ;)

m-fila:
Added IO tests

tmadlener (approved):
Thanks. Looks like nothing unexpected popped up :)

PR #715, opened 2024-12-02
Improve the detection of schema changes and add tests

### LINKED ISSUE #712: Schema evolution check script does not detect changes in non-members
The current version of the script that checks for impossible schema changes does not detect changes to `VectorMembers`, `OneToOneRelations` and `OneToManyRelations`. Instead it just let's schemas with these changes pass without warning or error. The schema evolution mechanism can in principle handle these, but will need some manual intervention obviously. As a first step, we should at least detect these changes and mark them as impossible, instead of letting users run into runtime issues when doing changes like these and trying to read old files.

tmadlener:
BEGINRELEASENOTES
- Improve the schema changes that `podio_schema_evolution.py` detects and reports on
  - Addition / removal of `VectorMembers`, `OneToOneRelations` and `OneToManyRelations`
- Make the script exit with a non-zero exit code in case an unsupported schema change is detected
- Add a test setup and some tests for testing the schema evolution script and make sure it detects unsupported changes

ENDRELEASENOTES

Fixes #712 

- [x] Implement detection of changes to `OneToOneRelations`
- [x] Implement detection of changes to `OneToManyRelations`

### DIFF
--- a/python/podio_schema_evolution.py
+++ b/python/podio_schema_evolution.py
@@ -4,6 +4,7 @@ Provides infrastructure for analyzing schema definitions for schema evolution
 """
 
 import yaml
+import sys
 
 from podio_gen.podio_config_reader import PodioConfigReader
 
@@ -423,6 +424,9 @@ class DataModelComparator:
             print("ERRORS:")
             for error in self.errors:
                 print(f" - {error}")
+            return False
+
+        return True
 
     def read(self) -> None:
         """read datamodels from yaml files"""
@@ -482,5 +486,6 @@ if __name__ == "__main__":
     comparator = DataModelComparator(args.new, args.old, evolution_file=args.evo)
     comparator.read()
     comparator.compare()
-    comparator.print_comparison()
+    if not comparator.print_comparison():
+        sys.exit(1)
     # print(comparator.get_changed_schemata(schema_filter=root_filter))
--- a/python/podio_schema_evolution.py
+++ b/python/podio_schema_evolution.py
@@ -369,6 +369,12 @@ class DataModelComparator:
                 f"Forbidden schema change in '{vmc.klassname}': Added vector member '{vmc.member}'"
             )
 
+        dropped_vecmems = [c for c in schema_changes if isinstance(c, DroppedVectorMember)]
+        for vmc in dropped_vecmems:
+            self.errors.append(
+                f"Forbidden schema change in '{vmc.klassname}': Added vector member '{vmc.member_name}'"
+            )
+
         # are the member changes actually supported/supportable?
         changed_members = [
             change for change in schema_changes if isinstance(change, ChangedMember)
--- a/python/podio_schema_evolution.py
+++ b/python/podio_schema_evolution.py
@@ -205,7 +205,7 @@ class DataModelComparator:
             ]
         )
 
-        self._compare_definitions(
+        self._compare_members(
             kept_components,
             self.datamodel_new.components,
             self.datamodel_old.components,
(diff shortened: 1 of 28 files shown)


tmadlener:
This works for now, but I am not extremely happy with the overall implementation at the moment. At some point there should be some refactoring of the whole detection system and script to make it easier to test and extend.

python/podio_schema_evolution.py
```diff
@@ -327,6 +425,33 @@ def heuristics(self):
         added_members = [change for change in schema_changes if isinstance(change, AddedMember)]
         self.heuristics_members(added_members, dropped_members, schema_changes)
 
+        for vmc in (c for c in schema_changes if isinstance(c, AddedVectorMember)):
+            self.errors.append(
+                f"Forbidden schema change in '{vmc.klassname}': Added vector member '{vmc.member}'"
+            )
+        for vmc in (c for c in schema_changes if isinstance(c, DroppedVectorMember)):
+            self.errors.append(
+                f"Forbidden schema change in '{vmc.klassname}': Added vector member '{vmc.member}'"
```
andresailer, line 434:
```suggestion
                f"Forbidden schema change in '{vmc.klassname}': Dropped vector member '{vmc.member}'"
```
?
What does vmc stand for? vector member change?
Why is this forbidden?

python/podio_schema_evolution.py
```diff
@@ -327,6 +425,33 @@ def heuristics(self):
...
+
+        for rc in (c for c in schema_changes if isinstance(c, AddedMultiRelation)):
+            self.errors.append(
+                f"Forbidden schema chage in '{rc.klassname}': Added OneToManyRelation '{rc.member}'"
+            )
+        for rc in (c for c in schema_changes if isinstance(c, DroppedMultiRelation)):
+            self.errors.append(
+                f"Forbidden schema chage in '{rc.klassname}': Dropped OneToManyRelation '{rc.member}'"
+            )
+
```
andresailer, line 454: Might be worth moving the error message into the specific schema change type for further simplification.
```suggestion
        for c in schema_changes:
            if isinstance(c, AddedVectorMember):
                self.errors.append(
                    f"Forbidden schema change in '{c.klassname}': Added vector member '{c.member}'"
                )
            if isinstance(c, DroppedVectorMember):
                self.errors.append(
                    f"Forbidden schema change in '{c.klassname}': Dropped vector member '{c.member}'"
               )
            if isinstance(c, AddedSingleRelation):
                self.errors.append(
                    f"Forbidden schema change in '{c.klassname}': Added OneToOneRelation '{c.member}'"
                )
            if isinstance(c, DroppedSingleRelation):
                self.errors.append(
                    f"Forbidden schema change in '{c.klassname}': Dropped OneToOneRelation '{cc.member}'"
               )
            if isinstance(c, AddedMultiRelation):
                self.errors.append(
                    f"Forbidden schema change in '{c.klassname}': Added OneToManyRelation '{c.member}'"
                )
            if isinstance(c, DroppedMultiRelation):
                self.errors.append(
                    f"Forbidden schema change in '{c.klassname}': Dropped OneToManyRelation '{c.member}'"
                )
```

python/podio_schema_evolution.py
```diff
@@ -327,6 +425,33 @@ def heuristics(self):
...
+                f"Forbidden schema change in '{vmc.klassname}': Added vector member '{vmc.member}'"
+            )
+        for vmc in (c for c in schema_changes if isinstance(c, DroppedVectorMember)):
+            self.errors.append(
+                f"Forbidden schema change in '{vmc.klassname}': Added vector member '{vmc.member}'"
+            )
+
+        for rc in (c for c in schema_changes if isinstance(c, AddedSingleRelation)):
+            self.errors.append(
+                f"Forbidden schema chage in '{rc.klassname}': Added OneToOneRelation '{rc.member}'"
```
andresailer, line 439:
```suggestion
                f"Forbidden schema change in '{rc.klassname}': Added OneToOneRelation '{rc.member}'"
```
Missing `n`s added in my other comment

python/podio_schema_evolution.py
tmadlener: Forbidden in this case is to be interpreted as *not implemented*. If users make this schema change trying to read old files will most likely crash with some sort of segmentation violation.

python/podio_schema_evolution.py
tmadlener: Actually we already have a message in the schema changes, so simply reusing that is indeed the way to go here. Thanks for the hint (trees and woods and such from my side).

PR #716, opened 2024-12-09
Mark the collection destructor with override and remove `#include <array>`

jmcarcell:
BEGINRELEASENOTES
- Mark the collection destructor with override and remove `#include <array>`

ENDRELEASENOTES

I think for `std::array` the only use-case in a `Collection.h` file is due to https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L857 and then since the `Object.h` is included and array is included there it's also not strictly needed in this file although used. In all other cases it's included without being used.

### DIFF
--- a/python/templates/Collection.h.jinja2
+++ b/python/templates/Collection.h.jinja2
@@ -22,7 +22,6 @@
 
 #include <string_view>
 #include <vector>
-#include <array>
 #include <algorithm>
 #include <ostream>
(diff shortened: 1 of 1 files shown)


tmadlener:
I think the `<array>` includes are a leftover from before these changes https://github.com/AIDASoft/podio/pull/337

tmadlener (approved):

PR #717, opened 2024-12-10
Delete all the non-virtual default destructors

jmcarcell:
BEGINRELEASENOTES
- Delete all the non-virtual default destructors. See https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c20-if-you-can-avoid-defining-default-operations-do

ENDRELEASENOTES

Triggered because `UserDataCollection`'s destructor was overriding without override and that was flagged in my editor, at that point all of them may as well be deleted. The side effect is a possibly useful comment in `Frame.h` and `Writer.h` removed.

### DIFF
--- a/include/podio/CollectionBufferFactory.h
+++ b/include/podio/CollectionBufferFactory.h
@@ -39,7 +39,6 @@ public:
   CollectionBufferFactory& operator=(CollectionBufferFactory const&) = delete;
   CollectionBufferFactory(CollectionBufferFactory&&) = delete;
   CollectionBufferFactory& operator=(CollectionBufferFactory&&) = delete;
-  ~CollectionBufferFactory() = default;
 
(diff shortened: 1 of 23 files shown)


m-fila (commented):
I think in most of these cases, if not all of them, it's actually better to keep the destructors since at least one of the assignment/copy is also deleted/defined. So by having destructors we actually obey the [rule of five](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c21-if-you-define-or-delete-any-copy-move-or-destructor-function-define-or-delete-them-all) ;)

tmadlener:
I think I agree with @m-fila here. I haven't checked all of them, but the ones I saw had at least one "special" case for one of the other constructors. In case all of them are defaulted, we should remove all of them. Otherwise we should indeed go for rule of five.

PR #719, opened 2024-12-19
Make sure RNTupleReader builds with ROOT > 6.34

tmadlener:
BEGINRELEASENOTES
- Make sure that the `RNTupleReader` still builds with ROOT > 6.34

ENDRELEASENOTES

`RException` (and other RNTuple API parts) have been moved out of the `Experimental` namespace with https://github.com/root-project/root/pull/17281

### DIFF
--- a/src/RNTupleReader.cc
+++ b/src/RNTupleReader.cc
@@ -10,6 +10,14 @@
 
 #include <memory>
 
+// Adjust for the move of this out of ROOT v7 in
+// https://github.com/root-project/root/pull/17281
+#if ROOT_VERSION_CODE >= ROOT_VERSION(6, 35, 0)
+using ROOT::RException;
+#else
+using ROOT::Experimental::RException;
+#endif
+
 namespace podio {
 
 template <typename T>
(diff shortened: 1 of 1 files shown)


tmadlener:
It looks like some ov the `dev3` stacks have not yet caught up with the latest ROOT (i.e. the `latest` symlink still points to beginning of the week, e.g. `Mon` or `Tue` for some of the stacks. `x86_64-ubuntu2404-gcc13-opt` is already on `Thu` and this works there).

@andresailer just in case you are not already aware of this.

jmcarcell:
Since there are going to be several changes like this one before the next version maybe it's simpler to keep a single PR instead of making N PRs.

tmadlener:
Yeah probably a good idea. Depending on how spread out the changes are, we will probably have to split it into a few. Keeping this open to collect a few of the things for now.

tmadlener:
It looks like things have settled a bit on the ROOT side, but this PR starts to block a few others on our side, so I will probably merge this rather soon.

PR #720, opened 2024-12-20
Make collection iterators `forward_iterator`* and beyond

m-fila:
BEGINRELEASENOTES
- Collection iterators fulfill `forward_iterator`, `bidirectional_iterator` and `random_access_iterator`concepts. The collections can be used with more categories of range algorithms - up to `random_access_range`, for example `std::adjacent_find`, `std::lower_bound`, `std::fold_right`, and more views like `std::ranges::views::reverse`. 
- `LinkCollection` and its iterators fulfill concepts up to `random_access_range` and `random_access_iterator`.
- Add reverse iterator methods to Collections, `LinkCollection` and `UserDataCollection`.
- **CollectionIterators of generated datamodels now define `operator<=>` instead of `operator!=`. This will require at least c++20 to compile, but should not affect existing behavior.**

ENDRELEASENOTES

**TL;DR** Iterators gain support up to `std::random_access_iterator` except that technically they don't meet a requirement for pointers obtained with `auto* ptr = it.operator->();` pattern to outlive the iterator

---- 

Here is a proposal how to make collection iterators fulfill `std::forward_iterator` which also opens the way for `std::bidirectional_iterator` and `std::random_access_iterator` by making a single **exception**

The collection iterators fulfill all the `forward_iterator` requirements **except** one semantic requirement:
>  Pointers and references obtained from a forward iterator into a range remain valid while the range exists. 

The problem here is the `operator->` that returns a pointer to data type object that is a member of the iterator (collections store data layer objects and data type are created on-the-fly so there is no data type object with lifetime livinf inside collection to which pointer we could return). Due to this the validity of obtained pointer is as long as the iterator is valid - to fulfill the requirement it should be as long the collection is valid.

In reality it's quite hard to abuse as things obtained through `->` (`iterator->energy()`) are safe since they go through the data layer that lives in the collection. The problem seems to be only getting the pointer directly with pattern like this:

```cpp
auto* ptr = iterator.operator->();
```

which is rather rare in normal usage. A pattern like this `auto particle = *(iterator.operator->());` again is okay since the pointer is used immediately 

Of course the final question is whether it's used somewhere in an implementation of the std algorithms. On top of that it should be used in a certain way like creating a temporary copy of iterator, taking pointer, disposing of the copy, using iterator

In the past there were some precedence event in the standard library that iterators were assigned some category **except** something (that was for *Legacy* iterators):
- [std::regex_iterator](https://en.cppreference.com/w/cpp/regex/regex_iterator)
  >std::regex_iterator is a read-only iterator that accesses the individual matches of a regular expression within the underlying character sequence. It meets the requirements of a [LegacyForwardIterator](https://en.cppreference.com/w/cpp/named_req/ForwardIterator), except that for dereferenceable values a and b with a == b, *a and *b will not be bound to the same object.
- [std::filesystem::path::iterator](https://en.cppreference.com/w/cpp/filesystem/path)
  >a constant [LegacyInputIterator](https://en.cppreference.com/w/cpp/named_req/InputIterator) with a value_type of path that meets all requirements of [LegacyBidirectionalIterator](https://en.cppreference.com/w/cpp/named_req/BidirectionalIterator) except that for two equal dereferenceable iterators a and b of type const_iterator, there is no requirement that *a and *b refer to the same object.
 It is unspecified whether const_iterator is actually a [LegacyBidirectionalIterator](https://en.cppreference.com/w/cpp/named_req/BidirectionalIterator)

### DIFF
--- a/doc/collections_as_container.md
+++ b/doc/collections_as_container.md
@@ -89,13 +89,14 @@ In the following tables a convention from `Collection` is used: `iterator` stand
 | `std::input_iterator` | ✔️ yes | ✔️ yes |
 | `std::output_iterator` | ❌ no | ❌ no |
 | `std::forward_iterator` | ✔️ yes (see note below) | ✔️ yes (see note below) |
-| `std::bidirectional_iterator` | ✔️ yes | ✔️ yes |
-| `std::random_access_iterator` | ✔️ yes | ✔️ yes |
+| `std::bidirectional_iterator` | ✔️ yes (see note below) | ✔️ yes (see note below) |
+| `std::random_access_iterator` | ✔️ yes (see note below) | ✔️ yes (see note below) |
 | `std::contiguous_iterator` | ❌ no | ❌ no |
 
-> [!NOTE]
->The collections' iterators fulfil the `std::forward_iterator` except that the pointers obtained with `->` remain valid only as long as the iterator is valid instead of as long as the range remain valid. In practice this means a `ptr` obtained with `auto* ptr = it.operator->();` is valid only as long as `it` is valid.
->The values obtained immediately through `->` (for instance `auto& e = it->energy();`) behaves as expected for `std::forward_iterator` as their validity is tied to the validity of a collection.
+:::{note}
+The collections' iterators fulfil the `std::forward_iterator` except that the pointers obtained with `->` remain valid only as long as the iterator is valid instead of as long as the range remain valid. In practice this means a `ptr` obtained with `auto* ptr = it.operator->();` is valid only as long as `it` is valid.
+The values obtained immediately through `->` (for instance `auto& e = it->energy();`) behaves as expected for `std::forward_iterator` as their validity is tied to the validity of a collection.
+:::
 
 ### LegacyIterator
 
@@ -207,7 +208,7 @@ std::fill(std::begin(collection), std::end(collection), value ); // requires For
 std::sort(std::begin(collection), std::end(collection)); // requires RandomAccessIterator and Swappable -> might compile, wrong result
 ```
 
-## Standard range algorithms
+## Collection and standard range algorithms
 
 The arguments of standard range algorithms are checked at compile time and must fulfil certain iterator concepts, such as `std::input_iterator` or `std::ranges::input_range`.
 
--- a/doc/collections_as_container.md
+++ b/doc/collections_as_container.md
@@ -88,10 +88,14 @@ In the following tables a convention from `Collection` is used: `iterator` stand
 | `std::input_or_output_iterator` | ✔️ yes | ✔️ yes |
 | `std::input_iterator` | ✔️ yes | ✔️ yes |
 | `std::output_iterator` | ❌ no | ❌ no |
-| `std::forward_iterator` | ❌ no | ❌ no |
+| `std::forward_iterator` | ✔️ yes (see note below) | ✔️ yes (see note below) |
 | `std::bidirectional_iterator` | ❌ no | ❌ no |
 | `std::random_access_iterator` | ❌ no | ❌ no |
 | `std::contiguous_iterator` | ❌ no | ❌ no |
+
+> [!NOTE]  
+>The collections' iterators fulfil the `std::forward_iterator` except that the pointers obtained with `->` remain valid only as long as the iterator is valid instead of as long as the range remain valid. In practice this means a `ptr` obtained with `auto* ptr = it.operator->();` is valid only as long as `it` is valid. 
+>The values obtained immediately through `->` (for instance `auto& e = it->energy();`) behaves as expected for `std::forward_iterator` as their validity is tied to the validity of a collection.
 
 ### LegacyIterator
 
@@ -180,7 +184,7 @@ In addition to the *LegacyForwardIterator* the C++ standard specifies also the *
 | `std::ranges::sized_range` | ✔️ yes |
 | `std::ranges::input_range` | ✔️ yes |
 | `std::ranges::output_range` | ❌ no |
-| `std::ranges::forward_range` | ❌ no |
+| `std::ranges::forward_range` | ✔️ yes |
 | `std::ranges::bidirectional_range` | ❌ no |
 | `std::ranges::random_access_range` | ❌ no |
 | `std::ranges::contiguous_range` | ❌ no |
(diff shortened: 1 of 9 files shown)


m-fila:
Ah used `<=>` to define comparison operators. I think I'll leave it with that since the move to C++20 is upcoming anyway

m-fila:
Now, yes :sunglasses:

doc/collections_as_container.md
```diff
@@ -88,11 +88,15 @@ In the following tables a convention from `Collection` is used: `iterator` stand
...
-| `std::bidirectional_iterator` | ❌ no | ❌ no |
-| `std::random_access_iterator` | ❌ no | ❌ no |
+| `std::forward_iterator` | ✔️ yes (see note below) | ✔️ yes (see note below) |
+| `std::bidirectional_iterator` | ✔️ yes | ✔️ yes |
+| `std::random_access_iterator` | ✔️ yes | ✔️ yes |
 | `std::contiguous_iterator` | ❌ no | ❌ no |
 
+> [!NOTE]
+>The collections' iterators fulfil the `std::forward_iterator` except that the pointers obtained with `->` remain valid only as long as the iterator is valid instead of as long as the range remain valid. In practice this means a `ptr` obtained with `auto* ptr = it.operator->();` is valid only as long as `it` is valid.
+>The values obtained immediately through `->` (for instance `auto& e = it->energy();`) behaves as expected for `std::forward_iterator` as their validity is tied to the validity of a collection.
```
tmadlener, line 98: How does this render in the final documentation that went through sphinx? I have used
````
```{note}
This syntax for making not boxes on the rendered documentation. They show up as simple markdown blocks without highlights in the rendered markdown document.
```
````

File: doc/collections_as_container.md [hunk 2]
```diff
@@ -88,11 +88,15 @@ In the following tables a convention from `Collection` is used: `iterator` stand
 | `std::input_or_output_iterator` | ✔️ yes | ✔️ yes |
 | `std::input_iterator` | ✔️ yes | ✔️ yes |
 | `std::output_iterator` | ❌ no | ❌ no |
-| `std::forward_iterator` | ❌ no | ❌ no |
-| `std::bidirectional_iterator` | ❌ no | ❌ no |
-| `std::random_access_iterator` | ❌ no | ❌ no |
+| `std::forward_iterator` | ✔️ yes (see note below) | ✔️ yes (see note below) |
+| `std::bidirectional_iterator` | ✔️ yes | ✔️ yes |
+| `std::random_access_iterator` | ✔️ yes | ✔️ yes |
```
tmadlener, line 93: Shouldn't the notes also apply to the `bidirectional_iterator` and `random_access_iterator`, as they refine the `forward_iterator`?

tests/unittests/std_interoperability.cpp
```diff
@@ -495,6 +495,239 @@ TEST_CASE("Collection and iterator concepts", "[collection][container][iterator]
...
+    {
+      REQUIRE(iterator{} == iterator{});
+      auto coll = CollectionType();
+      coll.create();
+      auto i = coll.begin();
+      auto j = coll.begin();
+      REQUIRE(i == j);
+      REQUIRE(++i == ++j);
+      i = coll.begin();
+      REQUIRE(((void)[](auto x) { ++x; }(i), *i) == *i);
```
tmadlener, line 511: This line is a bit too dense for me to parse without a comment. I assume it is using the comma operator to call the lambda and is then comparing the iterators after it has been incremented inside the lambda? Not sure if the `void` is simply there to discard the lambda return value? or whether it's an actual cast.

Could you write a short comment about why and how this is done to give future us a chance of understanding of what is going on?

tmadlener (commented):
I suppose this might need some rebasing to catch up with the latest developments, but it also doesn't contain anything that conflicts with any other recent developments.

I have a few minor comments inline. Otherwise this looks OK to me. Thanks for all the work.

python/templates/macros/iterator.jinja2
```diff
@@ -19,8 +21,8 @@ public:
   {{ iterator_type }}& operator=({{iterator_type}}&&) = default;
   ~{{ iterator_type }}() = default;
 
-  bool operator!=(const {{ iterator_type}}& x) const {
-    return m_index != x.m_index;
+  auto operator<=>(const {{ iterator_type}}& other) const {
+    return m_index <=> other.m_index;
```
tmadlener, line 25: This might need a bold statement in the release notes as it might affect behavior of generated datamodels.

doc/collections_as_container.md
m-fila: It would render badly.  I'll go with `:::{note}` so it still renders as text in plain markdown

doc/collections_as_container.md
m-fila: Good idea, added

doc/collections_as_container.md
```diff
@@ -88,11 +88,16 @@ In the following tables a convention from `Collection` is used: `iterator` stand
...
-| `std::forward_iterator` | ❌ no | ❌ no |
-| `std::bidirectional_iterator` | ❌ no | ❌ no |
-| `std::random_access_iterator` | ❌ no | ❌ no |
+| `std::forward_iterator` | ✔️ yes (see note below) | ✔️ yes (see note below) |
+| `std::bidirectional_iterator` | ✔️ yes (see note below) | ✔️ yes (see note below) |
+| `std::random_access_iterator` | ✔️ yes (see note below) | ✔️ yes (see note below) |
 | `std::contiguous_iterator` | ❌ no | ❌ no |
 
+:::{note}
+The collections' iterators fulfil the `std::forward_iterator` except that the pointers obtained with `->` remain valid only as long as the iterator is valid instead of as long as the range remain valid. In practice this means a `ptr` obtained with `auto* ptr = it.operator->();` is valid only as long as `it` is valid.
```
tmadlener, line 97:
```suggestion
The collections' iterators fulfil the `std::forward_iterator` except that the pointers obtained with `->` remains valid only as long as the iterator is valid instead of as long as the range remains valid. In practice this means a `ptr` obtained with `auto* ptr = it.operator->();` is valid only as long as `it` is valid.
```
Should we also add that this is something that probably doesn't (commonly) occur in "real world usage"

include/podio/detail/LinkCollectionIterator.h
```diff
@@ -29,8 +31,8 @@ class LinkCollectionIteratorT {
   LinkCollectionIteratorT& operator=(LinkCollectionIteratorT&&) = default;
   ~LinkCollectionIteratorT() = default;
 
-  bool operator!=(const LinkCollectionIteratorT& other) const {
-    return m_index != other.m_index;
+  auto operator<=>(const LinkCollectionIteratorT& other) const {
+    return m_index <=> other.m_index;
   }
 
   bool operator==(const LinkCollectionIteratorT& other) const {
```
tmadlener, line 38: If we have `operator<=>` we don't need `operator==` any longer, right?

python/templates/macros/iterator.jinja2
tmadlener: Also same here, we can remove `operator==` since it will be synthesized from `operator <=>`

tests/unittests/std_interoperability.cpp
```diff
@@ -497,6 +497,243 @@ TEST_CASE("Collection and iterator concepts", "[collection][container][iterator]
...
+      auto coll = CollectionType();
+      coll.create();
+      auto i = coll.begin();
+      auto j = coll.begin();
+      // multi-pass guarantee
+      REQUIRE(i == j);
+      REQUIRE(++i == ++j);
+      i = coll.begin();
+      REQUIRE(*i == ((void)[](auto x) { ++x; }(i), *i)); // iterating copy doesn't cause side effects that affect the
+                                                         // original iterator
```
tmadlener, line 515: Thanks :)

tests/unittests/std_interoperability.cpp
m-fila: Well I copied this check from the [reference](https://en.cppreference.com/w/cpp/iterator/forward_iterator). But the sides of comparison probably should be switched.

The idea is that iterating a copy of an iterator doesn't have any side effects that could affect the original. For instance, some iterator could consume the elements on iteration and not locally cache the last consumed element

tmadlener: Yeah, with the comment the intent is much more clear and with that also the implementation makes sense (even though it is still dense).

include/podio/detail/LinkCollectionIterator.h
m-fila: Both are needed. Only defaulted`operator<=>` creates an implicit `operator==`.
It's a bit convoluted https://en.cppreference.com/w/cpp/language/default_comparisons > Equality comparison > Implicit declaration

tmadlener: Huh, TIL.

python/templates/macros/iterator.jinja2
m-fila: I think I missed how this affects existing datamodels. What kind of bold statement should we include?

tmadlener: I think something like

- **`CollectionIterators` of generated datamodels now define `operator<=>` instead of `operator!=`. This will require at least c++20 to compile, but should not affect existing behavior.**

m-fila: I see, ok

tmadlener (approved):

PR #722, opened 2025-01-06
[WIP] Fix the test discovery on Ubuntu 24 for LCG/dev3 environments

tmadlener:
BEGINRELEASENOTES
- (Re-)introduce `DL_PATHS` for the `catch_discover_tests` call to make sure test discovery is run in the correct environment to avoid symbol lookup errors.

ENDRELEASENOTES

This has been removed in #588 and has been working fine for some time, but started to fail again on Ubuntu 24. I am not sure why but this has become necessary again for running on Ubuntu24 in an LCG/dev3 environment to run test discovery succesfully. Otherwise the environment for discovery is not set properly and running the executable (during the build stage) fails with a symbol lookup error. The latter is caused by the fact that, LCGs podio is not built with SIO support.

- [ ] Includes #719 
#698 will also need this to be merged first most likely.


tmadlener:
CI from #698 suggests that this might not be necessary after all.

PR #724, opened 2025-01-16
Refactor the read tests partially

tmadlener:
BEGINRELEASENOTES
- Start to refactor the read tests partially to break up the large `processEvent` function into smaller more manageable chunks.
- Fix a small logic bug in the creation of the test files

ENDRELEASENOTES

This is pulled out of #504 and probably part of a larger effort of cleaning up the I/O test code further. For now I have just done the ones that are necessary for #504

### DIFF
--- a/tests/read_test.h
+++ b/tests/read_test.h
@@ -12,6 +12,7 @@
 #include "datamodel/ExampleWithNamespace.h"
 #include "datamodel/ExampleWithOneRelationCollection.h"
 #include "datamodel/ExampleWithVectorMemberCollection.h"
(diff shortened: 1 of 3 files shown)


tests/read_test.h
```diff
@@ -12,6 +12,7 @@
 #include "datamodel/ExampleWithNamespace.h"
 #include "datamodel/ExampleWithOneRelationCollection.h"
 #include "datamodel/ExampleWithVectorMemberCollection.h"
+#include "datamodel/MutableExampleHit.h"
```
tmadlener, line 15:
```suggestion
```

PR #725, opened 2025-01-17
make `LinkCollectionIterator` fulfil `std::input_iterator` concept

m-fila:
BEGINRELEASENOTES
- Updated `LinkCollectionIterator` to fulifl the same iterator concept and iterator category as other collection iterators. `LinkCollection` fulfills the same range concepts as other collections. Algorithms like `std::ranges::find` or `std::count` are supported.

ENDRELEASENOTES

This is an update for `LinkCollectionIterator` to behave like component collection iterator -> be *LegacyInputIterator* and fulfil `std::input_iterator` which make it compatible with a broad range of standard algorithms.

For `LinkCollection` I don't add a documentation page with all the requirements. The test are also simpler and the tests for not supported requirements are skipped

### DIFF
--- a/include/podio/detail/LinkCollectionIterator.h
+++ b/include/podio/detail/LinkCollectionIterator.h
@@ -3,6 +3,7 @@
 
 #include "podio/detail/LinkFwd.h"
 #include "podio/utilities/MaybeSharedPtr.h"
+#include <iterator>
 
 namespace podio {
 template <typename FromT, typename ToT, bool Mutable>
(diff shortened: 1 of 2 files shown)


tmadlener:
Thanks. Is there an easy way to ensure that we don't forget about these e.g. in #720 ?

m-fila:
> Thanks. Is there an easy way to ensure that we don't forget about these e.g. in #720 ?

I'm afraid I don't have any ideas

tmadlener:
Stop adding more "special collections" it is then ;).

tmadlener (approved):

PR #726, opened 2025-01-17
Remove iterator-testing code redundant in C++20

m-fila:
BEGINRELEASENOTES
- Removed redundant check for C++20
- Removed iterator test that is no longer required in C++

ENDRELEASENOTES

Some leftovers from #698
The tests`std::iterator_traits<It>::value_type` were originally surrounded by `#if (__cplusplus < 202002L)` and could be removed


tmadlener (approved):

m-fila:
I forgot to remove the redundant check from documentation. Should be good now

PR #727, opened 2025-01-20
Make `RelationRange` a proper view

m-fila:
BEGINRELEASENOTES
- Make `RelationRange` fulfill the `std::ranges::view` and `std::ranges::borrowed_range` concepts

ENDRELEASENOTES

As suggested in https://github.com/AIDASoft/podio/pull/626#issuecomment-2373854377 adding tests for checking whether `RelationRange` is a std range  - it is.

Since the `RelationRange` is a very lightweight, non-owning view it can fulfill two more concepts that can be used by the std range algorithms and adapters to avoid unneeded operations:
- `view` - it's cheap to move/copy as it stores only iterators to the data living elsewhere
- `borrowed_range` - the iterators taken from range remain valid even when the range is destroyed since they are not local to the range

The `RelationRange` is very similar to `std::span` and `std::ranges::subrange` and could potential replaced by them except for two differences:
- `RealationRange` has `at` for bound-check element access whereas the alternatives don't (until C++26)
- `RelationRange` element access returns copy whereas the alternatives would return const reference. Since the elements are podio objects then copy is preferred.

The `RelationRange` could be also re-implemented with `std::ranges::view_interface` but it comes with issues like above so I think isn't really worth it


tmadlener (approved):
Nice. Thank you.

PR #728, opened 2025-01-21
collections.jinja2: improve exception message for untracked relation

veprbl:
BEGINRELEASENOTES
- Improved exception message for the case when relation to untracked object is to be persisted.

ENDRELEASENOTES

### DIFF
--- a/python/templates/macros/collections.jinja2
+++ b/python/templates/macros/collections.jinja2
@@ -29,7 +29,7 @@ std::vector<{{ member.full_type }}> {{ class.bare_type }}Collection::{{ member.n
 {{ _relation_index_handling(relation) }}
     for (const auto& it : (*m_rel_{{ relation.name }}_tmp[i])) {
       if (it.getObjectID().index == podio::ObjectID::untracked) {
-        throw std::runtime_error("Trying to persistify untracked object");
+        throw std::runtime_error("Trying to persistify relation to untracked object");
       }
       m_refCollections[{{ index }}]->emplace_back(it.getObjectID());
     }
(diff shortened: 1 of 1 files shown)


tmadlener (commented):
Thanks a lot. I am wondering whether we should also put some more information there, e.g. the name of the relation?

Also, a similar check for single relations doesn't exist. I am no longer sure why we treat them differently, but arguably there should not be a difference, I think.

veprbl:
That's a good name. A single relation can be checked too, but that would be a breaking change for the users.

tmadlener (approved):
Yeah that is true. Let's leave it at the non-breaking improvement for now. I will make a note of figuring out why one-to-one and one-to-many relations are treated differently.

python/templates/macros/collections.jinja2
```diff
@@ -46,7 +46,7 @@ std::vector<{{ member.full_type }}> {{ class.bare_type }}Collection::{{ member.n
 {% set real_index = start_index + index %}
   for (auto& obj : entries) {
     if (obj->m_{{ relation.name }}) {
-      m_refCollections[{{ real_index }}]->emplace_back(obj->m_{{ relation.name }}->getObjectID());
+       m_refCollections[{{ real_index }}]->emplace_back(obj->m_{{ relation.name }}->getObjectID());
     } else {
        m_refCollections[{{ real_index }}]->push_back({podio::ObjectID::invalid, 0});
```
jmcarcell, line 51:
```suggestion
      m_refCollections[{{ real_index }}]->emplace_back(obj->m_{{ relation.name }}->getObjectID());
    } else {
      m_refCollections[{{ real_index }}]->push_back({podio::ObjectID::invalid, 0});
```
two spaces instead of three

PR #729, opened 2025-01-27
add globbing support in `makeReader` and `CreateDataSource`

### LINKED ISSUE #686: Support for globbing in `podio::CreateDataFrame`, `podio::makeReader` and `podio.reading.get_reader`
tl; dr:

```python
import podio

# breaks
# reader = podio.reading.get_reader("some/path/*.edm4hep.root")

# works
reader = podio.root_io.Reader("some/path/*.edm4hep.root")

```

The `ROOTReader` was (and technically still is) able to read from multiple files at once using globbing as the names are passed to `TChain` directly

https://github.com/AIDASoft/podio/blob/46b02cfaa00b40f1356a1670860bd57ee533b0cd/src/ROOTReader.cc#L240-L251

Using the new `Reader` interface this does not work any more as every path is passed to `TFile` first

https://github.com/AIDASoft/podio/blob/46b02cfaa00b40f1356a1670860bd57ee533b0cd/src/Reader.cc#L38-L44

This is particularly annoying when using the podio DataSource as the normal RDataFrame behaviour is also to pass everything to TChain. I know this was probably never explicitly supported but it would be great to get this back :)

m-fila:
BEGINRELEASENOTES
- POSIX glob patterns can be used in `makeReader` and `CreateDataSource`. Added standalone helper `podio::utilities::expand_glob`to resolve globs.
- Added passing a list of files to `get_reader`

ENDRELEASENOTES

This is a proposition to make  `makeReader` and `CreateDataSource` overloads a taking single filename also accept and expand a POSIX glob pattern.

The behaviour is:
- if input isn't a glob pattern it's forwarded without expansion (regardless it exists or not)
- if no matches are found then an exception is thrown
- if any other error (memory allocation) is reported then an exception is thrown
- if some expanded paths can't be read message with path name and corresponding error code is printed to cerr. The problematic paths are then skipped in the results, but the expansion overall isn't aborted.

Closes #686 

- [x] `podio::makeReader`
- [x] `podio::CreateDataFrame`
~~- [x] Python side - `podio.get_reader`~~
- [x] Python side - `podio.CreateDataFrame`


python/podio/reading.py
```diff
@@ -50,32 +52,42 @@ def _determine_root_format(filename):
...
+            is able to process the input files.
 
     Raises:
-        ValueError: If the file cannot be recognized, or if podio has not been
+        ValueError: If the files cannot be recognized, or if podio has not been
             built with the necessary backend I/O support
     """
+    if isinstance(filenames, str):
+        expanded_filenames = glob(filenames)
+        if not expanded_filenames:
```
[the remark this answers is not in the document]
m-fila: Remote access, hidden files unless glob pattern explicitly starts with `.` (for completeness). Also error message that such file doesn't exist is probably more helpful than that the input is empty

tmadlener:
Just for completeness bringing the conclusions from [todays meeting](https://indico.cern.ch/event/1511117/) here:

- `podio::detail::expand_glob` should become a publicly available utility (also available from python)
- `makeReader(std::string)` will keep the implicit globbing (via calling the aforementioned utility). Similar for the `CreateDataFrame` that takes only a single string. In both cases the versions that take a file list should not change.
- The python side should not have implicit globbing (i.e. the "hidden" `glob` that is currently present should be removed).
- The python interface should be able to take a list of input files though
- (As a bonus goal to be done in another PR, we should check and see if the c++ and python API can be harmonized a bit better).

include/podio/utilities/Glob.h
```diff
@@ -0,0 +1,38 @@
+#ifndef PODIO_UTILITIES_GLOB_H // NOLINT(llvm-header-guard): internal headers confuse clang-tidy
```
tmadlener, line 1:
```suggestion
#ifndef PODIO_UTILITIES_GLOB_H // NOLINT(llvm-header-guard): internal headers confuse clang-tidy
```
Is the `NOLINT(llvm-header-guard)` really necessary here? I think in the public headers this should work as expected.

include/podio/utilities/Glob.h
```diff
@@ -0,0 +1,38 @@
...
+#include <string>
+#include <vector>
+
+namespace podio::utils {
+/**
+ * @brief Expands a given glob pattern into a list of matching file paths.
+ *
+ * This function takes a glob pattern as input and returns a vector of strings
+ * containing the paths that match the pattern. It supports standard glob rules
+ * extended with tilde expansion and brace expansion If the pattern doesn't
```
tmadlener, line 12:
```suggestion
 * extended with tilde expansion and brace expansion. If the pattern doesn't
```

python/podio/reading.py
```diff
@@ -50,32 +50,41 @@ def _determine_root_format(filename):
     return RootFileFormat.RNTUPLE
 
 
-def get_reader(filename):
-    """Get an appropriate reader for the passed file.
+def get_reader(filenames):
+    """Get an appropriate reader for the passed files.
```
tmadlener, line 54:
```suggestion
    """Get an appropriate reader for the passed files.

```
To make API documentation formatted nicely with a brief and long description.

python/podio/utils.py
```diff
@@ -5,6 +5,11 @@
 from collections.abc import Iterable
 from pathlib import Path
 
+import ROOT
+
+if ROOT.gInterpreter.LoadFile("podio/podioVersion.h") != 0:  # noqa: E402
+    raise ImportError("Cannot find the podio/podioVersion.h header")
+
```
tmadlener, line 12: Should this be the `podio/utilities/Glob.h`? I am not sure why we need the version header here.

m-fila: Yes the intent was to have `podio/utilities/Glob.h` but I think it isn't actually needed

src/Glob.cc
```diff
@@ -0,0 +1,64 @@
+#include "podio/utilities/Glob.h"
+#include <iostream>
+#include <stdexcept>
+#include <string>
+#include <vector>
+#if __has_include(<glob.h>)
+  #include <glob.h>
+#else
+  #include <system_error>
+#endif
```
tmadlener, line 10:
```suggestion
#if __has_include(<glob.h>)
  #include <glob.h>
  #define PODIO_HAS_GLOB_SUPPORT 1
#else
  #include <system_error>
  #define PODIO_HAS_GLOB_SUPPORT 0
#endif
```
Just to make it possible to have some checks downstream on this.

python/podio/utils.py
tmadlener: No, it should be covered by the dictionary where we export the functions

src/selection.xml
```diff
@@ -46,5 +46,8 @@
     <class name="podio::LinkData"/>
     <class name="std::vector<podio::LinkData>"/>
 
+    <function name="podio::utilities::is_glob_pattern"/>
+    <function name="podio::utilities::expand_glob"/>
```
tmadlener, line 50:
```suggestion
    <function name="podio::utils::is_glob_pattern"/>
    <function name="podio::utils::expand_glob"/>
```

Correct spelling of the namespace

src/Glob.cc
m-fila: added in header

src/selection.xml
tmadlener: The magic of ROOT ;)

tmadlener (approved):

PR #736, opened 2025-02-14
fix setting link with interface without specific link direction

### LINKED ISSUE #735: Templated `Link::set` does not work if the `FromT` is an interface type
Discovered this via the `edm4hep::TrackerHitSimTrackerHitLink`.

The problem is the incomplete check the `if constexpr` here

https://github.com/AIDASoft/podio/blob/ef5135095f5890d3dd18aed1c98731aca3b67163/include/podio/detail/Link.h#L258-L268

If an object that could be implicitly converted to the interface type is passed, it will not pass this `if constexpr` check and will be passed to `setTo`, where it will then fail to compile due to type mismatches.

m-fila:
BEGINRELEASENOTES
- fix setting link with interface without specifying link direction

ENDRELEASENOTES

Changing `set` to be symmetric with respect to `ToT` and `FromT` as well as detecting possible ambiguity when linking two different interfaces

closes #735

### DIFF
--- a/include/podio/detail/Link.h
+++ b/include/podio/detail/Link.h
@@ -272,7 +272,7 @@ public:
                          !detail::isInterfaceInitializableFrom<FromT, T>) {
       setTo(std::move(value));
     } else {
-      static_assert(false, "Argument type is ambiguous, can't determine link direction");
+      static_assert(detail::always_false<T>, "Argument type is ambiguous, can't determine link direction");
     }
   }
 
--- a/include/podio/detail/Link.h
+++ b/include/podio/detail/Link.h
@@ -5,6 +5,7 @@
 #include "podio/detail/LinkObj.h"
 #include "podio/utilities/MaybeSharedPtr.h"
 #include "podio/utilities/TypeHelpers.h"
+#include <type_traits>
 
 #ifdef PODIO_JSON_OUTPUT
   #include "nlohmann/json.hpp"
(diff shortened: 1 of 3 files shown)


include/podio/detail/Link.h
```diff
@@ -262,8 +263,16 @@ class LinkT {
...
+    } else if constexpr (std::is_same_v<T, ToT>) {
+      setTo(std::move(value));
+    } else if constexpr (detail::isInterfaceInitializableFrom<FromT, T> &&
+                         !detail::isInterfaceInitializableFrom<ToT, T>) {
+      setFrom(std::move(value));
+    } else if constexpr (detail::isInterfaceInitializableFrom<ToT, T> &&
+                         !detail::isInterfaceInitializableFrom<FromT, T>) {
       setTo(std::move(value));
+    } else {
+      static_assert(false, "Argument type is ambiguous, can't determine link direction");
```
tmadlener, line 275: Wondering whether this should go into a `requires` clause on the outset, but I don't think we run into the same issue here as with the interface initialization (see #683), but maybe I am missing something?

tmadlener (commented):
That was quick. Thanks :)

include/podio/detail/Link.h
m-fila: Initially I put some extra clauses in the `requires` but then in became quite complicated logical expression, so I think this way it's more readable. It also acts as a fail safe if we miss some case in the `if constexpr` above - without it we would have to either break again the `FromT`/`ToT` symmetry and change last `else if constexpr` to `else` or have empty `else`.

I think it should be fine at least for now as there aren't any other overloads for `Link::set` (and it's a simple method not a constructor as in #683).

m-fila: > That makes sense. Minor nitpick on whether we need the `always_false` helper; In principle a `std::false_type{}()` should also work in the `static_assert`, I think. Then we wouldn't have to introduce the extra helper.

Unfortunately`std::false_type` won't work as it doesn't depend on template argument, so clang doesn't need to instantiate it and will still happily evaluate discarded `else` branch.

tmadlener: OK. gcc and clang >= 18 seems to swallow it with c++20, but we will just go with what is present now.

PR #738, opened 2025-02-19
add `std::hash` for datatype objects, interfaces and links

m-fila:
BEGINRELEASENOTES
- Add `std::hash` for podio datatype objects, interfaces and links. Datatypes, interfaces and links can be used in unordered associative containers such as `std::unordered_set` or `std::unordered_map`

ENDRELEASENOTES

As suggested in https://github.com/AIDASoft/podio/pull/733#issuecomment-2665089308 adding specializations of `std::hash` for datatypes, interfaces and links. In all the cases the hash boils down to hashing internal object pointer - the hashes are the same if the underlying object is the same  not if the properties are the same (similarly to what is done with `operator<` and `operator==`).

### DIFF
--- a/include/podio/detail/Link.h
+++ b/include/podio/detail/Link.h
@@ -5,13 +5,14 @@
 #include "podio/detail/LinkObj.h"
 #include "podio/utilities/MaybeSharedPtr.h"
 #include "podio/utilities/TypeHelpers.h"
-#include <type_traits>
 
 #ifdef PODIO_JSON_OUTPUT
   #include "nlohmann/json.hpp"
 #endif
 
+#include <functional>
 #include <ostream>
+#include <type_traits>
 #include <utility> // std::swap
 
 namespace podio {
@@ -33,6 +34,7 @@ class LinkT {
   friend LinkCollection<FromT, ToT>;
   friend LinkCollectionIteratorT<FromT, ToT, Mutable>;
   friend LinkT<FromT, ToT, !Mutable>;
+  friend std::hash<LinkT>;
 
   /// Helper member variable to check whether FromU and ToU can be used for this
   /// Link. We need this to make SFINAE trigger in some cases below
@@ -366,5 +368,12 @@ void to_json(nlohmann::json& j, const podio::LinkT<FromT, ToT, false>& link) {
 #endif
 
 } // namespace podio
+
+template <typename FromT, typename ToT, bool Mutable>
+struct std::hash<podio::LinkT<FromT, ToT, Mutable>> {
+  std::size_t operator()(const podio::LinkT<FromT, ToT, Mutable>& obj) const {
+    return std::hash<typename podio::LinkT<FromT, ToT, Mutable>::LinkObjT>{}(obj.get());
+  }
+};
 
 #endif // PODIO_DETAIL_LINK_H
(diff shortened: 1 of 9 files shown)


include/podio/detail/Link.h
```diff
@@ -367,4 +369,11 @@ void to_json(nlohmann::json& j, const podio::LinkT<FromT, ToT, false>& link) {
 
 } // namespace podio
 
+template <typename FromT, typename ToT, bool Mutable>
+struct std::hash<podio::LinkT<FromT, ToT, Mutable>> {
+  std::size_t operator()(const podio::LinkT<FromT, ToT, Mutable>& obj) const {
+    return std::hash<typename podio::LinkT<FromT, ToT, Mutable>::LinkObjT>{}(obj.get());
```
tmadlener, line 375: Even with having written the `LinkT` class I am not sure I understand what `obj.get()` actually returns and why this even compiles. AFAIU there is no untemplated `get` on `LinkT` (public or private). Shouldn't this be a `obj.m_obj.get()` to have the same behavior as other classes? What am I missing here?

tests/unittests/links.cpp
```diff
@@ -154,6 +158,64 @@ TEST_CASE("Link basics", "[links]") {
...
+  link3.set(cluster2);
+  link3.setWeight(3.0);
+  link4.set(hit1);
+  link4.set(cluster1);
+  link4.setWeight(4.0);
+
+  std::set<TestL> linkSet;
+  linkSet.insert(link1);
+  linkSet.insert(link2);
+  linkSet.insert(link2);
```
tmadlener, line 181: I think this would be more explicit by using `auto link5 = link2` and then `linkSet.insert(link5)`. We would then also show case that its the underlying object that matters not the handle that we use. Similar for the cases with `map` below.

tests/unittests/unittest.cpp
```diff
@@ -455,6 +494,34 @@ TEST_CASE("AssociativeContainer", "[basics]") {
...
+  cUnorderedSet.insert(clu3);
+  cUnorderedSet.insert(clu4);
+  cUnorderedSet.insert(clu5);
+  cUnorderedSet.insert(clu1);
+  cUnorderedSet.insert(clu2);
+  cUnorderedSet.insert(clu3);
+  cUnorderedSet.insert(clu4);
+  cUnorderedSet.insert(clu5);
+
+  REQUIRE(cUnorderedSet.size() == 5);
```
tmadlener, line 512:
```suggestion
  REQUIRE(cUnorderedSet.size() == 5);

```
Minor nit to make association of assertion and setup easier to see.

tmadlener (commented):
Very nice, thanks. There is one thing that I don't understand for the implementation for links. would it be possible to have similar lower level tests working on the `std::hash` directly for the links and interfaces? That would make it easier to spot potential issues and would make the whole testing a bit more symmetric.

include/podio/detail/Link.h
m-fila: Right, that's a typo, it definitely should be `obj.m_obj.get()`. I don't quite understand why this even compiles :neutral_face:

tests/unittests/interface_types.cpp
```diff
@@ -61,6 +61,29 @@ TEST_CASE("InterfaceTypes static checks", "[interface-types][static-checks]") {
   STATIC_REQUIRE(TypeWithEnergy::typeName == "TypeWithEnergy"sv);
 }
 
+TEST_CASE("InterfaceTypes hash", "[interface-types][hash]") {
+  auto hit = ExampleHit();
+  auto wrapper1 = TypeWithEnergy(hit);
+  auto hash1 = std::hash<TypeWithEnergy>{}(wrapper1);
+
```
tmadlener, line 68:
```suggestion
  REQUIRE(hash1 == std::hash<ExampleHit>{}(hit));
```
Just to be very clear and documenting about this. (I am assuming that the hashes are equal here, but it doesn't actually matter, we should just put this here explicitly).

m-fila: Good idea!

tmadlener (approved):

PR #740, opened 2025-02-19
Use `sizeof(T) == 0` instead of always_false

jmcarcell:
Introduced in https://github.com/AIDASoft/podio/pull/737 (also see https://github.com/AIDASoft/podio/pull/736#discussion_r1957837515) I think it is not necessary to add that and the included header can be removed.

> This is turbo pedantic but

Sorry :(

BEGINRELEASENOTES
- Use `sizeof(T) == 0` instead of `always_false`

ENDRELEASENOTES

### DIFF
--- a/include/podio/detail/Link.h
+++ b/include/podio/detail/Link.h
@@ -272,7 +272,7 @@ public:
                          !detail::isInterfaceInitializableFrom<FromT, T>) {
       setTo(std::move(value));
     } else {
-      static_assert(sizeof(T) && false, "Argument type is ambiguous, can't determine link direction");
+      static_assert(sizeof(T) == 0, "Argument type is ambiguous, can't determine link direction");
     }
   }
 
--- a/include/podio/detail/Link.h
+++ b/include/podio/detail/Link.h
@@ -272,7 +272,7 @@ public:
                          !detail::isInterfaceInitializableFrom<FromT, T>) {
       setTo(std::move(value));
     } else {
-      static_assert(detail::always_false<T>, "Argument type is ambiguous, can't determine link direction");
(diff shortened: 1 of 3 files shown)


m-fila:
In my defense [P2593R1](https://isocpp.org/files/papers/P2593R1.html), accepted for C++23. But I agree in the end it doesn't really matter. We could also go back to `sizeof(T) == 0`

jmcarcell:
Ah I didn't get that it was `static_assert(false)` what wasn't compiling, I thought it was `sizeof(T) == 0`

m-fila:
Yeah, `sizeof(T) && false`, `sizeof(T) == 0` and `always_false` are workarounds for `static_assert(false)` not behaving as desired. It's just that according to the standard  `always_false` actually works,  while the rest work by accident on most of the compilers

tmadlener:
Given that the current `always_false` is a workaround for `static_assert(false)` only starting to work as expected in c++23. I would be in favor of keeping `always_false` because it's very clear in its name, whereas `sizeof(T) == 0` needs some a bit more knowledge about why it is there.

jmcarcell:
I don't have a strong opinion about this one. My preference would be not to include something that may or not be removed when switching to C++23 when there is already a (simple) way. So feel free to do whatever you want with this PR.

m-fila:
> My preference would be not to include something that may or not be removed when switching to C++23 when there is already a (simple) way.

Whichever workaround we choose now with C++23 we can replace it with `static_assert(false)`

tmadlener:
Let's keep the `always_false` because I think it brings across the intended meaning a bit more obvious than the `sizeof(T) == 0` version.

PR #743, opened 2025-03-04
Prevent assignment to temporary datatype and link objects

### LINKED ISSUE #281: Changes to dereferenced collection iterators are not reflected in the underlying collection
I have to point out that being able to compile code that has incorrect results is already there in `master`. This PR doesn't introduce anything new in that sense (though I agree that it may increase expectations among users that some algorithms will work when they don't, and that we should fail to compile for things that don't work). Using the current `master` branch the following fairly straightforward example compiles but fails to do what was intended:
```cpp
  auto collection = ExampleClusterCollection();
  collection.create();
  for (auto i = collection.begin(); i != collection.end(); ++i) {
    *i = MutableExampleCluster(42);
  }
  REQUIRE(collection.begin()->energy() == 42);
```

_Originally posted by @wdconinc in https://github.com/AIDASoft/podio/issues/273#issuecomment-1090493117_

m-fila:
BEGINRELEASENOTES
- Prevent error-prone copy-assignment to temporary datatype and link objects. The compilation error instead of unexpected behaviour when by mistake using the podio collection iterators as "LegacyOutputIterator" is more very likely (e.g. `std::fill` will most likely not compile instead compiling but giving wrong results)

ENDRELEASENOTES

Datatypes act as rebindable custom references  - `particle = different_particle` is supposed to rebind the `particle` to have the same internal objects as `different_particle` instead of modifying the `particle` internal object.

With such semantics it makes little sense to have `operator=` for temporary objects as the changes will not propage to collections etc. For instance this a problem with iterators - `*iterator = object` will create a temporary object then rebind it `object` but the changes will not propagate to the `iterator` or collection from which it was obtained.

I propose here to `delete` the copy-assignment for temporary datatype objects since  it's error prone and users may get confused by the semantics.
This also makes it harder to abuse the iterators and treat them as if there were fulfilling "LegacyOutputIterator" (which they don't) - code like  `*iterator = object` will not compile instead of compiling but having wrong semantics. I think this should cause any std algorithm expecting a "LegacyOutputIterator" to not compile with collection iterators (since now the relevant operations are not defined instead of being defined but wrong)

(`std::output_iterator` concept already fails even without this)

Closes #281

I hope this is the last chapter of the podio collection iterators saga :slightly_smiling_face:


m-fila:
> Just to confirm that I didn't miss anything: Interfaces do not need this because they don't have collections, right?

Yeah, I don't think we ever had any problem like these with the interfaces. But disabling assigning to temporary makes little sense for them too. And it shouldn't hurt. So maybe we want to do this just in case or for consistency with other types

tmadlener (approved):
Thanks. This looks good. Giving others a chance to comment. I will merge this either tomorrow or Friday, unless there are any new developments or comments.

wdconinc:
> Pinging @wdconinc as he was also there from the beginning and I would like to confirm that this is also going in the right direction for EIC.

Looks good to me too.

PR #745, opened 2025-03-05
Don't rely on ROOT JIT to define python build_version

tmadlener:
BEGINRELEASENOTES
- Fix the (potential) import error for `podio.version.build_version` by not relying on ROOTs JIT, but rather parsing the python `__version__` into the correct type.

ENDRELEASENOTES

Fixes https://github.com/key4hep/key4hep-spack/issues/670

### DIFF
--- a/python/podio/version.py
+++ b/python/podio/version.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 """Module that facilitates working with the podio::version::Version"""
 
-from podio import __version__
+from podio import __version__  # pylint: disable=wrong-import-order
 
 import ROOT
 
--- a/python/podio/version.py
+++ b/python/podio/version.py
@@ -1,6 +1,8 @@
 #!/usr/bin/env python3
 """Module that facilitates working with the podio::version::Version"""
 
+from podio import __version__
+
 import ROOT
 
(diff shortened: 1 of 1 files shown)


python/podio/version.py
```diff
@@ -37,5 +39,7 @@ def parse(*args):
     return ver
 
 
-# The version with which podio has been built. Same as __version__
-build_version = podio.version.build_version
+# The version with which podio has been built.
+# Same as defined in C++ as static constexpr podio::version::build_version
+# See: https://github.com/key4hep/key4hep-spack/issues/670
+build_version = __version__
```
tmadlener, line 45:
```suggestion
build_version = parse(__version__)
```

jmcarcell:
If you tried it and it fixes it then I guess it's good, I never knew why this happened and it disappears in the newer versions of ROOT. I even think I wasn't able to reproduce locally with the ROOT version that we have in the stack, it has always worked fine for me (I have been working with debug builds for a long time).

tmadlener:
Things have been working nicely with me locally and it is a 1:1 replacement of the same thing, but now we construct it in python, rather than relying on ROOTs JIT.

I have been running with the other version for months now without issue, but yesterday it started failing even though the ROOT version is unchanged. Not sure if it's an issue in how podio builds and links things, or if it's something else. Anyhow with this it doesn't matter.

PR #747, opened 2025-03-07
Fix handling of forced python version.

scott-snyder:
We run into problems if we configure with a python version different from that found by the default find_dependency call for Python.

We generate a cmake fragment with the lines

```
if(@REQUIRE_PYTHON_VERSION@)
  find_dependency(Python @REQUIRE_PYTHON_VERSION@ COMPONENTS Interpreter)
else()
  find_dependency(Python COMPONENTS Interpreter)
endif()
```

The if condition expands to something like

```
if(3.12.9)
```

cmake tries to interpret the argument here as a floating-point number, but fails to convert it, and (silently!) treats the condition as false. Need to do this test explictly as a string comparison instead.

BEGINRELEASENOTES
- Fix a build issue when the python version used is different than the one found by default by cmake.
ENDRELEASENOTES

### DIFF
--- a/cmake/podioConfig.cmake.in
+++ b/cmake/podioConfig.cmake.in
@@ -25,7 +25,7 @@ set(PODIO_IO_HANDLERS "@PODIO_IO_HANDLERS@")
 
 include(CMakeFindDependencyMacro)
 find_dependency(ROOT @ROOT_VERSION@)
-if(@REQUIRE_PYTHON_VERSION@)
+if(NOT "@REQUIRE_PYTHON_VERSION@" STREQUAL "")
   find_dependency(Python @REQUIRE_PYTHON_VERSION@ COMPONENTS Interpreter)
 else()
   find_dependency(Python COMPONENTS Interpreter)


tmadlener (approved):
The fix looks sensible. I am fairly certain you are well aware of this, but just to be sure: We have this explicit version check because we need to ensure that we get the same python version that was also used for building ROOT. Otherwise python bindings will probably not work.

We have run into issues with this in the past, where podio (or DD4hep where a similar clause is present) was requested from other CMake packages as dependency. In that case if an other package discovered a newer python version (e.g. from the underlying system) cmake will cache this and always return that in `find_package`. Depending on the details of the `find_package` call, this fails due to incompatible versions or potentially also other requirements (e.g. the system not having the `-dev` parts installed). In those cases it was necessary to either
- Change the order of how the dependencies are found by re-arranging the `find_package` calls
- Or explicitly make cmake find the desired python, e.g. by passing `-DPython_EXECUTABLE=$(which python)`

scott-snyder:
Yeah, that's what i was running into.  I've been building on a Fedora system,
where sometimes versions of python are installed that are newer than the
default.  I build ROOT to correspond to the python version given by
/usr/bin/python, but sometimes cmake will find a newer version.
The cmake code here is then supposed to choose the version that ROOT
was built with, but due to the issue i mentioned, it didn't actually work.

PR #748, opened 2025-03-10
Add typeName member to link collections

### LINKED ISSUE #731: Link collections do not have a `typeName` member like other collections
And therefore it is impossible to use `coll::typeName` as with the rest of the collections (see https://github.com/AIDASoft/podio/blob/master/python/templates/Collection.h.jinja2#L65). This was uncovered in https://github.com/key4hep/k4FWCore/pull/276 (and fixed in https://github.com/key4hep/k4FWCore/pull/277).

There exists a general function to obtain the type name for a Link collection: https://github.com/AIDASoft/podio/blob/master/include/podio/detail/LinkFwd.h#L21 but implementing this as constexpr to add it generally to all collections seems to be non-trivial. At the very least, `constexpr` string concatenation would have to be implemented, and I'm not sure if this is going to work in C++20.

Another option is to implement this in the template, making the actual collection type have this attribute by inheriting from `podio::LinkCollection<From, To>` and being slightly different from `podio::LinkCollection<From, To>`. This would be very easy to do in https://github.com/AIDASoft/podio/blob/master/python/templates/LinkCollection.h.jinja2#L14 because all the information is known at this time. I have a working implementation with this method.

And the last option is simply not to add this since there are not a lot of use cases.

m-fila:
BEGINRELEASENOTES
- Change type of `typeName` members to `std::string_view`. Add utility to concatenate `std::string_view`s at compile time.
- Add `typeName` static member to link collections. `typeName` can be checked at compilation time.
- Deprecate `linkCollTypeName` and `linkTypeName` helper functions.

ENDRELEASENOTES

Adding utility to concatenate multiple `std::string_view`s at compile time. Then using the utility to add missing `typeName` member for link collections.

The `linkCollTypeName` and `linkTypeName` functions will be now redundant so I propose to deprecate them. The only usage of them I could find is in EDM4hep in schema evolution

Closes #731

### DIFF
--- a/include/podio/utilities/StaticConcatenate.h
+++ b/include/podio/utilities/StaticConcatenate.h
@@ -10,14 +10,15 @@ namespace podio::utils {
 template <const std::string_view&... strs>
 struct static_concatenate {
   static constexpr auto init_arr() {
-    constexpr auto total_size = (strs.size() + ... + 0);
+    constexpr auto total_size = (strs.size() + ... + 1); // reserve space for '\0'
     auto arr = std::array<char, total_size>();
     auto it = arr.begin();
     ((it = std::ranges::copy(strs, it).out), ...);
+    arr.back() = '\0';
     return arr;
   }
   constexpr static auto array = init_arr();
-  constexpr static auto value = std::string_view(array.data(), array.size());
+  constexpr static auto value = std::string_view(array.data(), array.size() - 1); // skip '\0'
 };
 
 /// Variable template for concatenating a set of string_views into a single string_view at compile time
--- a/include/podio/utilities/StaticConcatenate.h
+++ b/include/podio/utilities/StaticConcatenate.h
@@ -1,0 +1,28 @@
+#ifndef PODIO_UTILITIES_STATICCONCATENATE_H
+#define PODIO_UTILITIES_STATICCONCATENATE_H
+
+#include <algorithm>
+#include <array>
+#include <string_view>
+namespace podio::utils {
+
+/// Helper struct to concatenate a set of string_views into a single string_view at compile time
+template <const std::string_view&... strs>
+struct static_concatenate {
+  static constexpr auto init_arr() {
+    constexpr auto total_size = (strs.size() + ... + 0);
+    auto arr = std::array<char, total_size>();
+    auto it = arr.begin();
+    ((it = std::ranges::copy(strs, it).out), ...);
+    return arr;
+  }
+  constexpr static auto array = init_arr();
+  constexpr static auto value = std::string_view(array.data(), array.size());
+};
+
+/// Variable template for concatenating a set of string_views into a single string_view at compile time
+template <const std::string_view&... strs>
+inline constexpr std::string_view static_concatenate_v = static_concatenate<strs...>::value;
+
+} // namespace podio::utils
+#endif // PODIO_UTILITIES_STATICCONCATENATE_H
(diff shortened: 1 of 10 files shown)


include/podio/utilities/StaticConcatenate.h
```diff
@@ -0,0 +1,29 @@
...
+    constexpr auto total_size = first.size() + (rest.size() + ... + 0);
+    auto arr = std::array<char, total_size>();
+    auto it = arr.begin();
+    it = std::ranges::copy(first, it).out;
+    ((it = std::ranges::copy(rest, it).out), ...);
+    return arr;
+  }
+  constexpr static auto array = init_arr();
+  constexpr static auto value = std::string_view(array.data(), array.size());
+};
```
m-fila, line 22: I didn't notice, `first` template argument here could be removed

include/podio/utilities/StaticConcatenate.h
```diff
@@ -0,0 +1,28 @@
...
+#include <array>
+#include <string_view>
+namespace podio::utils {
+
+/// Helper struct to concatenate a set of string_views into a single string_view at compile time
+template <const std::string_view&... strs>
+struct static_concatenate {
+  static constexpr auto init_arr() {
+    constexpr auto total_size = (strs.size() + ... + 0);
+    auto arr = std::array<char, total_size>();
```
tmadlener, line 14: This will result in a non null-terminated array of `char`, right?

Maybe 
```suggestion
    constexpr auto total_size = (strs.size() + ... + 1);
    auto arr = std::array<char, total_size>();
```
and then explicitly adding a `\0` as the last element of `arr` could be a good idea, just in case anything that needs to consume this (e.g. via `.data()`) expects a null-terminated string.

include/podio/utilities/StaticConcatenate.h
```diff
@@ -0,0 +1,28 @@
+#ifndef PODIO_UTILITIES_STATICCONCATENATE_H
+#define PODIO_UTILITIES_STATICCONCATENATE_H
+
+#include <algorithm>
+#include <array>
+#include <string_view>
+namespace podio::utils {
+
+/// Helper struct to concatenate a set of string_views into a single string_view at compile time
+template <const std::string_view&... strs>
```
tmadlener, line 10: Do we have to pass them as `const&` here? It is very nit-picky (and it really doesn't matter here, I think), but isn't the general advice to pass `string_view` by value?

tmadlener (commented):
Very nice. Effectively I only have one real comment.

Regarding the deprecation / removal of the `linkCollTypeName` and `linkTypeName` functions, I think it's fine to remove them, they are in a `detail` namespace, and once this is merged we can get rid of the usage in EDM4hep and remove them here as well.

include/podio/utilities/StaticConcatenate.h
m-fila: Unfortunately yes, otherwise it wouldn't be a "structural type" (https://en.cppreference.com/w/cpp/language/template_parameters) and valid type to use as a non-type template argument :/

include/podio/utilities/StaticConcatenate.h
m-fila: Good idea

tmadlener:
As discussed during the meeting today. Merging this and then the corresponding PR in EDM4hep so that nightlies don't break.

PR #757, opened 2025-03-25
Keep RNTuple API compatibility down to ROOT v6.34

tmadlener:
BEGINRELEASENOTES
- Make sure that podio RNTuple readers and writers work with ROOT 6.34 and later versions.
 - **Bump the minimum required version of ROOT for RNTuple support to 6.34**. That is where the file format has been stabilized, but the API has only moved out of the `Experimental` namespace in 6.36.

ENDRELEASENOTES

See https://github.com/root-project/root/pull/17804


tmadlener:
Keeping this open for now to see if more things are moved out of the `Experimenatl` namespace on the ROOT side. Merging before the next tag in any case.

tmadlener:
More stabilization in:
- https://github.com/root-project/root/pull/18175
- https://github.com/root-project/root/pull/18080

tmadlener:
Coming up next:
- https://github.com/root-project/root/pull/18203
- https://github.com/root-project/root/pull/18225

tmadlener:
It looks like these should be all of the typenames from RNTuple that we use, so I don't expect more changes from this at the moment and this is ready for review.

include/podio/RNTupleWriter.h
```diff
@@ -10,11 +10,20 @@
...
 #include <unordered_map>
 #include <vector>
 
+#if ROOT_VERSION_CODE < ROOT_VERSION(6, 35, 0)
+namespace ROOT {
+using REntry = ROOT::Experimental::REntry;
+using RNTupleModel = ROOT::Experimental::RNTupleModel;
+using RNTupleWriter = ROOT::Experimental::RNTupleWriter;
+} // namespace ROOT
+#endif
```
tmadlener, line 25: The file format has been finalized in 6.34. Need a cleaner solution here to not pollute the ROOT namespace.

CMakeLists.txt
```diff
@@ -81,7 +81,7 @@ set(root_components_needed RIO Tree)
 set(root_min_version 6.28.04)
 if(ENABLE_RNTUPLE)
   list(APPEND root_components_needed ROOTNTuple)
-  set(root_min_version 6.32)
+  set(root_min_version 6.34)
```
tmadlener, line 84: This is technically not necessary, as things still build with 6.32. However, the files that are produced have no guarantees wrt backwards or forwards compatibility (from ROOT). So the question is whether we want to be on the safe side here or not. Bumping this has quite some implications for our CI, as many workflows still run on older versions of ROOT.

.github/workflows/key4hep.yml
```diff
@@ -34,7 +34,7 @@ jobs:
             -DCMAKE_CXX_STANDARD=20 \
             -DCMAKE_CXX_FLAGS=" -fdiagnostics-color=always -Werror -Wno-error=deprecated-declarations " \
             -DUSE_EXTERNAL_CATCH2=AUTO \
-            -DENABLE_RNTUPLE=ON \
+            -DENABLE_RNTUPLE=OFF \
```
tmadlener, line 37: As soon as we have a recent enough ROOT version in Key4hep this should be switched on again.

.github/workflows/pre-commit.yml
```diff
@@ -35,7 +35,7 @@ jobs:
           cd build
           cmake .. -DENABLE_SIO=ON \
             -DENABLE_JULIA=ON \
-            -DENABLE_RNTUPLE=ON \
+            -DENABLE_RNTUPLE=OFF \
```
tmadlener, line 38: As soon as we have root 6.34 in Key4hep, this should be switched on again.

jmcarcell:
Hmm after this the version of ROOT in the stack needs to be updated or the nightlies don't build; I think I'll add 6.34 temporarily and then switch to 6.36 once it's ready

tmadlener:
We can also put it back down to 6.32 to keep the stack building and remove it later. It's only the cmake check that needs to change in this case.

PR #758, opened 2025-03-27
Formalize common interface for collections with a concept

m-fila:
BEGINRELEASENOTES
- Formalize expected collection interface with a concept
- Add `at` and `create` to `UserDataCollection`

ENDRELEASENOTES

This is an idea to formalize to some degree the expected interface of a podio collection.
Currently we have three types of collections: `UserDataCollection`, link collection and datatype collection. Each of them implements some common methods and type aliases that aren't and can't be part of abstract class `CollectionBase`. 
Yet from time to time we discover that one of them is missing something that the others have or worse we discover that some generic code doesn't work with some collection (for example #731, #718, #598)

The concept can be at first used in tests to ensure at compile time that none of the collections is missing part of expected interface.
Secondly it could be possibly used in derived packages to constrain templates better than `std::is_base_of<podio::CollectionBase, T>`

I propose to include the following:
- being a concrete class derived from `CollectionBase` (which also handles that all the interfaces from `CollectionBase` are there)
- constructors, assignments,
- implementing ranges
- constexpr static strings with human readable types (`typeName`), after #748
- typedefs,
- `create()`, `push_back(obj)`,
- all combinations of `begin()` and `end()`,
- element access with `operator[]` and `at`

Perhaps something else worth having in all collections is still missing?
Datatype collections have LCIO-style `operator->` but I'm not sure if the others also should have it?

### DIFF
--- a/include/podio/utilities/TypeHelpers.h
+++ b/include/podio/utilities/TypeHelpers.h
@@ -280,7 +280,7 @@ concept CollectionType = requires(T t, const T ct) {
   typename T::reverse_iterator;
   requires std::random_access_iterator<typename T::reverse_iterator>;
   // member functions
-  { t.create() } -> std::convertible_to<typename T::mutable_type>;
+  requires std::same_as<std::remove_cvref_t<decltype(t.create())>, typename T::mutable_type>;
   { t.push_back(std::declval<std::add_lvalue_reference_t<std::add_const_t<typename T::mutable_type>>>()) };
   { t.push_back(std::declval<std::add_lvalue_reference_t<std::add_const_t<typename T::value_type>>>()) };
   { t.begin() } -> std::same_as<typename T::iterator>;
@@ -295,10 +295,14 @@ concept CollectionType = requires(T t, const T ct) {
   { t.rend() } -> std::same_as<typename T::reverse_iterator>;
   { t.crend() } -> std::same_as<typename T::const_reverse_iterator>;
   { ct.rend() } -> std::same_as<typename T::const_reverse_iterator>;
-  { t[std::declval<typename T::size_type>()] } -> std::convertible_to<typename T::mutable_type>;
-  { ct[std::declval<typename T::size_type>()] } -> std::convertible_to<typename T::value_type>;
-  { t.at(std::declval<typename T::size_type>()) } -> std::convertible_to<typename T::mutable_type>;
-  { ct.at(std::declval<typename T::size_type>()) } -> std::convertible_to<typename T::value_type>;
+  requires std::same_as<std::remove_cvref_t<decltype(t[std::declval<typename T::size_type>()])>,
+                        typename T::mutable_type>;
+  requires std::same_as<std::remove_cvref_t<decltype(ct[std::declval<typename T::size_type>()])>,
+                        typename T::value_type>;
+  requires std::same_as<std::remove_cvref_t<decltype(t.at(std::declval<typename T::size_type>()))>,
+                        typename T::mutable_type>;
+  requires std::same_as<std::remove_cvref_t<decltype(ct.at(std::declval<typename T::size_type>()))>,
+                        typename T::value_type>;
 };
 
 namespace utils {
--- a/include/podio/utilities/TypeHelpers.h
+++ b/include/podio/utilities/TypeHelpers.h
@@ -254,7 +254,7 @@ concept CollectionType = requires(T t, const T ct) {
   requires std::default_initializable<T>;
   requires std::destructible<T>;
   requires std::movable<T>;
-  requires std::move_constructible<T>;
+  requires !std::copyable<T>;
   requires std::ranges::random_access_range<T>;
   // typeName's
   { T::typeName } -> std::convertible_to<std::string_view>;
--- a/include/podio/utilities/TypeHelpers.h
+++ b/include/podio/utilities/TypeHelpers.h
@@ -247,63 +247,59 @@ class CollectionBase;
 
 /// Concept for checking whether a passed type T is a collection
 template <typename T>
-concept CollectionType = requires(T t, const T ct) {
-  // general concepts
-  requires !std::is_abstract_v<T>;
-  requires std::derived_from<T, CollectionBase>;
-  requires std::default_initializable<T>;
-  requires std::destructible<T>;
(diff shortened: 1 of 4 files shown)


include/podio/utilities/TypeHelpers.h
```diff
@@ -244,8 +247,59 @@ class CollectionBase;
...
-concept CollectionType = std::is_base_of_v<CollectionBase, T>;
-
+concept CollectionType = requires(T t, const T ct) {
+  // general concepts
+  requires !std::is_abstract_v<T>;
+  requires std::derived_from<T, CollectionBase>;
+  requires std::default_initializable<T>;
+  requires std::destructible<T>;
+  requires std::movable<T>;
+  requires std::move_constructible<T>;
```
tmadlener, line 257: Should we also explicitly add that they are not copy-able?

include/podio/utilities/TypeHelpers.h
```diff
@@ -244,8 +247,59 @@ class CollectionBase;
...
+  typename T::const_iterator;
+  requires std::random_access_iterator<typename T::const_iterator>;
+  typename T::iterator;
+  requires std::random_access_iterator<typename T::iterator>;
+  typename T::const_reverse_iterator;
+  requires std::random_access_iterator<typename T::const_reverse_iterator>;
+  typename T::reverse_iterator;
+  requires std::random_access_iterator<typename T::reverse_iterator>;
+  // member functions
+  { t.create() } -> std::convertible_to<typename T::mutable_type>;
```
tmadlener, line 283: Do we want `convertible_to` or `same_as` here? After all, we want the exact type here, not something that is convertible. Similar below for the indexed access.

m-fila: `UserDataCollection` is returning references or const references which would have to be treated with `std::remove_cvref_t` to be `std::same_as` collection `value_type` or `mutable_type`. I think I actually prefer to do it this way rather than check if it's convertible.

There is possibly another thing. We could also check that returned value is still "linked" to the container - either it must be a reference (as in `UserDataCollection`) or proxy (as in link and datatype collections).  That would probably require adding some more helpers and also there is no built-in concept for proxy reference

include/podio/utilities/TypeHelpers.h
m-fila: I wasn't sure on that one.  At first  it isn't necessary for the interface to be not copy-able, but then making collections would be probably quite challenging with all the internals.
I think we could at this anyway as at this point having a copyable collection would be by mistake
btw I checked that `std::move_constructible<T>` is redundant with `std::movable<T>`

include/podio/utilities/TypeHelpers.h
```diff
@@ -244,8 +247,63 @@ class CollectionBase;
...
-
+concept CollectionType = requires(T t, const T ct) {
+  // general concepts
+  requires !std::is_abstract_v<T>;
+  requires std::derived_from<T, CollectionBase>;
+  requires std::default_initializable<T>;
+  requires std::destructible<T>;
+  requires std::movable<T>;
+  requires !std::copyable<T>;
+  requires std::ranges::random_access_range<T>;
```
jmcarcell, line 258:
```suggestion
concept CollectionType = 
  !std::is_abstract_v<T> &&
  std::derived_from<T, CollectionBase> &&
  std::default_initializable<T> &&
  std::destructible<T> &&
  std::movable<T> &&
  !std::copyable<T> &&
  std::ranges::random_access_range<T> &&
  requires(T t, const T ct) {
```

include/podio/utilities/TypeHelpers.h
```diff
@@ -244,8 +247,63 @@ class CollectionBase;
...
+  { T::typeName } -> std::convertible_to<std::string_view>;
+  { std::bool_constant<(T::typeName, true)>() } -> std::same_as<std::true_type>; // ~is annotated with constexpr
+  { T::valueTypeName } -> std::convertible_to<std::string_view>;
+  { std::bool_constant<(T::valueTypeName, true)>() } -> std::same_as<std::true_type>; // ~is annotated with constexpr
+  { T::dataTypeName } -> std::convertible_to<std::string_view>;
+  { std::bool_constant<(T::dataTypeName, true)>() } -> std::same_as<std::true_type>; // ~is annotated with constexpr
+  // typedefs
+  typename T::value_type;
+  typename T::mutable_type;
+  requires std::convertible_to<typename T::mutable_type, typename T::value_type>;
```
jmcarcell, line 269: I think for this one and the ones below the requirement of the typename to exist doesn't need to be there before since it won't compile when it reaches this `requires`. I guess it's there for explicitness.

include/podio/utilities/TypeHelpers.h
```diff
@@ -244,8 +247,63 @@ class CollectionBase;
...
+  typename T::difference_type;
+  requires std::signed_integral<typename T::difference_type>;
+  typename T::size_type;
+  requires std::unsigned_integral<typename T::size_type>;
+  typename T::const_iterator;
+  requires std::random_access_iterator<typename T::const_iterator>;
+  typename T::iterator;
+  requires std::random_access_iterator<typename T::iterator>;
```
jmcarcell, line 277: At least this one is already implied by `std::ranges::random_access_range`. Are none of the others already implied? If so they could be removed.

include/podio/utilities/TypeHelpers.h
m-fila: Added comments and using `std::remove_cvref_t` and `std::remove_ref_t` instead of `std::convertible_to`

tmadlener (approved):

include/podio/utilities/TypeHelpers.h
tmadlener: I suppose in most of the cases this shouldn't really matter as the majority of the collections are generated with the right interface in any case. I do like the more explicit  message a bit better though, even if the information content / readability is probably the same at the point where you actually have to look at these errors ;)

tmadlener:
Anything speaking against merging this as is?

tmadlener:
I am fine with keeping some redundancy for fewer questions / riddles when reading the concept. If it becomes an issue with compile times we can always prune it down to a minimal set, IMO.

PR #768, opened 2025-04-23
Re-enable RNTuple support for ROOT 6.32

tmadlener:
BEGINRELEASENOTES
- Keep RNTuple support enabled for ROOT 6.32, but emit a warning at CMake stage

ENDRELEASENOTES

See discussion in #757


tmadlener:
FYI @jmcarcell

PR #773, opened 2025-04-25
Restore compatibility with gcc 11

### LINKED ISSUE #765: podio does no longer build with gcc 11
~**We are planning to remove support for building with gcc11 in `v01-03`**. Tests have been failing to build since #626 and with #620 also building without tests has started to fail with gcc11, as c++20 support is not yet fully there.~

~This is related removing c++17 support (see #701).~

~**In case you are still using gcc11 for building podio and have requirements that make moving to a newer version impossible or hard, please let us know** (e.g. by a comment to this issue).~

~This is technically a breaking change since we remove a "feature". However, we have decided against bumping the major version of podio for this for now, as the API and the generated code remains unchanged.~

We originally planned to remove gcc11 support, but have since realized it is rather straight forward to restore it. Hence, we for now continue supporting gcc 11. We will create a new issue / notification once we actually remove it.

tmadlener:
BEGINRELEASENOTES
- Make sure compilation with gcc11 is possible again and add a workflow based on LCG_107 and Ubuntu 22 to check compatibility in CI

ENDRELEASENOTES

Fixes #765 

Thanks to Mistral who suggested the two fixes for the ranges.

- In `podio-dump-tool` it looks like gcc11 is still missing a `string_view` constructor (my interpretation)
- For the std interoperability checks it looks like gcc11 has not yet fixed https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2415r2.html

### DIFF
--- a/tests/unittests/std_interoperability.cpp
+++ b/tests/unittests/std_interoperability.cpp
@@ -1342,7 +1342,17 @@ TEST_CASE("Collection as range", "[collection][ranges][std]") {
   // std::range::common_range
   STATIC_REQUIRE(std::ranges::common_range<CollectionType>);
   // std::range::viewable_range
+  // special handling of gcc 11
+#if defined(__GNUC__)
+  #if __GNUC__ < 12
+  auto view = std::ranges::ref_view(coll);
+  STATIC_REQUIRE(std::ranges::viewable_range<decltype(view)>);
+  #else
   STATIC_REQUIRE(std::ranges::viewable_range<CollectionType>);
+  #endif
+#else
+  STATIC_REQUIRE(std::ranges::viewable_range<CollectionType>);
+#endif
 }
 
 TEST_CASE("Collection and std algorithms", "[collection][iterator][std]") {
@@ -1568,7 +1578,18 @@ TEST_CASE("LinkCollection and range concepts", "[links][ranges][std]") {
   STATIC_REQUIRE(std::ranges::random_access_range<link_collection>);
   STATIC_REQUIRE(std::ranges::sized_range<link_collection>);
   STATIC_REQUIRE(std::ranges::common_range<link_collection>);
+  // Dance around gcc11 issue
+#if defined(__GNUC__)
+  #if __GNUC__ < 12
+  auto coll = link_collection{};
+  auto view = std::ranges::ref_view(coll);
+  STATIC_REQUIRE(std::ranges::viewable_range<decltype(view)>);
+  #else
   STATIC_REQUIRE(std::ranges::viewable_range<link_collection>);
+  #endif
+#else
+  STATIC_REQUIRE(std::ranges::viewable_range<link_collection>);
+#endif
 }
 
 TEST_CASE("RelationRange as range", "[relations][ranges][std]") {
--- a/tests/unittests/std_interoperability.cpp
+++ b/tests/unittests/std_interoperability.cpp
@@ -1313,7 +1313,7 @@ TEST_CASE("Collection and std iterator adaptors", "[collection][container][adapt
   }
 }
 
-TEST_CASE("Collection as range", "[collection][ranges][std]") {
+TEST_CASE("Collection as range", "[collection][ranges][std][GCC11-FAIL]") {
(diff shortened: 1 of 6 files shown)


tests/unittests/std_interoperability.cpp
```diff
@@ -1342,7 +1342,17 @@ TEST_CASE("Collection as range", "[collection][ranges][std]") {
...
+#if defined(__GNUC__)
+  #if __GNUC__ < 12
+  auto view = std::ranges::ref_view(coll);
+  STATIC_REQUIRE(std::ranges::viewable_range<decltype(view)>);
+  #else
   STATIC_REQUIRE(std::ranges::viewable_range<CollectionType>);
+  #endif
+#else
+  STATIC_REQUIRE(std::ranges::viewable_range<CollectionType>);
+#endif
```
m-fila, line 1355:
```suggestion
#if defined(__GNUC__) && __GNUC__ < 12
  DOCUMENTED_STATIC_FAILURE(std::ranges::viewable_range<CollectionType>);
#else
  STATIC_REQUIRE(std::ranges::viewable_range<CollectionType>);
#endif
```


Since this is basically checking whether collections fulfill `viewable_range` concept for sake of documentation, I think I'd prefer to expect them to fail the test for GCC 11 (and add comment in docs) rather than test a workaround

tmadlener: Yeah that would also be a possibility. I went with the workaround for now, because I wanted to have an gcc 11 build in CI. But another possibility would be to ensure that things build in CI and we skip the tests for gcc 11.

tmadlener: The two test cases that are failing are now tagged with `[GCC11-FAIL]` and we inject a filter flag for running the tests when we build with gcc11.

tests/unittests/std_interoperability.cpp
```diff
@@ -1568,6 +1569,7 @@ TEST_CASE("LinkCollection and range concepts", "[links][ranges][std]") {
   STATIC_REQUIRE(std::ranges::random_access_range<link_collection>);
   STATIC_REQUIRE(std::ranges::sized_range<link_collection>);
   STATIC_REQUIRE(std::ranges::common_range<link_collection>);
+  // Dance around gcc11 issue
```
tmadlener, line 1572:
```suggestion
```

tests/unittests/std_interoperability.cpp
```diff
@@ -1313,7 +1313,7 @@ TEST_CASE("Collection and std iterator adaptors", "[collection][container][adapt
   }
 }
 
-TEST_CASE("Collection as range", "[collection][ranges][std]") {
+TEST_CASE("Collection as range", "[collection][ranges][std][GCC11-FAIL]") {
```
m-fila, line 1316:
```suggestion
TEST_CASE("Collection as range", "[collection][ranges][std]") {
```
This isn't needed anymore, right? At least it isn't present in the other affected test

tmadlener: Missed to remove it. Thanks a lot :)

tools/src/podio-dump-tool.cpp
```diff
@@ -70,19 +70,21 @@ std::vector<size_t> parseEventRange(const std::string_view evtRange) {
   };
 
   // Split by ',' and transform into a range of string views
-  auto splitRange = evtRange | rv::split(',') |
-      rv::transform([](auto&& subrange) { return std::string_view(subrange.begin(), subrange.end()); });
+  auto splitRange = evtRange | rv::split(',') | rv::transform([](auto&& subrange) {
+                      return std::string_view(&*subrange.begin(), std::ranges::distance(subrange));
```
jmcarcell, line 74: This is a bit ugly, maybe it would be good to have a preprocessor check with the current and new version to make cleanup easier and then no one has to wonder why it's done like this. Same for below.

tmadlener: I moved it into a `splitString` utility function where there is now a pre-processor differentiation.

Issue #770, opened 2025-04-24
Reading every library in `LD_LIBRARY_PATH` when using Python


jmcarcell:
I have observed that when doing certain things in python, every library found in `LD_LIBRARY_PATH` (and also including `/usr/lib`) will be read. To reproduce (with EDM4hep):
```
strace -e open,openat,creat python -c "import edm4hep; edm4hep.CaloHitContribution();" 2>&1 | wc -l
```
gives ~20k lines and if one inspects the output, it's opening every library. A similar issue has been reported before in ROOT: 
https://github.com/root-project/root/issues/14277. The result is a slower startup time, in particular when using CVMFS it takes much longer since I believe all the libraries have to be downloaded. I can reproduce locally, with the Key4hep stack and also with the LCG stack.

Further debugging hints that it is the `::typeName` symbol that is not being found by Cling, triggering the search. Indeed, by removing `static constexpr` from https://github.com/AIDASoft/podio/blob/aa5927b06f6333511a173c37ab951b2f3e9eec66/python/templates/Object.h.jinja2#L57 and removing the code that uses `typeName` in `LinkCollectionImpl.h` this is fixed.

For EDM4hep, I checked with a for loop for which objects this search is triggered (object and line count of strace):

```
EventHeader 19371
MCParticle 2007
SimTrackerHit 2007
CaloHitContribution 19370
SimCalorimeterHit 2007
RawCalorimeterHit 19370
CalorimeterHit 2007
ParticleID 19370
Cluster 2007
TrackerHit3D 19370
TrackerHitPlane 19370
RawTimeSeries 19370
Track 2007
Vertex 2007
ReconstructedParticle 2007
TimeSeries 19370
RecDqdx 19370
GeneratorEventParameters 19370
GeneratorPdfInfo 19370
RecoMCParticleLink 1814
CaloHitMCParticleLink 1814
ClusterMCParticleLink 1814
TrackMCParticleLink 1814
CaloHitSimCaloHitLink 1814
TrackerHitSimTrackerHitLink 1814
VertexRecoParticleLink 1814
```
It seems those that have a Link do not have this problem? Maybe because of some include?

For some it is working fine (~2k files opened) but for others it is reading every library (~20k files opened on my machine).

the for loop:
```
[...]

PR #775, opened 2025-05-02
Add a workaround for avoiding opening every library in LD_LIBRARY_PATH

jmcarcell:
Workaround for https://github.com/AIDASoft/podio/issues/770.

BEGINRELEASENOTES
- Define `typeName` outside of each object class, in addition to the existing definition inside the class.

ENDRELEASENOTES

The redundant declaration outside of the class is deprecated, so we have to remove the warnings. With this, opening all the libraries doesn't happen for any of the EDM4hep object classes. There is the possibility to add a test but `strace` is not available in all the container images used in CI I think.

### DIFF
--- a/python/templates/Object.h.jinja2
+++ b/python/templates/Object.h.jinja2
@@ -78,4 +78,19 @@ std::ostream& operator<<(std::ostream& o, const {{ class.bare_type }}& value);
 
 {{ macros.std_hash(class) }}
 
+// This is needed to avoid triggering opening every library in LD_LIBRARY_PATH
+// until it's fixed in ROOT. See https://github.com/root-project/root/issues/18489
+// and https://github.com/AIDASoft/podio/issues/770
+#if defined(__clang__)
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-redundant-constexpr-static-def"
+constexpr std::string_view {{ class.full_type }}::typeName;
+#pragma clang diagnostic pop
+#elif defined(__GNUC__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated"
+constexpr std::string_view {{ class.full_type }}::typeName;
+#pragma GCC diagnostic pop
+#endif
+
 #endif
--- a/python/templates/Object.h.jinja2
+++ b/python/templates/Object.h.jinja2
@@ -83,7 +83,9 @@ std::ostream& operator<<(std::ostream& o, const {{ class.bare_type }}& value);
 // and https://github.com/AIDASoft/podio/issues/770
 #if defined(__clang__)
 #pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunknown-warning-option"
 #pragma clang diagnostic ignored "-Wdeprecated-redundant-constexpr-static-def"
+#pragma clang diagnostic ignored "-Wdeprecated"
 constexpr std::string_view {{ class.full_type }}::typeName;
 #pragma clang diagnostic pop
 #elif defined(__GNUC__)


m-fila:
I'm not sure about adding things deprecated in standard older than the one we are targeting

Would making a `typeName` a non-member variable template instead of `constexpr static` member help?

jmcarcell:
> I'm not sure about adding things deprecated in standard older than the one we are targeting

It works fine also with C++23 and by that time it should hopefully be fixed in ROOT and we can get rid of this.

> Would making a `typeName` a non-member variable template instead of `constexpr static` member help?

Templated on the object type? Maybe, but that would require more changes, wouldn't be worse than having a member variable for each object like now? The workaround doesn't change the design because something doesn't work in ROOT.

m-fila:
> It works fine also with C++23 and by that time it should hopefully be fixed in ROOT and we can get rid of this.

I think it isn't removed in C++26 too so they should have plenty of time indeed. I just conceptually don't like adding things that are already deprecated and also require compiler specific fences.

> > Would making a `typeName` a non-member variable template instead of `constexpr static` member help?
> 
> Templated on the object type? Maybe, but that would require more changes, wouldn't be worse than having a member variable for each object like now? The workaround doesn't change the design because something doesn't work in ROOT.

yup, something like `podio::typeName<MCParticleCollection>` vs `MCParticleCollection::typeName`. I agree it'd be more changes (altough I'd call them syntax changes rather than design changes) just to accommodate for ROOT - also I don't know if this would actually help

tmadlener:
So far, *I think* that `podio::typeName<XYZ>` vs `XYZ::typeName` should not be too much of a change. We mostly rely on the run-time behavior in any case in the `getXYZTypeName` etc. member functions. The one thing that needs to be ensured is that at some point there is a (probably static) string literal to which we can actually point the `string_view`.

How much work would it be to actually figure out whether the external template function makes things better with ROOT?

python/templates/Object.h.jinja2
```diff
@@ -78,4 +78,21 @@ std::ostream& operator<<(std::ostream& o, const {{ class.bare_type }}& value);
...
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunknown-warning-option"
+#pragma clang diagnostic ignored "-Wdeprecated-redundant-constexpr-static-def"
+#pragma clang diagnostic ignored "-Wdeprecated"
+constexpr std::string_view {{ class.full_type }}::typeName;
+#pragma clang diagnostic pop
+#elif defined(__GNUC__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated"
+constexpr std::string_view {{ class.full_type }}::typeName;
```
tmadlener, line 94: This and L89 might be double declared, depending on whether clang pretends to be gcc. See e.g. the changes and pre-processor guards in #773

jmcarcell:
I have a version that works with `podio::typename<>` and this issue does not appear. However, specializations are needed for the data type collections, interfaces, link collections and links, on top of other changes in podio and EDM4hep versus this that is a one line change if it wasn't because of the compiler warnings. Then we still have some other members, like `valueTypeName` in the link collections that would also need to be changed for consistency or they could remain members. It's certainly more complex than the current approach.

tmadlener:
> However, specializations are needed for the data type collections, interfaces, link collections and links

Can you expand a bit here? Does this mean that we need to generate the specializations (effectively what we are already doing now)? The link collections should just be able to put them together as they already do now since #748?

jmcarcell:
> > However, specializations are needed for the data type collections, interfaces, link collections and links
> 
> Can you expand a bit here? Does this mean that we need to generate the specializations (effectively what we are already doing now)? The link collections should just be able to put them together as they already do now since #748?

For links we can put them together from the collections that are linked, for link collections there is a specialization. This is what I had, it's dirty but both the podio and EDM4hep tests pass and all the libraries in LD_LIBRARY_PATH are not opened: https://github.com/AIDASoft/podio/pull/775#issuecomment-2858939976

tmadlener:
Do we still want this to go into the next tag? I think it would be a nice improvement. However, that probably also means that we can't really make deep reaching changes here.

Is it easily possible to check whether having the free standing `podio::typeName` template function and using that for populating the static members still solves the loading problem? I.e. something like

```cpp
template<typename T>
constexpr std::string_view typeName();

// some specialization
struct XYZCollection {
  constexpr static std::string_view typeName = typeName<XYZCollection>();
  // all the rest
}
```

I am asking because I do actually like the free function approach quite a lot, but ideally we can make a somewhat smooth transition with some deprecation warning in between.

jmcarcell:
No, that doesn't work as expected since the issue happens when there is that `static constexpr`. So the options are either a change directly to the free function or a transition that would involve the changes in this PR not to have it `static constexpr`. Or the changes in this PR without the free function.

m-fila (approved):
During meeting we decided to go with this version as it's much smaller change than the alternative with templated `podio::typeName`

tmadlener (approved):

jmcarcell:
I can confirm this fixes the issue. With the nightlies from 2025-05-12, I get:
```
$ strace -e open,openat,creat python -c "import edm4hep; edm4hep.CaloHitContribution();" 2>&1 | wc -l
18720
```
With today's nightlies (2025-05-14) I get:
```
$ strace -e open,openat,creat python -c "import edm4hep; edm4hep.CaloHitContribution();" 2>&1 | wc -l
6226
```

PR #776, opened 2025-05-02
Rename NOTICE to LICENSE

hegner:
This rename is to allow GitHub to detect the license chosen. No functional change


tmadlener:
It looks like this needs a small change to not break installing:

https://github.com/AIDASoft/podio/blob/5b7329701c60772bb9939edb5b4866e2672b0ad0/CMakeLists.txt#L178-L180

PR #711, opened 2024-11-20
Store the collection information in a struct instead of a tuple

tmadlener:
BEGINRELEASENOTES
- Store the collection information in a proper `struct` instead of using a `tuple` to facilitate access for non podio based backends (e.g. Julia)
- Harmonize the format of RNTuple and TTree based ROOT backends
  - Both now store the collection information into the `podio_metadata` TTree / Model as a `vector<podio::root_utils::CollectionWriteInfo>`, which contains all the necessary information for reading a collection.
  - The collection ID and the name are part of this struct, so the `CollectionIDTable` is no longer written separately.
- **This is a breaking change in the format if you rely on reading the metadata about the stored collections.**

ENDRELEASENOTES

Fixes #705 

@peremato This addresses the TTree part of the problem. I just realized that RNTuple stores this in several vectors that run in parallel. I suppose it would be easiest for you (and also for us) to have similar layouts for TTree and RNTuple?

- [x] Harmonize the TTree and RNTuple formats.
- [x] #744 included here
- [x] https://github.com/key4hep/EDM4hep/pull/413
- [x] #745 included here
- [x] Documentation update
- [x] Backwards compatibility for RNTuple? given #757, I think we don't have to worry too much about backwards compatibility with RNTuple

PR #777, opened 2025-05-08
Clean up the ROOT readers and writers

jmcarcell:
BEGINRELEASENOTES
Clean up the ROOT readers and writers
  - Add const where possible
  - Use functions from `std::ranges` when possible
  - Do not use `new` and `delete` when it's not needed (`ROOTReader.cc`)
  - Clean up includes
  - Change some indexes to size_t instead of int to avoid implicit conversions (doesn't change anything)
ENDRELEASENOTES

It's possible I have missed some places where const could be added.
This came up after checking the changes in https://github.com/AIDASoft/podio/pull/711 and noticing at least some of the `new` and `delete`s may not be needed.

### DIFF
--- a/src/ROOTReader.cc
+++ b/src/ROOTReader.cc
@@ -187,7 +187,7 @@ void ROOTReader::initCategory(CategoryInfo& catInfo, const std::string& category
   auto* collInfoBranch = root_utils::getBranch(m_metaChain.get(), root_utils::collInfoName(category));
 
   auto collInfo = std::vector<root_utils::CollectionWriteInfo>();
-  auto collInfoPtr = &collInfo;
+  auto* collInfoPtr = &collInfo;
   if (m_fileVersion >= podio::version::Version{1, 2, 99}) {
     collInfoBranch->SetAddress(&collInfoPtr);
     collInfoBranch->GetEntry(0);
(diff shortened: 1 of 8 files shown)


m-fila:
I'm not sure if there is much benefit changing std algorithms to range algorithms, we don't chain them here anyway. I think the rule of the thumb is that algorithms are slightly faster to both compile and execute, although I don't know if this is noticeable here.

src/ROOTReader.cc
```diff
@@ -293,42 +297,41 @@ void ROOTReader::openFiles(const std::vector<std::string>& filenames) {
     versionBranch->GetEntry(0);
   }
   m_fileVersion = versionPtr ? *versionPtr : podio::version::Version{0, 0, 0};
-  delete versionPtr;
```
jmcarcell, line 296: This is the only case where there is a `delete` without a `new. It doesn't seem to leak without the `delete` so it seems removing it it's fine, also I don't understand why it should be deleted since there is no `new`. I think it is fine to delete it because the contents get copied to `m_fileVersion` so they will never be used again, although it is not needed to do so.

tmadlener: This might be a leftover from a time when there was a `auto versionPtr = new podio::version::Version()`, but I am not sure. I think it actually should do that and it only works at the moment because `sizeof(podio::version::Version)` is smaller than that of a pointer. Or ROOT does some allocation for us (and then maybe also cleans up properly?).

Looking at the logic, it should also be possible to simply use a pointer to `m_fileVersion` and pass that into `versionBranch->SetAddress`, since if the branch is present, then also reading should work and we don't actually need the intermediate pointer to store this into. `m_fileVersion` is initialized to `0,0,0` in any case.

tmadlener (commented):
I don't have a strong opinion on whether to use ranges algorithms or not. In this case it's a single occurrence, so it shouldn't matter too much, I suppose.

tmadlener (approved):

src/ROOTReader.cc
jmcarcell: Yes, I thought the same when I was going through this part and this change was merged in this PR.

PR #780, opened 2025-05-13
Read the collection information from structs starting with version 1.2.999

jmcarcell:
Since we have had version 1.2.99 for a few months, the changes in https://github.com/AIDASoft/podio/pull/711 make the files that have been produced since then incompatible. With this change, and since there is going to be a tag very soon, files produced with 1.2.99 until https://github.com/AIDASoft/podio/pull/711 was merged will be readable in the future. The downside is files produced since https://github.com/AIDASoft/podio/pull/711 was merged are now incompatible, but few days have passed since and I think it is less annoying than all files produced with 1.2.99 before #711 being not readable.

BEGINRELEASENOTES
- Read the collection information starting with version 1.2.999 to maintain compatibility with the files produced with podio 1.2.99
- Bump the version to 1.2.999

ENDRELEASENOTES


tmadlener (commented):
I don't think we should do this, as `1.2.99` is not an existing tagged and released version of podio.  Hence, the only files that could have been written with this problem are from nightly builds. We (or at least I have in the past) used a patch version of 99 to denote "development", i.e. non-released versions of podio. I am not sure whether we want to set a precedent here to mimic some sort of "long term" support for nightly builds (or files written with them).

Where did the breaking files appear in your case?

jmcarcell:
I don't see how not breaking files is not better than breaking them. If #711 + this PR was merged and tagged afterwards, no files would break and the cost is just starting a check on a version that will exist instead of a version that will not. When there is a good reason it's OK, but here there is a choice. Basically every file that I have created for the past 2 or 3 months is broken since I always use the master branch of podio.

tmadlener (commented):
That makes sense. I don't like touching the minor version outside of tagging things, though. Looking at the implementation we should be able to do something like

```cpp
  const auto structMetaData =
      (collInfoBranch->GetClassName() == std::string_view("vector<podio::root_utils::CollectionWriteInfo>"));
```
and then use that instead of the explicit version check.

This should work for all files regardless of when they have been written. (Also this is probably what should have been done for all other changes to the metadata as well, rather than relying on explicit version checks, but I think it's too late now and we don't touch previous versions).

tmadlener:
We could also bump the patch version to `999` and adjust the version checks to that. I think that would have a similar effect as changing to `1.3.0` but without changing minor version outside of the tagging loop.

PR #782, opened 2025-05-20
Make it possible to get the collection for an object

tmadlener:
BEGINRELEASENOTES
- Add `Frame::get` overloads that take a podio generated object handle as argument and return the collection this object belongs to *if the collection is known to the Frame*.

ENDRELEASENOTES

See the discussion in https://github.com/key4hep/k4FWCore/issues/311 for a rationale of this functionality

The main question at this point is whether the existing `getName` functionality that takes a `collectionID` should be deprecated and removed eventually.

### DIFF
--- a/include/podio/Frame.h
+++ b/include/podio/Frame.h
@@ -208,7 +208,7 @@ public:
   /// @returns       A const reference to the collection if it is available or to
   ///                an empty (static) collection
   template <CollectionType CollT>
-  const CollT& get(const CollT::value_type& object) const;
+  const CollT& get(const typename CollT::value_type& object) const;
 
   /// Get the collection pointer to which the passed object belongs from the
   /// Frame.
@@ -406,7 +406,7 @@ const CollT& Frame::get(const std::string& name) const {
 }
 
 template <CollectionType CollT>
-const CollT& Frame::get(const CollT::value_type& object) const {
+const CollT& Frame::get(const typename CollT::value_type& object) const {
   const auto name = m_self->getIDTable().name(object.id().collectionID);
   return get<CollT>(name.value_or(""));
 }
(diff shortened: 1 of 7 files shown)


include/podio/Frame.h
```diff
@@ -199,6 +199,29 @@ class Frame {
...
+  /// Get the collection to which the passed object belongs from the Frame.
+  ///
+  /// @tparam CollT  The type of the desired collection
+  /// @param  object The object for which the collection it belongs to should
+  ///                be retrieved
+  ///
+  /// @returns       A const reference to the collection if it is available or to
+  ///                an empty (static) collection
+  template <CollectionType CollT>
+  const CollT& get(const CollT::value_type& object) const;
```
m-fila, line 211:
```suggestion
  template <CollectionType CollT>
  const CollT& get(const typename CollT::value_type& object) const;
```
Adding a typename here (and in definition below) shouldn't hurt (although we already `value_type` to be type in the concept)

include/podio/Frame.h
```diff
@@ -199,6 +199,29 @@ class Frame {
...
+  /// Frame.
+  ///
+  /// @tparam O      The type of the object
+  /// @param  object The object for which the collection it belongs to should
+  ///                be retrieved
+  ///
+  /// @returns A const pointer to a collection if it is available or a nullptr
+  ///          if it is not
+  template <ObjectType O>
+  inline const podio::CollectionBase* get(const O& object) const;
```
m-fila, line 223: Do we want to have this for compatibility with other `get`s in frame? Otherwise, I think the objects already have typedef to their collection so in principle we could use only the overload returning concrete collection

include/podio/Frame.h
tmadlener: Interesting that this seems to work without `typename` for gcc11, but not for gcc13. Fixed.

PR #783, opened 2025-05-26
Use size_t for the argument of operator[] for link collections

jmcarcell:
BEGINRELEASENOTES
- Use size_t instead of unsigned int for `operator[]`, `at`, `getID` and `setID` for Link collections, since it is the type used for data type collections and UserDataCollections.

ENDRELEASENOTES

### DIFF
--- a/include/podio/detail/LinkCollectionImpl.h
+++ b/include/podio/detail/LinkCollectionImpl.h
@@ -85,11 +85,11 @@ public:
   }
 
   /// Returns the immutable object of given index
-  value_type operator[](unsigned int index) const {
+  value_type operator[](size_t index) const {
     return value_type(m_storage.entries[index]);
   }
   /// Returns the mutable object of given index
-  mutable_type operator[](unsigned int index) {
+  mutable_type operator[](size_t index) {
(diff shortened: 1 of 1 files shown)


m-fila (commented):
Nice one. Can we also update `at` the same way?

m-fila (approved):

tmadlener (commented):
Doesn't this run slightly counter to what we did in https://github.com/AIDASoft/podio/pull/752 ? Links still get an ObjectID, so we will never be able to index to `numeric_limits<size_t>::max` in any case.

tmadlener (commented):
Good point. Just for consistency with the generated collections, can we make it `std::size_t`?

https://github.com/AIDASoft/podio/blob/26aea63b9be373b506b8467ad770db3b0fd61428/python/templates/Collection.h.jinja2#L112-L118

tmadlener:
Another data point: For the generated models we explicitly put `std::` before all fixed width types:
https://github.com/AIDASoft/podio/blob/26aea63b9be373b506b8467ad770db3b0fd61428/python/podio_gen/generator_utils.py#L280-L285

PR #785, opened 2025-05-30
Add missing override to the destructor of UserDataCollection and LinkCollectionImpl

jmcarcell:
BEGINRELEASENOTES
- Add missing override to the destructor of UserDataCollection and LinkCollectionImpl

ENDRELEASENOTES

It has been appearing everywhere from clangd for a long time

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -27,8 +27,6 @@ Checks: -*,
 WarningsAsErrors: ''
 HeaderFilterRegex: ''
 CheckOptions:
-  - key:             modernize-use-override.IgnoreDestructors
-    value:           '1'
   - key:             cppcoreguidelines-non-private-member-variables-in-classes.IgnoreClassesWithAllMemberVariablesBeingPublic
     value:           '1'
   - key:             readability-braces-around-statements.ShortStatementLines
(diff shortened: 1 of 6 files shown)


m-fila:
Strangely, my local clangd is completely happy not having any specifier.
Should we also mark the destructor in `LinkCollection` ?

m-fila (approved):
Maybe I'm missing something but couldn't find any other classes with missing override on a destructor

tmadlener:
> Perhaps we should also ran clang-tidy over example generated classes?

We could in principle, but we do not currently, see e.g.

https://github.com/AIDASoft/podio/blob/823c0e6e6bf6ce2fe498f0f27992b8029fa95cb8/cmake/podioMacros.cmake#L377-L378
https://github.com/AIDASoft/podio/blob/823c0e6e6bf6ce2fe498f0f27992b8029fa95cb8/cmake/podioMacros.cmake#L103-L105
https://github.com/AIDASoft/podio/blob/823c0e6e6bf6ce2fe498f0f27992b8029fa95cb8/cmake/podioMacros.cmake#L274-L278

If we run over the generated sources, we need to still switch it off for downstream packages like EDM4hep, because they might have a different clang-tidy configuration than we have, so it's not entirely trivial to enforce something here from the podio perspective.

We could try to get to a base set of clang-tidy checks that we want to run and only use that, but there I am not entirely sure how easy it is to integrate that into CMake (i.e. having different clang-tidy configs for different parts of the build tree without mixing them).

jmcarcell:
CI is fixed now

PR #786, opened 2025-06-02
Introduce a FrameDataType concept

tmadlener:
Make the interface of the EmptyFrameData into a concept and replace the unconstrained constructors with it.

BEGINRELEASENOTES
- Introduce a `FrameDataType` concept and constrain the `Frame` constructors with it

ENDRELEASENOTES

### DIFF
--- a/include/podio/Frame.h
+++ b/include/podio/Frame.h
@@ -31,11 +31,18 @@ concept RValueType = !std::is_lvalue_reference_v<T>;
 template <typename T>
 concept CollectionRValueType = CollectionType<T> && RValueType<T>;
 
+/// Concept encoding the minimal interface a type has to provide to usable as
+/// raw data from which a Frame can be constructed.
+///
+/// Since the Frame either locks on raw data access, or is guaranteed to only
+/// run on a single thread when doing so (e.g. constructors) none of these have
+/// to be thread-safe either, even though implementations might be able to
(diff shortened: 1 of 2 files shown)


include/podio/Frame.h
```diff
@@ -22,13 +23,31 @@
...
+/// raw data from which a Frame can be constructed.
+///
+/// Since the Frame either locks on raw data access, or is guaranteed to only
+/// run on a single thread when doing so (e.g. constructors) none of these have
+/// to be thread-safe either, even though implementations might be able to
+/// provide reasonable guarantees anyway.
 template <typename T>
-concept RValueType = !std::is_lvalue_reference_v<T>;
+concept FrameDataType = requires(T t) {
+  { t.getIDTable() } -> std::same_as<podio::CollectionIDTable>;
```
jmcarcell, line 43: Maybe it should be considered the case when the types can be converted to that one (`std::convertible_to` instead of `std::same_as`). I don't have a use case but it may be restrictive. Although there are not that many different types that can convert to the types below...

tmadlener: I think I would leave it as it is, also because I am not entirely sure the `convertible_to` actually applies in the current implementation since we do

https://github.com/AIDASoft/podio/blob/36f43c49a65734586779d7557f65a2dc05bce331/include/podio/Frame.h#L416

PR #788, opened 2025-06-11
Add ROOT_INCLUDE_PATH from the environment to the tests

jmcarcell:
This is a fix for the workflows that build on Alma 9 and Ubuntu 24.04 (key4hep workflow). After switching `runtime_cxxmodules` ON in ROOT, as it is the default, there is an issue in ROOT when building with `Vc` standalone: https://github.com/root-project/root/issues/18949. The workaround is to have the path to `<Vc>/include` in `ROOT_INCLUDE_PATH`, and this is done in the stack. However, podio doesn't use this causing now an error in the CI tests. Example error messages:

```
2025-06-11T07:34:27.3475585Z <<< cling interactive line includer >>>: fatal error: module file '/tmp/root/spack-stage/spack-stage-root-6.36.00-hjvchjeqsujyxmw5skcpbbzc5kk65hdv/spack-build-hjvchje/lib/Vc.pcm' not found: module file not found
2025-06-11T07:34:27.3477929Z <<< cling interactive line includer >>>: note: imported by module 'MathCore' in '/cvmfs/sw-nightlies.hsf.org/key4hep/releases/2025-06-07/x86_64-almalinux9-gcc14.2.0-opt/root/6.36.00-hjvchj/lib/root/MathCore.pcm'
```

Why does it work for the LCG stacks? Because there, they also set `CPLUS_INCLUDE_PATH`, which bypasses the settings that we have in the tests. We could also set this in the Key4hep stack but given that this is a global setting and only podio seems to have issues with `ROOT_INCLUDE_PATH`, I would prefer having the workaround in podio. In addition, in other stacks with the configuration above and without `CPLUS_INCLUDE_PATH` the tests would fail. One possible downside would be that when removing something it would still be findable from the installation in the stack, so I added a call to `k4_local_repo` to make sure this doesn't happen.

BEGINRELEASENOTES
- Add the value of `ROOT_INCLUDE_PATH` from the environment to the tests
- Call `k4_local_repo` in the key4hep workflow

ENDRELEASENOTES


tmadlener:
pre-commit will be fixed with #789

PR #791, opened 2025-06-13
Workaround the Julia isues that we observe in CI

tmadlener:
BEGINRELEASENOTES
- Disable the julia tests in the sanitizers workflows as they do not test anything meaningful there
- Use podman for now to run the CI that requires Julia

ENDRELEASENOTES

Workaround from: https://its.cern.ch/jira/browse/SPI-2838

Once https://github.com/AIDASoft/run-lcg-view/pull/10 is merged, we can switch back to the `aidasoft` version of that action


tmadlener:
The upstream cvmfs issue is https://github.com/cvmfs/cvmfs/issues/3868

tmadlener:
Key4hep failures will be fixed by #788

Issue #581, opened 2024-04-10
Warnings when generating ROOT dictionaries


jmcarcell:
When compiling, this warnings appear:
```
[113/120] Generating TestDataModel_v3Dict.cxx, TestDataModel_v3DictDict.rootmap
Warning: Unused class rule: ExampleWithArrayv2
Warning: Unused class rule: ex42::ExampleWithNamespacev2
Warning: Unused class rule: ExampleWithUserInitv2
```
It would be nice for them not to appear; we should strive for clean builds without warnings or other messages. This is not very important and I'm not annoyed enough by them to have a look (yet)

tmadlener:
Yeah, these come from the dictionary generation for the schema evolution tests. I am not sure they are completely silencable though. I think I tried for a bit, but it wasn’t trivial so I moved on.

PR #792, opened 2025-06-17
Remove old schema components from selection.xml

jmcarcell:
BEGINRELEASENOTES
- Remove old schema components from the selection.xml

ENDRELEASENOTES

I think they shouldn't be needed since they are not accessible anymore from Python.

Fix https://github.com/AIDASoft/podio/issues/581.


tmadlener:
I am closing this because this is also fixed in #799 in a slightly different way. There we make sure to generate these as `XYZData` instead of only `XYZ` which is the reason we get the warning.

PR #798, opened 2025-06-26
Allow calling finish() multiple times, with no effect after the first

jmcarcell:
`finish()` can be called now multiple times for the TTree and RNTuple writers and it will crash after the first one. We check in the destructor but we can check instead inside `finish()` and disallow this behaviour.

BEGINRELEASENOTES
- Allow calling finish() multiple times, with no effect after the first by checking the value of `m_finished`.

ENDRELEASENOTES

The case where this happened (with @fredrikshaw) was trying to write immediately a file in Python which is not trivial. Doing:

``` python
writer = Writer(output_file)
writer.write_frame(frame, 'events')
writer._writer.finish()
```
currently crashes because `finish` will be called later (at the end of the script) again because of https://github.com/AIDASoft/podio/pull/539. Whether `_writer.finish()` should be used is a different question, because `del writer` does not always (in my tests, never) trigger the destructor and sometimes doing `gc.collect()` is not enough. In any case it's probably good not to let this crash happen. Alternatively, some public method could be added and then we could check https://github.com/AIDASoft/podio/blob/master/python/podio/base_writer.py#L19.

### DIFF
--- a/src/RNTupleWriter.cc
+++ b/src/RNTupleWriter.cc
@@ -28,9 +28,7 @@ RNTupleWriter::RNTupleWriter(const std::string& filename) :
 }
 
 RNTupleWriter::~RNTupleWriter() {
-  if (!m_finished) {
-    finish();
-  }
+  finish();
 }
 
 template <typename T>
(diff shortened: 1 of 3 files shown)


tmadlener (commented):
This looks very sensible to me. Similar changes will probably also apply for SIO(?).

jmcarcell:
Yes, I always forget

tmadlener:
Thanks.

Issue #800, opened 2025-06-27
Schema evolution does not work for renamed datatypes


tmadlener:
When a datatype is renamed it can currently not be read back, although we claim that we can do this. This has slipped through until now, because there simply was not test to check this.

PR #797, opened 2025-06-26
Refactor schema evolution related code generation

tmadlener:
BEGINRELEASENOTES
- Refactor parts of the code generation that are related to schema evolution
- Remove some duplication and homogenize information from schema changes detection system

ENDRELEASENOTES

This is probably still somewhat ongoing work that falls out of trying to fix some short comings of the whole system on the short run. This PR will collect the changes that are purely function preserving. Other PRs will build on top of this. Parts of the changes here already introduce new / generalized functionality that will be required for those.

PR #801, opened 2025-06-27
[WIP] Reproducer for #800

tmadlener:
BEGINRELEASENOTES
- ...

ENDRELEASENOTES

- [x] Includes  #797 

This is a simple reproducer for #800; ~[actual diff without #797](https://github.com/tmadlener/podio/compare/refactor-schema-evol-detection...tmadlener:podio:fix-datatype-rename-schema-evolution?expand=1)~


tmadlener:
Now we catch this already during code generation, but we still don't support it.

jmcarcell:
Included in https://github.com/AIDASoft/podio/pull/979.

Issue #794, opened 2025-06-24
Bypassing pythonizations in Python to get a mutable object from a read collection


jmcarcell:
Noticed with @fredrikshaw, it is possible to not import EDM4hep when reading from a file and then get mutable objects. What is the expected result? When using podio as it is, `operator[]` returns a mutable object, while `at` returns the immutable object. With pythonizations (https://github.com/AIDASoft/podio/blob/master/python/podio/pythonizations/collection_subscript.py#L48), `operator[]` calls `at` and therefore both return immutable objects. The pythonizations are opt-in from the datamodel, so it is possible to either not opt-in or not to import the model, which will not load the pythonizations, and allow code like this to happen:

``` python
import podio
reader = podio.root_io.Reader('file.root')
frame = reader.get('events')[0]
coll = frame.get('CollectionName')
print(coll.at(0)) # Immutable
print(coll[0]) # Mutable! Then I can set member values!
hit = coll[0]
hit.setEnergy(123)
print(hit.getEnergy()) # 123, all good, the energy changed
print(coll[0].getEnergy()) # 123
```
And you can also save the collection to a frame which won't save the new value, and maybe there are more subtleties when one uses this "new value". By adding `import edm4hep` at the top, pythonizations are loaded and this code can not run. I think this shouldn't be allowed, since it is not allowed in C++. I thought a bit about the solutions:
- Force at least some pythonizations to load for all datamodels. I think this is not very hard to do since importing podio is mandatory
- There is a way of selecting overloads in `cppyy` (https://cppyy.readthedocs.io/en/latest/functions.html#overloading, I think supported since recently in `cppyy`) that works. In the example above I could do:
``` python
cppyy.gbl.edm4hep.SimCalorimeterHitCollection.__getitem__ = cppyy.gbl.edm4hep.SimCalorimeterHitCollection.__getitem__.__overload__(':any:', True)
```
and this will select the `const` overload for `operator[]`, so `setEnergy` will not be possible anymore. However this would live in edm4hep, so if pythonizations are not loaded because it is not imported, I don't see how this would work.

jmcarcell:
For some reason, in plain podio, `operator[]` returns the mutable object and `at` returns the immutable object. In the example above:
```
print(coll.at(0))
print(coll[0])
```
gives
```
<cppyy.gbl.edm4hep.SimCalorimeterHit object at 0x4018c03ee70>
[...]

PR #806, opened 2025-07-02
Make [] return an immutable object in Python

jmcarcell:
BEGINRELEASENOTES
- Make [] return an immutable object in Python when the pythonizations from Python are not loaded (before this would be a mutable object)
- Add a test that checks that we get an immutable object in that case
- Include python directories in a few places when building

ENDRELEASENOTES

Fix https://github.com/AIDASoft/podio/issues/794. Out of all the options I think this is the cleanest, remapping `[]` in Python to `at` in C++. See https://github.com/AIDASoft/podio/pull/806#issuecomment-3035130121 for the possible options and their status. Since raising `IndexError` from C++ (not needing any pythonizations from Python) didn't work in both LCG and Key4hep, I went with this compromise.

Note: `__cppyy_explicit_pythonize` has been in cppyy at least since version 1.5.5 (https://github.com/wlav/cppyy/commit/999ccf6f3423aca5ec014259538dcb59257c7b1d),  which is earlier than what is required by the minimum version of ROOT that we require (I get 1.6.2 for ROOT 6.28.04).


python/templates/Collection.h.jinja2
```diff
@@ -189,6 +191,13 @@ public:
     return rend();
   }
 
+  // For cppyy pythonization, to make operator[] return
+  // the immutable object instead of a mutable one
+  static void __cppyy_explicit_pythonize__(PyObject* klass, const std::string&) {
+    auto* methodObject = PyObject_GetAttrString(klass, "at");
+    PyObject_SetAttrString(klass, "__getitem__", methodObject);
+    Py_DECREF(methodObject);
```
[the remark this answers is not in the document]
tmadlener: I would be in favor of keeping the `IndexError` otherwise we might break some scripts that rely on that instead of a `cppyy` wrapped std exception.

CMakeLists.txt
```diff
@@ -87,6 +87,7 @@ if(ENABLE_DATASOURCE)
   list(APPEND root_components_needed ROOTDataFrame)
 endif()
 find_package(ROOT ${root_min_version} REQUIRED COMPONENTS ${root_components_needed})
+find_package(Python3 REQUIRED COMPONENTS Development)
```
tmadlener, line 90:
```suggestion
```
This is unnecessary, since we already do that here:

https://github.com/AIDASoft/podio/blob/7ebfe4193f400c6852b82db53be3abdcc16bc6e9/cmake/podioBuild.cmake#L201-L204

tmadlener (commented):
Since python headers now become necessary to compile generated EDMs, I think we also need to make sure that the `Python3 Development` components are found for dependent packages:

https://github.com/AIDASoft/podio/blob/7ebfe4193f400c6852b82db53be3abdcc16bc6e9/cmake/podioConfig.cmake.in#L28-L32

python/templates/Collection.h.jinja2
```diff
@@ -189,6 +191,28 @@ public:
     return rend();
   }
 
+  // For cppyy pythonization, to make operator[] return
+  // the immutable object instead of a mutable one
+  std::unique_ptr<{{ class.bare_type }}> __get_const_idx(size_t idx) const {
```
m-fila, line 196: Double underscore in names is reserved in C++

Why does it return a `unique_ptr`, isn't it supposed to be `at` but with different exception type?

jmcarcell: Because the Python API expects pointers, not objects, and if an object is returned after IndexError is raised then you don't get IndexError but cppyy complains. Unfortunately it seems this is only working for the key4hep nightlies. Then it's not the version of ROOT since it doesn't work for dev3 and dev4.

m-fila: Mysterious are the ways of cppyy

tmadlener: Just for completeness and to make sure someone else also thinks about this; This does not introduce any new lifetime issues (apart from the ones we already have), right? I.e. The element pointed to with the `unique_ptr` can in principle outlive the collection and than the internal `Obj*` will just point to freed memory. This is effectively the same as is already the case if we keep the objects around longer than the collection. However, we should make sure that it doesn't attempt to double free. I think it does not, but maybe another set of eyes can look at it as well?

jmcarcell: I think that is correct. It seems to be fine, I can index collections that are later written and the Python scripts I have tried finish without crashes; those would also show up in CI if there were any.

tmadlener (commented):
Sounds a bit like that is the best we can do for now? It is a bit unfortunate that it is two different error categories that users have to be aware of if they want to catch it in python, but I also don't think it is really worth the time to figure out why it only works on certain ROOT installations.

Can we make a minimal reproducer to send this upstream somehow (and where would it go)?

m-fila:
The thing I originally wanted to try is binding the `__getitem__` to a C-function operating with pure C-API. The previous approach was binding to a C++ function that internally was using C-API to rise Python exception but itself was pythonized with cppyy. 

I mean something like this:

```c
PyObject * get_item(PyObject *self, PyObject *idx);
```

instead of 

```cpp
std::unique_ptr<MCParticle> MCParticleCollection::__get_const_idx(size_t idx) const;
```

tmadlener:
Is there an estimate of how much work that entails? It doesn't sound trivial to me as it likely entails also reaching into some cppyy internals to get to the `PyObject` in the first place?

jmcarcell:
It's certainly not trivial. You also need to bind one of the arguments since `__getitem__` only gets one argument. Then what @tmadlener said above taking into account that what `[]` and `at` return are temporaries and an address of that can not be taken. With no example doing this in `cppyy` and having seen how this seems to be fragile (raising `IndexError` from python doesn't always work with the `get_const_idx` function) I'm not convinced.

python/templates/Collection.h.jinja2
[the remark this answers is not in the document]
jmcarcell: Answering to the first comment here, I removed `{Python3_LIBRARIES}` from linking which may have been there from a previous iteration. I think this is the maximum that can be done.

jmcarcell:
Close in favor of https://github.com/AIDASoft/podio/pull/810

PR #807, opened 2025-07-04
Make PODIO_GENERATE_DICTIONARY a thin wrapper around REFLEX_GENERATE_DICTIONARY

tmadlener:
BEGINRELEASENOTES
- Make `PODIO_GENERATE_DICTIONARY` a thin wrapper around `REFLEX_GENERATE_DICTIONARY` instead of a re-implementation

ENDRELEASENOTES

We depend on more recent ROOT versions now, so there is no need to keep this re-implementation around. `PODIO_GENERATE_DICTIONARY` becomes a thin wrapper around `REFLEX_GENERATE_DICTIONARY` where we only keep the additional clang-tidy settings on the dictionary target and the disabling of `Woverlength-strings`.

### DIFF
--- a/cmake/podioMacros.cmake
+++ b/cmake/podioMacros.cmake
@@ -13,6 +13,15 @@
 # ---------------------------------------------------------------------------------------------------
 function(PODIO_GENERATE_DICTIONARY dictionary)
   REFLEX_GENERATE_DICTIONARY(${ARGV})
+
+  # Dictionary payloads become very large for our generated EDMs, so we
+  # explicitly silence this warning here.
+  set_source_files_properties(${gensrcdict}
+    PROPERTIES
+    GENERATED TRUE
+    COMPILE_FLAGS "-Wno-overlength-strings"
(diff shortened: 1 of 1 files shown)


cmake/podioMacros.cmake
```diff
@@ -5,100 +5,22 @@
...
-  IF(TARGET ${dictionary})
-    target_sources(${dictionary} PRIVATE ${gensrcdict})
-  ENDIF()
-  set(gensrcdict ${dictionary}.cxx PARENT_SCOPE)
+  # Dictionary payloads become very large for our generated EDMs, so we
+  # explicitly silence this warning here.
   set_source_files_properties(${gensrcdict}
     PROPERTIES
     GENERATED TRUE
     COMPILE_FLAGS "-Wno-overlength-strings"
```
jmcarcell, line 22: Maybe it's not great that the length is 10 times what compilers are required to support, maybe that's a problem some day...

PR #809, opened 2025-07-07
Simplify constructors by using member initializers

jmcarcell:
BEGINRELEASENOTES
- Simplify constructors by using member initializers and only initialize in the constructor what changes with respect to the defaults

ENDRELEASENOTES

The reason was that in one file there was a member initializer that had some value but all the constructors were setting that member to a different value, then there is no point in having a default value that will never be used. Most constructors get simplified and in some cases the default constructor in the implementation file can be removed and declared with `= default`. I think I got most of them, I would say this is how it is being done for most constructors in the repo.


python/templates/Collection.h.jinja2
```diff
@@ -203,8 +203,8 @@ private:
   bool m_isValid{false};
   mutable bool m_isPrepared{false};
   bool m_isSubsetColl{false};
-  uint32_t m_collectionID{0};
-  mutable std::unique_ptr<std::mutex> m_storageMtx{nullptr};
+  uint32_t m_collectionID{static_cast<uint32_t>(podio::ObjectID::untracked)};
+  mutable std::unique_ptr<std::mutex> m_storageMtx{std::make_unique<std::mutex>()};
```
tmadlener, line 207: I think we had this construct at some point and moved to the initialization in the member initializer list due to some nvcc issues in k4clue. https://github.com/AIDASoft/podio/pull/333 

Could you verify that nvcc can now handle this?

python/templates/CollectionData.h.jinja2
```diff
@@ -83,11 +84,11 @@ public:
 private:
   // members to handle 1-to-N-relations
 {% for relation in OneToManyRelations %}
-  podio::UVecPtr<{{ relation.namespace }}::{{ relation.bare_type }}> m_rel_{{ relation.name }};  ///< Relation buffer for read / write
+  podio::UVecPtr<{{ relation.namespace }}::{{ relation.bare_type }}> m_rel_{{ relation.name }}{new std::vector<{{ relation.namespace }}::{{ relation.bare_type }}>()};  ///< Relation buffer for read / write
```
tmadlener, line 87: Not sure if it's more readable as it is already quite bad at the moment, but could we do this with `std::make_unique`?

python/templates/Collection.h.jinja2
```diff
@@ -203,8 +203,8 @@ private:
   bool m_isValid{false};
   mutable bool m_isPrepared{false};
   bool m_isSubsetColl{false};
-  uint32_t m_collectionID{0};
-  mutable std::unique_ptr<std::mutex> m_storageMtx{nullptr};
+  uint32_t m_collectionID{static_cast<uint32_t>(podio::ObjectID::untracked)};
```
tmadlener, line 206: This changes the behavior. I think it's the right thing to do, but we should probably declare this at least in the release notes.

include/podio/GenericParameters.h
```diff
@@ -170,14 +170,14 @@ class GenericParameters {
...
-  MapType<double> _doubleMap{};          ///< The map storing the double values
-  mutable MutexPtr m_doubleMtx{nullptr}; ///< The mutex guarding the double map
+  MapType<int> _intMap{};                                        ///< The map storing the integer values
+  mutable MutexPtr m_intMtx = std::make_unique<std::mutex>();    ///< The mutex guarding the integer map
+  MapType<float> _floatMap{};                                    ///< The map storing the float values
+  mutable MutexPtr m_floatMtx = std::make_unique<std::mutex>();  ///< The mutex guarding the float map
+  MapType<std::string> _stringMap{};                             ///< The map storing the string values
+  mutable MutexPtr m_stringMtx = std::make_unique<std::mutex>(); ///< The mutex guarding the string map
+  MapType<double> _doubleMap{};                                  ///< The map storing the double values
+  mutable MutexPtr m_doubleMtx = std::make_unique<std::mutex>(); ///< The mutex guarding the double map
```
m-fila, line 180: This is very minor but is there a reason to change initialization syntax for mutexes? Usually we go with direct-list-initialization

python/templates/Collection.h.jinja2
jmcarcell: CUDA stuff was removed from k4Clue in https://github.com/key4hep/k4Clue/commit/13709a97ee286d9420d2789db05cf4016e1ae698 so it doesn't use it anymore.

include/podio/GenericParameters.h
jmcarcell: Changed to initializer list

python/templates/Collection.h.jinja2
jmcarcell: In which case? Both constructors were initializing to `podio::ObjectID::untracked`, so it was impossible to get 0.

PR #810, opened 2025-07-10
replace Python callback with C++ callback for subscript pythonization

### LINKED ISSUE #794: Bypassing pythonizations in Python to get a mutable object from a read collection

m-fila:
BEGINRELEASENOTES
- Add automatic pythonization of collection subscript with C++ callback
- Deprecate pythonization of collection subscript with Python callback

ENDRELEASENOTES

This is an alternative to #806 that keeps the `IndexError`

The C++ callback should be always applied so there is no need to call any extra functions to get the pythonization.
The callback is using pure Python C-API to delegate call to `__getitem__` to `at` and raise `IndexError` if there was an exception.

Closes #794

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -59,7 +59,7 @@ IF (BUILD_TESTING)
           ${strace_program} -e open,openat,creat -f -o ${CMAKE_CURRENT_BINARY_DIR}/strace.log \
             python3 ${CMAKE_CURRENT_SOURCE_DIR}/podio/test_strace.py && \
           lines=\$(wc -l < ${CMAKE_CURRENT_BINARY_DIR}/strace.log) && \
-          if [ \$lines -ge 12000 ]; then
+          if [ \$lines -ge 16000 ]; then
             echo \"Too many lines in strace.log (\$lines), test failed!\";
             exit 1;
           fi"


jmcarcell:
Interesting, I did something similar but apparently `auto* at_method = PyObject_GetAttrString(self, "at");` followed by `PyObject_CallFunctionObjArgs(at_method, key, NULL);` doesn't seem to work (I get a null pointer). Maybe because it's a method? Anyway it doesn't matter now.
There is an additional test in https://github.com/AIDASoft/podio/pull/806 without the pythonizations that would be good to include here.

jmcarcell:
This is a comment I made in the Key4hep meeting today: If we change `[]` and also iterators `for elem in coll` to always return immutable in python then the only way to get mutable in Python would be using `.create()` which limits modifying collections (unless the mutable handles are saved somewhere, for example in list, and passed around).

python/podio/pythonizations/collection_subscript.py
tmadlener, line 1: Can this be removed such that we have a deprecation warning (and ideally a transparent switch to the new behavior)?

include/podio/detail/Pythonizations.h
```diff
@@ -0,0 +1,60 @@
...
+
+// Helper to register the subscript pythonization callback as `__getitem__` method
+static inline void pythonize_subscript(PyObject* klass, const std::string& name) {
+  static PyMethodDef ml = {"subscipt_pythonization", subscript, METH_VARARGS, R"(
+        Raise an `IndexError` exception if an index is invalid.
+        The `__getitem__` will return immutable datatype objects instead of the mutable ones.
+  )"};
+  auto* func = PyCFunction_New(&ml, klass);
+  if (!func) {
+    throw std::runtime_error("Failed to create Python subscript function for class " + name);
```
tmadlener, line 42: When will these throw if they throw? During `import XYZ`? Or earlier / later?

tmadlener (commented):
Very nice.

Do we have a test already for checking that we now get immutable object handles when looping, resp. when using `[]` for accessing collections? Mainly asking because I can't see it in the diff for this PR.

include/podio/detail/Pythonizations.h
tmadlener: Thanks for the clarification. I don't think we want to catch these as I wouldn't even be sure how we would recover from this. In the end we will more likely than not just leave the whole loaded cppyy module in a broken state in case we ever throw this exception(?).

But all the information that is spat out should at least point users towards an appropriate github repository to open an issue ;)

m-fila:
> Unless you want to do it, I will resolve the merge conflict and then I don't see anything that keeps us from merging this.

I'll do it this afternoon, also check the deprecation

python/podio/pythonizations/collection_subscript.py
m-fila: I brought back the opt-in collection subscript pythonization and deprecated that module.

Currently we have to ways to load the opt-in pythohonizations:
- import specific pythonization from `podio.pythonizations` and apply directly. This prints deprecation warning correctly
- `load_pythonizations` that will apply all available. This is what we have in EDM4hep. The Python mechanism hides the warning and this is expected from what I understood about `warnings` - `python -Wd` should be used to opt-in to hidden warnings. Or we could force the visibility with `warnings.simplefilter("always", DeprecationWarning)` in that module

Another point is that now it's possible to load both automatic C++ callback pythonization and opt-in Python callback (this is what will happen in EDM4hep). This is redundant but shouldn't produce any errors or problems.
I couldn't find any information on the order in which C++ and Python side pythonizations are loaded, at least in my build it seems to be first C++ then Python

.cppcheck-suppressions.txt
```diff
@@ -1,20 +1,15 @@
 noExplicitConstructor:*
+unusedFunction:*
```
m-fila, line 2: cppcheck is seems to be really struggling with `unusedFunction`. Several new false positives were reported https://github.com/AIDASoft/podio/actions/runs/17125786028/job/48577015115#step:4:1140 . I think at this point it's easier to just give up on this check

python/podio/pythonizations/collection_subscript.py
```diff
@@ -1,8 +1,17 @@
...
 from .utils.pythonizer import Pythonizer
 
+warnings.warn(
+    "The module 'collection_subscript' is deprecated and will be removed in a future release. "
+    "The subscript pythonization is now defined in generated C++ class definition "
+    "and applied automatically.",
+    FutureWarning,
+    stacklevel=2,
+)
+
```
tmadlener, line 14: Thinking about it, `DeprecationWarning` might be more appropriate here. Sorry for being slow.

Just to be sure that I understand how things work **now**: We get the `collection_subscript` pythonization automatically (without possibility to opt out). Previously, you either get it via the catch-all `load_pythonizations` or via the explicit registration of just this pythonization.

Hence, people just using `load_pythonizations` have no change in behavior, but now see a`FutureWarning`, for which there is not even an easy fix, because for them the import happens "magically" behind `load_pythonizations`.

m-fila (commented):
CI is [failing](https://github.com/AIDASoft/podio/actions/runs/17126325819/job/48578794160?pr=810#step:5:1113) due to too many lines in strace trace (13114 while the limit is 12000). Should we just raise the limit or is there something to debug here?

.cppcheck-suppressions.txt
tmadlener: Curious that they haven't been detected before. But for me it's OK to suppress the check entirely.

tmadlener (approved):

jmcarcell:
> > Should we just raise the limit or is there something to debug here?
> 
> For that @jmcarcell is the right person to ask, but he is not currently around. Juding from [#808 (comment)](https://github.com/AIDASoft/podio/pull/808#issuecomment-3051715679) it looks like it's still quite a bit from what it was before the fix for that check.
> 
> I have check this for this PR on an EL9 machine
> 	`@master` 	this PR
> with #808 	14104 	16463
> without #808 	43982 	46347
> 
> To me it looks a bit like the baseline is increased already without this PR. I would be OK with increasing the threshold for this PR until we understand why the baseline seems to be moving.

This is mostly ROOT going through every place in `ROOT_INCLUDE_PATH` trying to find the python includes. We have almost all if not all the repositories that depend on ROOT there so it does increase a bit.

tmadlener:
I think this might be fixable by declaring the `Python_INCLUDE_DIRS` as `PUBLIC SYSTEM` here:
https://github.com/AIDASoft/podio/blob/5f5c6140c951970e06c3ec82dd0e6a15229df655/src/CMakeLists.txt#L26-L30

That should have the side-effect of silencing warnings related to these headers, as they will then be put into the [`INTERFACE_SYSTEM_INCLUDE_DIRECTORIES`](https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_SYSTEM_INCLUDE_DIRECTORIES.html#prop_tgt:INTERFACE_SYSTEM_INCLUDE_DIRECTORIES) target (see [docs](https://cmake.org/cmake/help/latest/command/target_include_directories.html))

PR #811, opened 2025-07-29
Deprecate the usage of -> instead of . for objects that are not pointers

jmcarcell:
An issue in python (https://github.com/wlav/cppyy/issues/312) where a `RecursionError` is thrown seems to happen because we have an `operator->` for collections. Regardless of that, I think it doesn't fit with the rest of podio and should be removed. I have made a few PRs not to use this:
- https://github.com/key4hep/k4geo/pull/500
- https://github.com/key4hep/k4Reco/pull/39
- https://github.com/AIDASoft/DD4hep/pull/1480

BEGINRELEASENOTES
- Deprecate the usage of -> instead of . for objects that are not pointers

ENDRELEASENOTES

### DIFF
--- a/python/templates/Collection.h.jinja2
+++ b/python/templates/Collection.h.jinja2
@@ -74,6 +74,7 @@ public:
   void print(std::ostream& os=std::cout, bool flush=true) const final;
 
   /// operator to allow pointer like calling of members a la LCIO
+  [[deprecated("Use of operator->() is deprecated. Use . instead.")]]
   {{ class.bare_type }}Collection* operator->() { return static_cast<{{ class.bare_type }}Collection*>(this); }
 
   /// Append a new object to the collection, and return this object.


tmadlener (approved):
Curious that this is only in place for the collections. I thought we also had this for the handles, but I couldn't find it.

jmcarcell:
The deprecation seems to generate a lot of warnings in python scripts. Reproducer:
``` python
import podio
import edm4hep
frame = podio.Frame()
particles = frame.put(edm4hep.MCParticleCollection(), "MCParticleCollection")
```
and the warnings:
```
input_line_76:7:90: warning: 'operator->' is deprecated: Use of operator->() is deprecated. Use . instead. [-Wdeprecated-declarations]
      new (ret) (edm4hep::MCParticleCollection*) (((edm4hep::MCParticleCollection*)obj)->operator->());
                                                                                         ^
libedm4hepDict dictionary payload:2494:5: note: 'operator->' has been explicitly marked deprecated here
  [[deprecated("Use of operator->() is deprecated. Use . instead.")]]
    ^
input_line_76:11:53: warning: 'operator->' is deprecated: Use of operator->() is deprecated. Use . instead. [-Wdeprecated-declarations]
      (void)(((edm4hep::MCParticleCollection*)obj)->operator->());
                                                    ^
libedm4hepDict dictionary payload:2494:5: note: 'operator->' has been explicitly marked deprecated here
  [[deprecated("Use of operator->() is deprecated. Use . instead.")]]
```

tmadlener:
Does this script still work with #812 in place? If yes, then we could make a tag and merge that one after that.

tmadlener:
@wdconinc @nathanwbrei I don't think you will have any issues with it, but just as a heads up to look out for some potential deprecation warnings when switching to podio@1.4

PR #812, opened 2025-07-29
Remove the deprecated operator-> in collections

jmcarcell:
BEGINRELEASENOTES
- Remove the deprecated operator-> in collections

ENDRELEASENOTES

Contains https://github.com/AIDASoft/podio/pull/811 and will serve as reminder. In addition, I won't fix the `RecursionError` (https://github.com/wlav/cppyy/issues/312) that would require adding another pythonization since it's not a big issue and it will disappear once this PR is merged.

### DIFF
--- a/python/templates/Collection.h.jinja2
+++ b/python/templates/Collection.h.jinja2
@@ -73,10 +73,6 @@ public:
   /// Print this collection to the passed stream
   void print(std::ostream& os=std::cout, bool flush=true) const final;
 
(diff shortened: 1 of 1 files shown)


tmadlener:
Keeping this open until we have a tag with #811 in to see if we have any actual users of this (I don't think so though).

tmadlener:
> Keeping this open until we have a tag with #811 in to see if we have any actual users of this (I don't think so though).

Well this has obviously been entirely wrong. There are actually quite a few users.

PR #815, opened 2025-08-18
Remove an unnecessary datatype from the test datamodel

tmadlener:
BEGINRELEASENOTES
- Remove unused datatype from the original model and re-use another one to do the same test

ENDRELEASENOTES

Removed datatype has been introduced in #663 but is actually not necessary.


tmadlener:
In light of #817 this PR can be simplified to only the removal of the unnecessary datatype.

PR #817, opened 2025-08-19
Refactor the way schema evolution tests are done

### LINKED ISSUE #581: Warnings when generating ROOT dictionaries

tmadlener:
BEGINRELEASENOTES
- Refactor the way schema evolution tests are done to make them easier to adapt and understand
  - Introduce a mini testing framework that allows to focus on a single schema change for a given test
  - Remove the duplication of the test datamodel with various schema changes sprinkled across it
- Enable some schema evolution tests for RNTuple 

ENDRELEASENOTES

Fixes #581 (in passing, since those dictionaries are no longer generated).

- [x] Includes changes from #815 

These changes should simplify the tests for schema evolution as it allows to test a single type of evolution in isolation. The necessary code overhead to introduce a mini framework is already compensated by the removal of the datamodel duplication, and we already test more this way and have better documentation. Please check the `code_gen/README.md` file for a brief description of how the mini framework works and what is necessary to add new test cases.

One annoying thing at the moment is that due to the way the mini framework is setup it is necessary to move the output files of the ROOT `dictgen` step which triggers it to run on every invocation of the build, regardless of whether any of the inputs have changed. **If anyone knows how to fix that, comments are very much appreciated**. We might also just hide these tests behind a CMake option to make local developments on other parts of podio more seamless.


tmadlener:
It looks like we need some more version guards around the RNTuple tests for schema evolution as not all versions of RNTuple that we test against already support all the necessary operations.

tests/schema_evolution/code_gen/test_utilities.cmake
```diff
@@ -0,0 +1,138 @@
...
+      IO_BACKEND_HANDLERS ${PODIO_IO_HANDLERS}
+      OUTPUT_FOLDER ${output_base}
+      OLD_DESCRIPTION ${test_case}/old.yaml
+      SCHEMA_EVOLUTION ${test_case}/evolution.yaml
+    )
+  else()
+    PODIO_GENERATE_DATAMODEL(datamodel ${test_case}/${model_version}.yaml headers sources
+      IO_BACKEND_HANDLERS ${PODIO_IO_HANDLERS}
+      OUTPUT_FOLDER ${output_base}
+      OLD_DESCRIPTION ${test_case}/old.yaml
```
tmadlener, line 28:
```suggestion
```

tests/schema_evolution/code_gen/README.md
```diff
@@ -0,0 +1,82 @@
...
+    // Additionally, a single element of name elem is present
+    elem.s(42);  // Set the s-member to 42
+  });
+  
+  READ_AS(TestTypeCollection, {
+    // In this code block a collection of type TestTypeCollection
+    // has been read from the provided file and is available as coll.
+    // Additionally, the first element of this collection is present
+    // as elem
+    ASSERT_EQUAL(elem.s(), 42, "Member s does not have the expected value");
```
tmadlener, line 59:
```suggestion
    ASSERT_EQUAL(elem.s(), 42, "Member s does not have the expected value");
```

tests/schema_evolution/code_gen/README.md
```diff
@@ -0,0 +1,82 @@
+# *Code generation* tests for schema evolution
+
+This folder contains small and targeted test cases to ensure that code
+generateion related to schema evolution works as expected. The goal is to have
+rather small test cases that focus on one specific (or a limited set of related)
+schema changes to ensure that the related code generation works as intended.
```
m-fila, line 6:
```suggestion
This folder contains small and targeted test cases to ensure that code
generation related to schema evolution works as expected. The goal is to have
rather small test cases that focus on one specific (or a limited set of related)
schema changes to ensure that the related code generation works as intended.
```

tests/schema_evolution/code_gen/README.md
```diff
@@ -0,0 +1,82 @@
...
+    // as elem
+    ASSERT_EQUAL(elem.s(), 42, "Member s does not have the expected value"); 
+  });
+}
+```
+
+### Additional technical details
+
+For developers who have to touch the internals of the whole machinery a few
+additional information bits about the expecations / assumptions:
```
m-fila, line 65:
```suggestion
For developers who have to touch the internals of the whole machinery a few
additional information bits about the expectations / assumptions:
```

tests/schema_evolution/code_gen/README.md
```diff
@@ -0,0 +1,82 @@
...
+  injected via `target_compile_definitions` in `ADD_SCHEMA_EVOLUTION_TEST` the
+  latter is defined in `check_base.h` directly depending on other pre-processor
+  flag definitions
+- `{WRITE,READ}_AS` are only defined as non-empty macros if
+  `PODIO_SCHEMA_EVOLUTION_TEST_{WRITE,READ}` are defined respectively. This is
+  what makes it possible to re-use the same source file.
+- Both macros are implemented via `{WRITE,READ}_WITH` macros respecitvely. These
+  are slightly more generic and take a writer or reader type as well as a
+  filename as argument to setup the basic writer or reader and a corresponding
+  Frame called `event`.
```
m-fila, line 77:
```suggestion
- Both macros are implemented via `{WRITE,READ}_WITH` macros respectively. These
  are slightly more generic and take a writer or reader type as well as a
  filename as argument to setup the basic writer or reader and a corresponding
  Frame called `event`.
```

tests/schema_evolution/code_gen/test_utilities.cmake
```diff
@@ -0,0 +1,147 @@
...
+    OUTPUT_FOLDER ${output_base}
+  )
+  PODIO_ADD_ROOT_IO_DICT(${model_base}Dict ${model_base} "${headers}" ${output_base}/src/selection.xml
+    OUTPUT_FOLDER ${output_base}
+  )
+
+  # Make sure that each model can be "toggled" at runtime separately.
+  # Effecitvely amounts to moving the byproducts of the dictgen above to the
+  # appropriate place and making sure libraries are built into the right output
+  # directory
```
m-fila, line 50:
```suggestion
  # Make sure that each model can be "toggled" at runtime separately.
  # Effectively amounts to moving the byproducts of the dictgen above to the
  # appropriate place and making sure libraries are built into the right output
  # directory
```

tmadlener:
Found a way to isolate the tests environments without re-running the `dictgen` step for every build. In the end the simplest solution was to copy the files over and leave the original ones in place and simply use a "clean" `WORKING_DIRECTORY` for running the tests. Leaving the original files in places keeps the dependency tracking for the build intact.

tmadlener:
Interesting. I tested quite extensively with different versions of ROOT locally (and in CI) and for me that only started with #803. Can you try to figure out which test is the culprit and also for which setup it fails?

jmcarcell:
I had a look and it fails for the test that reads `datatypes_new_member` whenever the path to the old libraries is present before the path to the new libraries. In this case I had the path to the old library in `ROOT_LIBRARY_PATH` which is not touched by tests. Then, what happens is that the vector `std::vector<ExampleHitData>` that is read is garbage and that goes to an infinite loop that will eventually make the system run out of memory (with Clang 20 and GCC 15 in Release mode). With GCC 15 in debug mode, there is a segmentation fault. I haven't tested #803, but with this information I think https://github.com/AIDASoft/podio/pull/836 shouldn't be needed, at least it wouldn't have changed anything here.
I'm thinking that in the future we can see this happening again, for example adding a new member to EDM4hep and then the old libraries are still around (for example if we use `ROOT_LIBRARY_PATH` as we probably should).
Since it seems to be possible to construct a garbage vector, something like 
```
#ifndef NDEBUG
    if (tmp->size() == 0) {
      for (auto elem : *tmp) {
        assert(false);
      }
    }
#endif
```
in https://github.com/AIDASoft/podio/blob/master/include/podio/CollectionBuffers.h#L88 catches it. Any other check I tried seems to pass fine. Maybe we could add this if it's not too ugly.

PR #824, opened 2025-08-26
add subscript pythonization to podio collections

m-fila:
BEGINRELEASENOTES
- Add automatic collection subscript pythonization to `UserDataCollection` and `LinkCollection`

ENDRELEASENOTES

Follow up to #810,  which missed the C++ pythonization callback for `podio::UserDataCollection` and `podio::LinkCollection`.

Requires #826

### DIFF
--- a/include/podio/UserDataCollection.h
+++ b/include/podio/UserDataCollection.h
@@ -5,6 +5,7 @@
 #include "podio/CollectionBuffers.h"
 #include "podio/DatamodelRegistry.h"
 #include "podio/SchemaEvolution.h"
+#include "podio/detail/Pythonizations.h"
 #include "podio/utilities/TypeHelpers.h"
 
 #define PODIO_ADD_USER_TYPE(type)                                                                                      \
@@ -163,6 +164,11 @@ public:
   /// fully qualified type name of stored POD elements - with namespace
   const std::string_view getDataTypeName() const override {
     return dataTypeName;
+  }
+
+  /// Cppyy protocol to setup the pythonizations for this class. Not to be called directly.
+  static void __cppyy_pythonize__(PyObject* klass, const std::string& name) {
+    podio::detail::pythonizations::pythonize_subscript(klass, name);
   }
 
   /// clear the collection and all internal states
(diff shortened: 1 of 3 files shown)


m-fila:
Yes, unfortunately they were completely missed earlier. Also `load_pythonizations("edm4hep")` doesn't affect any link collections from EDM4hep as they are just aliases to `podio::LinkCollection` which is different namespace

tmadlener:
It looks a bit like in ROOT 6.28 the pythonization is not properly applied? Or something escapes a catch block somehow and confuses cppyys overload resolution: 

https://github.com/AIDASoft/podio/actions/runs/17230837570/job/48884306714?pr=824#step:5:1276

If not easily fixable, I would be OK with just skipping that test for that ROOT version.

m-fila:
Ugh, so actually we have a typo in non-const `LinkCollection::at` which causes compilation error (#826). The never versions of cppyy are happy to propagate the C++ exception and discard the error but older will raise exception for it

tmadlener:
Any ideas why this does not work for `UserDataCollection` for me locally? root 6.36.00 and yesterdays master branch both fail that test and it looks like the pythonization is indeed not loaded for `UserDataCollection`. I tried to figure out whether the symbol is simply missing for `UserDataCollection`, but I also haven't figured out yet into which lib the symbol actually goes because I also cannot find it e.g. in `libTestDataModel.so` (nor the dictionary).

tmadlener:
It's only the `UserDataCollection` for me. It also seems to be independent of optimization level for the compiler (which I assume would affect potential inlining). Additionally, `ROOT.gInterpreter.LoadFile("podio/UserDataCollection.h")` also doesn't change anything for me.

Not sure if it's a python version thing or not. Using the nightlies and building on top of them works for me. So for now this is just a minor annoyance for me and probably not worth investigating in too many details.

tmadlener:
Locally I can reproduce it on Ubuntu 24, with gcc 13, ROOT 6.36.00 and master from yesterday. Python is 3.12 for me locally, maybe that also plays a role?

PR #826, opened 2025-08-26
Fix typo causing error in non-const overload of `LinkCollection::at(size_t)`

m-fila:
BEGINRELEASENOTES
- Fix typo causing error in non-const overload of `LinkCollection::at(size_t)`

ENDRELEASENOTES

Discovered in #824, there was a typo in non-const overload of `LinkCollection::at(size_t)`. `LinkCollectionData` (`m_storage`) doesn't have `at` method
I'm not sure how it was able to pass this far


m-fila:
`decltype` makes unevaluated context so it just looks at the signatures - there could be no implementation at all and it should still work.

PR #829, opened 2025-08-27
Add an `UserDataCollectionList` with the supported types and a test

jmcarcell:
BEGINRELEASENOTES
- Add an `UserDataCollectionList` with the supported types and a test

ENDRELEASENOTES

See https://github.com/key4hep/EDM4hep/pull/443#issuecomment-3224132875. In use in https://github.com/key4hep/EDM4hep/pull/443


tmadlener:
One question (looking at key4hep/EDM4hep#443). For the generated datatypes the list contains the datatypes (aka the handles), but for UserDataCollection it contains the full collection type, right? Should we also generate the datamodel type lists containing the collections? At this point it's probably too late to switch, so we would have to generate that in addition. I think it would make the usage a bit more intuitive as the user data collection type list doesn't contain a different category of types, compared to the one that we generate for the datamodel type.s

Additionally, it looks like we are (always?) interested in the collection type more than in the datatype, e.g.:

- https://github.com/key4hep/k4FWCore/blob/8784c58ed3c6b19dd1e7727071fb7e8dee4128c3/k4FWCore/components/EfficiencyFilter.cpp#L99-L100
- https://github.com/key4hep/k4FWCore/blob/8784c58ed3c6b19dd1e7727071fb7e8dee4128c3/k4FWCore/components/CollectionMerger.cpp#L85
- https://github.com/key4hep/EDM4hep/blob/cff2a2d318f5e3dc5e340350ed6072dcbfe754a9/tools/include/edm4hep2json.hxx#L38

m-fila:
Oh, right I completely forgot that we generate list of datatypes not collections. I agree that it's too late to switch (also we have type list of interfaces for which there is no collection type). Having both list of datatype and list of collections seems a bit redundant but should be ok

jmcarcell:
Well originally I implemented the collections in https://github.com/AIDASoft/podio/pull/761 but after some comments the decision was not to. I can add them in another PR.

jmcarcell:
> Nice. I have no strong opinion on whether the test for the `podio::CollectionType` concept should be using the type list or the tuple. I would be happy to merge this as is, but if you want to change it let me know.

I'll leave as it is then since I don't have a strong opinion. Also https://github.com/AIDASoft/podio/blob/master/src/UserDataCollection.cc#L56 could be changed if we really wanted to.

andresailer:
I think the release note needs an update to `UserDataCollectionTypes` ? And some explicit explanation what is added here?

tmadlener:
Good catch. Thanks. Done

PR #832, opened 2025-08-28
Add types with all the collection types of a datamodel in datamodel.h

jmcarcell:
BEGINRELEASENOTES
- Add types with all the collection types of a datamodel in datamodel.h
- Add tests for the new types
- Use single spaces instead of two spaces between types in the list

ENDRELEASENOTES

See https://github.com/AIDASoft/podio/pull/829 and the original PR https://github.com/AIDASoft/podio/pull/761

### DIFF
--- a/python/templates/datamodel.h.jinja2
+++ b/python/templates/datamodel.h.jinja2
@@ -12,27 +12,27 @@
 namespace {{ package_name }} {
 
 using {{ package_name }}DataTypes = podio::utils::TypeList<
-  {{ datatypes | map('string') | join(', ')}}
+  {{ datatypes | join(', ')}}
 >;
 
 using {{ package_name }}DataCollectionTypes = podio::utils::TypeList<
 {%  if datatypes %}
-  {{ datatypes | map('string') | join('Collection, ') ~ 'Collection' }}
+  {{ datatypes | join('Collection, ') ~ 'Collection' }}
 {% endif %}
 >;
 
 using {{ package_name }}LinkTypes = podio::utils::TypeList<
-  {{ links | map('string') | join(', ')}}
+  {{ links  | join(', ')}}
 >;
 
 using {{ package_name }}LinkCollectionTypes = podio::utils::TypeList<
 {% if links %}
-  {{ links | map('string') | join('Collection, ') ~ 'Collection' }}
+  {{ links | join('Collection, ') ~ 'Collection' }}
 {% endif %}
 >;
 
 using {{ package_name }}InterfaceTypes = podio::utils::TypeList<
-  {{ interfaces | map('string') | join(', ')}}
+  {{ interfaces | join(', ')}}
 >;
 
 }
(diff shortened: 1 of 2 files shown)


python/templates/datamodel.h.jinja2
```diff
@@ -13,13 +13,25 @@ namespace {{ package_name }} {
 
 using {{ package_name }}DataTypes = podio::utils::TypeList<
 {% for name in datatypes %}
-  {{ name }}{% if not loop.last %},{% endif %}
+{% if loop.first %}  {% else %} {% endif %}{{ name }}{% if not loop.last %},{% endif %}
+{% endfor %}
+>;
```
tmadlener, line 18: This looks suspiciously like it could simply be a 
```jinja2
using {{ package_name }}DataTypes = podio::utils::TypeList<
  {{ datatypes | map(attribute="name") | join(", ") }}
>;
```

Similar below.

For the version with the collections, it might be necessary to implement a small filter, similar to `quoted_sv` that we already have:
https://github.com/AIDASoft/podio/blob/73f77abe6739aad130d8daf30cdf4a455756100a/python/podio_gen/cpp_generator.py#L609-L612
usage:
https://github.com/AIDASoft/podio/blob/73f77abe6739aad130d8daf30cdf4a455756100a/python/templates/DatamodelDefinition.h.jinja2#L24

python/templates/datamodel.h.jinja2
```diff
@@ -12,21 +12,27 @@
...
 using {{ package_name }}DataTypes = podio::utils::TypeList<
-{% for name in datatypes %}
-  {{ name }}{% if not loop.last %},{% endif %}
-{% endfor %}
+  {{ datatypes | join(', ')}}
+>;
+
+using {{ package_name }}DataCollectionTypes = podio::utils::TypeList<
+{%  if datatypes %}
+  {{ datatypes | join('Collection, ') ~ 'Collection' }}
```
tmadlener, line 20: I didn't know this existed. This is pretty neat. I hope we still get what it does in a few months ;)

python/templates/datamodel.h.jinja2
jmcarcell: Simplified but without filter, adding `Collection` to `join` and making sure it's added to the last element.

python/templates/datamodel.h.jinja2
jmcarcell: Yeah I'm not sure it's much better than the for loop but apparently it's not so easy to concatenate to every element

tmadlener: I think for this case it is OK as it is quite obvious what the result should be from the name and the surrounding context.

python/templates/datamodel.h.jinja2
```diff
@@ -12,21 +12,27 @@
 namespace {{ package_name }} {
 
 using {{ package_name }}DataTypes = podio::utils::TypeList<
-{% for name in datatypes %}
-  {{ name }}{% if not loop.last %},{% endif %}
-{% endfor %}
+  {{ datatypes | join(', ')}}
+>;
+
+using {{ package_name }}DataCollectionTypes = podio::utils::TypeList<
```
tmadlener, line 18:
```suggestion
using {{ package_name }}CollectionTypes = podio::utils::TypeList<
```
Just realized that this should have been without the `Data` here. Currently, it reads `edm4hepDataCollectionTypes` for example.

Issue #825, opened 2025-08-26
Could you use test fixtures to better manage test dependencies


tmadlener:
(Opening this because I just got annoyed by it and it would be nice to fix it at some point).

We have quite a few tests where one test writes a file and then one / several tests read that file. At the moment we usually do this via the `DEPENDS` test property, e.g.:

https://github.com/AIDASoft/podio/blob/abf67b30b44c1779fa728a9d6ac52815b5f21c98/tests/root_io/CMakeLists.txt#L41-L50

However, `DEPENDS` only orders the test running temporally, which means that if the test that writes the file fails, the test that attempts to read the file still runs and also fails. See, e.g.

```console
  The following tests FAILED:
  Errors while running CTest
  	  7 - read_python_frame_root (Subprocess aborted)
  	 36 - write_python_frame_root (Timeout)
```

from [this CI run](https://github.com/AIDASoft/podio/actions/runs/17216566786/job/48841521821).

Additionally, simply running the read test, will not automatically run the write test, which either means that the file does not exist yet, or potentially even worse, we try to read an older file, leading to spurious test failures (e.g. in local development).

CMake / CTest has support for test fixtures, which make sure that dependent tests are only run if the "setup part" succeeds, and also ensure that all the setup fixtures are run before the actual read test is run. See, e.g. the [CMake documentation](https://cmake.org/cmake/help/latest/prop_test/FIXTURES_REQUIRED.html#prop_test:FIXTURES_REQUIRED), and some usage in k4FWCore:

https://github.com/key4hep/k4FWCore/blob/8784c58ed3c6b19dd1e7727071fb7e8dee4128c3/test/k4FWCoreTest/CMakeLists.txt#L116
https://github.com/key4hep/k4FWCore/blob/8784c58ed3c6b19dd1e7727071fb7e8dee4128c3/test/k4FWCoreTest/CMakeLists.txt#L119

PR #834, opened 2025-09-02
Add fixtures for tests that read and write files

jmcarcell:
Fix https://github.com/AIDASoft/podio/issues/825

BEGINRELEASENOTES
- Add CMake fixtures for tests that read and write files so that reading doesn't happen if writing fails and running the reading tests requires the write tests

ENDRELEASENOTES
  
Seems to be correct since running `ctest -j 16 --schedule-random` doesn't fail.

Remember the useful `ctest -FA ".*"` to ignore fixtures.


tmadlener:
Thanks for getting this started (although, I didn't think about you when labeling this as *good first issue* ;) ). I have added fixtures for some more tests where we used `DEPENDS` before. Most importantly some of the tools and python tests. I have also tried to harmonize the names of all the fixtures, such that they follow the convention `podio_write[_<sub-feature>]_<io-backend>_fixture`

PR #836, opened 2025-09-05
Add a memory limit to schema evolution tests

tmadlener:
Avoids freezing systems and killing CI when running into issues where ROOT starts to run away with memory

BEGINRELEASENOTES
- Run schema evolution read tests with limited memory to avoid ROOT running away with all the memory.

ENDRELEASENOTES

Triggered by https://github.com/AIDASoft/podio/pull/817#issuecomment-3257539523

@jmcarcell Had the same issue as you, but for me the all tests in #817 ran without issue and only when I started to add failing tests things started to break.

Lifted from #803

### DIFF
--- a/.github/workflows/sanitizers.yaml
+++ b/.github/workflows/sanitizers.yaml
@@ -52,6 +52,7 @@ jobs:
               -DENABLE_JULIA=OFF \
               -DENABLE_RNTUPLE=ON \
               -DENABLE_DATASOURCE=ON \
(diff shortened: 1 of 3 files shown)


tmadlener:
I will merge this for now, until we have a better solution for this (see https://github.com/AIDASoft/podio/pull/817#issuecomment-3266748609 and following comments). At least it will prevent freezing of machines.

PR #838, opened 2025-09-09
Remove potentially confusing "Data" from generated type list name

tmadlener:
BEGINRELEASENOTES
- Remove the `Data` from the generated type list containing all generated datatype collection names

ENDRELEASENOTES

As mentioned here: https://github.com/AIDASoft/podio/pull/832/files#r2333176757

Originally introduced in #832

### DIFF
--- a/python/templates/datamodel.h.jinja2
+++ b/python/templates/datamodel.h.jinja2
@@ -15,7 +15,7 @@ using {{ package_name }}DataTypes = podio::utils::TypeList<
   {{ datatypes | join(', ')}}
 >;
 
-using {{ package_name }}DataCollectionTypes = podio::utils::TypeList<
+using {{ package_name }}CollectionTypes = podio::utils::TypeList<
 {%  if datatypes %}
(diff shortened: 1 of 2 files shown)


jmcarcell:
Ah the reason for this is that the types without collections have either `Data` or `Link` on it so currently the name is the same for the collection types with `Collection` inserted
```
edm4hepDataTypes -> edm4hepDataCollectionTypes
edm4hepLinkTypes -> edm4hepLinkCollectionTypes
```
Maybe it would be more consistent if `Data` is removed in both. `edm4hepTypes` and `edm4hepCollectionTypes`. `Data` maybe makes sense in podio since it's the types that are pointing to the `datamodel::nameData` types, but not so much outside of that.

tmadlener:
I am not sure about removing the `Data` from the other type list. In the end those are the types that are defined as `datatypes` in the YAML file, so there is a somewhat direct link to that at least. I mainly think that `DataCollectionTypes` is a weird ordering, but I couldn't find a better one of those parts that makes more sense. Hence, I arrived at the one without `Data`.

PR #839, opened 2025-09-09
Add the Python headers to the system includes to avoid warnings

jmcarcell:
BEGINRELEASENOTES
- Add the Python headers to the system includes to avoid warnings

ENDRELEASENOTES

See https://github.com/AIDASoft/podio/pull/810#issuecomment-3248092637

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -26,6 +26,8 @@ FUNCTION(PODIO_ADD_LIB_AND_DICT libname headers sources selection )
   target_include_directories(${libname} PUBLIC
     $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
     $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
+  )
+  target_include_directories(${libname} SYSTEM PUBLIC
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -26,6 +26,8 @@ FUNCTION(PODIO_ADD_LIB_AND_DICT libname headers sources selection )
   target_include_directories(${libname} PUBLIC
     $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
     $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
+  )
+  target_include_directories(${libname} SYSTEM PUBLIC
```
tmadlener, line 30:
```suggestion
    SYSTEM PUBLIC
```
Does this work with one call? (Mainly out of interest, rather than as an actual suggestion for fixing here)

jmcarcell:
> Should this be also changed?
> 
> https://github.com/AIDASoft/podio/blob/14c47b5c030445ffdb1b8d85da56b8289fa43179/cmake/podioMacros.cmake#L248-L252

I think this one is very unlikely to trigger a warning but sure. Maybe in a datamodel using `CLHEP` :confused:

PR #842, opened 2025-09-11
[WIP] Make sure to not consume all available memory on garbage data

tmadlener:
BEGINRELEASENOTES
- Put in an assertion to avoid eating all available memory in case there are problems when reading back data.

ENDRELEASENOTES

Following: https://github.com/AIDASoft/podio/pull/817#issuecomment-3266748609


jmcarcell:
I put the code in https://github.com/AIDASoft/podio/pull/846 and tested the 6 builds right now. I think checking for `s == 0` is better outside not to have to go over every element always.

tmadlener:
Closing this in favor of #846

PR #844, opened 2025-09-12
Refactor schema evolution detection

tmadlener:
BEGINRELEASENOTES
- Split schema evolution detection into two parts; detecting changes and judging whether they can be handled automatically or not

ENDRELEASENOTES

Partial lift of the changes from #828. The changes here do not yet include the change of grammar / format for the evolution files.

The main changes here are the splitting of the `DatamodelComparator` into two bits, the `DatamodelComparator` that detects differences between two datamodels and the `SchemaEvolutionJudge` that then determines whether these changes need manual intervention or not. It's a better split of responsibilities and should make it easier to adapt this whole functionality in the future. Additionally, it removes enough internal state from the whole machinery to make it possible, to re-use it for multiple schema versions.


tmadlener:
Merging this later today, unless someone still wants to have a look and check things.

PR #845, opened 2025-09-12
Change the grammar of the schema evolution files

tmadlener:
BEGINRELEASENOTES
- Change the grammar of the schema evolution files to be more descriptive and also can handle multiple schema evolutions
- **This is a breaking change for code generation if you are using multiple schema and need manual intervention**

ENDRELEASENOTES

Factored out of #828 as a single change

- [x] Includes #844 
- [x] Documentation

### DIFF
--- a/doc/advanced_topics.md
+++ b/doc/advanced_topics.md
@@ -91,7 +91,7 @@ synchronize with any other for file operations).
 
 The podio toolkit supports schema evolution for EDMs that are generated with it.
 it also tries to leverage existing schema evolution functionality of the chosen
-backend if available, e.g. via ROOT. In order to opt-in to schema evolution, it
+backend if available, e.g. via ROOT. In order to opt-in to manual schema evolution, it
 is necessary to pass the old datamodel definition(s) to the code generator, via:
 
 ::::{tab-set}


tests/schema_evolution/code_gen/components_rename_member/evolution.yaml
```diff
@@ -1,8 +1,8 @@
...
-evolutions:
+migrations:
   nsp::TestComponent:
-    member_rename:
-      - y
-      - y_new
+    - from_version: 1
+      to_version: 2
+      change:
+        type: rename_member
```
tmadlener, line 6:
```suggestion
      rename_member:
```
Make the type the key here instead of introducing an extra `change` sub-dict with a dedicated `type` key.

doc/advanced_topics.md
```diff
@@ -86,6 +86,120 @@ The Readers and Writers that ship with podio are assumed to run on a single
...
+### Checking possiblities for schema evolution
+
+The podio toolkit also provides a small script that allows to check up-front
+whether two versions of an EDM need manual intervention as well as confirming
+that an evolution file indeed handles all changes that require manual
+intervention. This can be called as
+
+```bash
+python3 python/podio_schema_evolution.py \
+  datamodel.yaml old_datamodel.yaml
```
jmcarcell: Maybe this should go from older to newer? That is the order followed in every other place in the new documentation.

doc/advanced_topics.md
```diff
@@ -86,6 +86,120 @@ The Readers and Writers that ship with podio are assumed to run on a single
...
+```
+
+:::
+::::
+
+### Supported manual evolutions
+
+| `<change_type>` | description              | `<change_args>`                      |
+|-----------------|--------------------------|--------------------------------------|
+| `rename_member` | Rename a member variable | `{from: <old-name>, to: <new-name>}` |
```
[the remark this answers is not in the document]
tmadlener: But that test effectively just checks that we can still read the file by ignoring the type with the renamed relation. It's a bit unfortunate in where it is placed, but the infrastructure for the other tests makes it pretty straight forward to add it there (and rather hard to add it anywhere else). Renaming a relation would still need some work to actually work as expected.

python/podio_gen/schema_evolution.py
```diff
@@ -0,0 +1,129 @@
...
+            raise ValueError("from_version and to_version must be integers")
+
+        if len(migration_dict) == 0:
+            valid_types = [ct.value for ct in ChangeType]
+            raise ValueError(f"Migration must contain one change type. Valid types: {valid_types}")
+
+        if len(migration_dict) > 1:
+            raise ValueError("Migration can only have one change type")
+
+        (change_type_key, change_details), = migration_dict.items()
```
tmadlener:
```suggestion
        ((change_type_key, change_details),) = migration_dict.items()
```

doc/advanced_topics.md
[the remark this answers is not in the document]
jmcarcell: Sorry I edited after noticing that doesn't work, so I can ask again. How does one find which ones are automatic or which ones are not? There are three categories: manual, automatic or not supported, therefore something not being in the manual list doesn't tell us if it's automatic or not working.

doc/advanced_topics.md
```diff
@@ -86,6 +86,136 @@ The Readers and Writers that ship with podio are assumed to run on a single
 thread only (more precisely we assume that each Reader or Writer doesn't have to
 synchronize with any other for file operations).
 
+
+## Schema evolution
+
+The podio toolkit supports schema evolution for EDMs that are generated with it.
+it also tries to leverage existing schema evolution functionality of the chosen
+backend if available, e.g. via ROOT. In order to opt-in to schema evolution, it
```
jmcarcell:
```suggestion
backend if available, e.g. via ROOT. In order to opt-in to manual schema evolution, it
```

jmcarcell (approved):

PR #846, opened 2025-09-18
Add ugly code to trigger an error when reading garbage

jmcarcell:
BEGINRELEASENOTES
- Add ugly code to trigger an error when reading garbage

ENDRELEASENOTES

This prevents an infinite loop from happening and consuming all the memory. Tested on GCC 15 and Clang 20 with all the combinations of {Debug,RelWithDebInfo,Release} builds, in 5 out of 6 it triggers one of the errors and with GCC in Debug it crashes from ROOT.

I think that `s` is a random 64 bit number (or at least it was big one when I checked with gdb), so the probability of not triggering the error should 1e15/2**64=5e-5 if this is true.

### DIFF
--- a/python/templates/CollectionData.cc.jinja2
+++ b/python/templates/CollectionData.cc.jinja2
@@ -38,6 +38,21 @@
   if (!isSubsetColl) {
     m_data.reset(buffers.dataAsVector<{{ class.full_type }}Data>());
 
+  // The following is ugly code for the case when reading garbage data from ROOT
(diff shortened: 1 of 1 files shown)


tmadlener (commented):
Is there a conceivable way to test that this truly triggers? Not sure how we would realistically get the garbage data, but maybe creating a faulty file with a slightly different EDM and then trying to read it similar as we do for the backwards compatibility tests?

PR #849, opened 2025-09-23
Remove support for CPack

jmcarcell:
BEGINRELEASENOTES
- Same as https://github.com/key4hep/EDM4hep/pull/445, remove it since it is not supported

ENDRELEASENOTES

Out of curiosity this time I tried to build enabling it but it didn't work:

```
CMake Error at podio/cmake/podioCPack.cmake:49 (string):
  string no output variable specified
  podio/cmake/podioCPack.cmake:142 (hsf_get_platform)
  podio/CMakeLists.txt:158 (include)
```


tmadlener:
I suppose it relies on some cmake modules that we have long removed without realizing that it was still used here.

Issue #861, opened 2025-10-14
`RNTupleReader` cannot read datafiles containing datamodels where types have been removed


tmadlener:
Trying to read a file that has been written with an older version of a datamodel where it still contained a datatype which has since been removed is currently not possible with the `RNTupleReader`. This is because we try to create a complete model, here:

https://github.com/AIDASoft/podio/blob/a159df6d3cf17186dc5753d684b54f1340444f2f/src/RNTupleReader.cc#L173

The problem is already the call to `GetModel` in this case.

Quoting from a mattermost discussion:
> RNTupleReader::GetModel will recreate the full model from the on-disk information. If you only need a handful of fields for reading, perhaps a better way would be through RNTuple views ([RNTupleReader::GetView](https://root.cern/doc/master/classROOT_1_1RNTupleReader.html#a6010f3fff28d4fecf2763f4300d36c25)). However, if you really require the full model, RNTupleReader::Open has an overload that allows you to pass an [RCreateModelOptions](https://root.cern/doc/master/structROOT_1_1RNTupleDescriptor_1_1RCreateModelOptions.html). These options have SetEmulateUnknownTypes() which when enabled, will read class fields without a dictionary into untyped record-type fields instead.

PR #862, opened 2025-10-19
Fix reading an RNTuple with a deleted type, add a test to do so

jmcarcell:
See https://github.com/AIDASoft/podio/issues/861

BEGINRELEASENOTES
- Make it possible to read RNTuples with removed types, using `SetEmulateUnknownTypes`
- Add a test that fails without this fix and works with this fix (only RNTuple was failing)

ENDRELEASENOTES

From a comment in the code (https://root.cern/doc/master/structROOT_1_1RNTupleDescriptor_1_1RCreateModelOptions.html#af0c47a78c9fa9cadbaea51ed6b33cd7d), this is the documentation about the flag that `SetEmulateUnknownTypes` sets.

```
If true, fields with a user defined type that have no available dictionaries will be reconstructed as record fields from the on-disk information; otherwise, they will cause an error. 
```

With this, the tests in https://github.com/key4hep/EDM4hep/pull/446 work for me

### DIFF
--- a/src/RNTupleReader.cc
+++ b/src/RNTupleReader.cc
@@ -104,10 +104,14 @@ unsigned RNTupleReader::getEntries(const std::string& name) {
     m_readerEntries[name].push_back(0);
     for (const auto& filename : m_filenames) {
       try {
+#if ROOT_VERSION_CODE >= ROOT_VERSION(6, 36, 0)
         ROOT::RNTupleDescriptor::RCreateModelOptions options;
         // Read unknown types (like deleted ones) without errors
         options.SetEmulateUnknownTypes(true);
         m_readers[name].emplace_back(root_compat::RNTupleReader::Open(options, name, filename));
+#else
+        m_readers[name].emplace_back(root_compat::RNTupleReader::Open(name, filename));
+#endif
         m_readerEntries[name].push_back(m_readerEntries[name].back() + m_readers[name].back()->GetNEntries());
       } catch (const RException&) {
         std::cout << "Category " << name << " not found in file " << filename << std::endl;


tests/schema_evolution/code_gen/datatypes_remove_type/check.cpp
```diff
@@ -0,0 +1,33 @@
...
+    auto coll2 = ExampleHitStaysCollection();
+    auto elem2 = coll2.create();
+    elem2.t(99);
+    event.put(std::move(coll2), "evolution_collection");
+  });
+#ifdef PODIO_SCHEMA_EVOLUTION_TEST_READ
+  auto reader = ReaderT();
+  std::cout << "Reading file " << TEST_CASE FILE_SUFFIX << std::endl;
+  reader.openFile((TEST_CASE FILE_SUFFIX));
+  // TODO: Make this work when reading all collections
```
tmadlener: Where / how does this currently fail? I thought we had the necessary bits and pieces in places also already for the `RNTupleReader` to simply not instantiate / populate the fields we don't want to read (?)

tmadlener: Doesn't work with `ROOTReader` either at the moment, because we either read the full data or none at all. Let's fix this in another PR and move forward with this to unblock EDM4hep#446.

jmcarcell: I thought it was you who found out in https://github.com/key4hep/EDM4hep/pull/446 when only a subset of the collections is read. It certainly doesn't work right now.

src/RNTupleReader.cc
```diff
@@ -104,9 +104,16 @@ unsigned RNTupleReader::getEntries(const std::string& name) {
     m_readerEntries[name].push_back(0);
     for (const auto& filename : m_filenames) {
       try {
+#if ROOT_VERSION_CODE >= ROOT_VERSION(6, 36, 0)
+        ROOT::RNTupleDescriptor::RCreateModelOptions options;
```
tmadlener: This isn't available earlier in some `Experimental` namespace? Or does it in any case not matter because with earlier ROOT versions this does not yet work in any case?

PR #868, opened 2025-10-21
Make Frame::get<T> throw on non-existing collections instead of returning an empty collection

tmadlener:
BEGINRELEASENOTES
- Make `Frame::get<T>` throw on non-existing collections instead of silently returning an empty collection. **Note that this is a behavioral change with respect to previous versions of podio for non-existing collections.** 

ENDRELEASENOTES

This is a first shot to address parts of the discussion in https://github.com/AIDASoft/podio/pull/863#issuecomment-3425511940

---- 
### Original design with policies

Things that warrant discussion IMO:
- Does the policy belong to `get` or should it be set for the Frame on its creation? (If we go with "global" policies, that would affect the `Reader` interface as well, as the policies would have to get to the construction of the Frame somehow).
- Should we already now perceive more policies, e.g. for `Frame::put` where we could also have policies for dealing with collisions.

Also trying to summon @m-fila to see if the concept constraints could be improved or implemented differently somehow.

### DIFF
--- a/tests/schema_evolution/code_gen/datatypes_rename_relation/check.cpp
+++ b/tests/schema_evolution/code_gen/datatypes_rename_relation/check.cpp
@@ -28,7 +28,12 @@ int main() {
     ASSERT_EQUAL(hitColl[0].z(), 3.45, "Readable elements should not change");
     ASSERT_EQUAL(hitColl[0].energy(), 5.67, "Readable elements should not change");
 
-    const auto& clusterColl = event.get<ExampleClusterCollection>("evolution_collection");
-    ASSERT_EQUAL(clusterColl.hasID(), false, "Unreadable collection should be skipped");
+    try {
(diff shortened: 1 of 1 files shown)


doc/examples.md
```diff
@@ -135,8 +135,11 @@ necessary to highlight the change of ownership that happens in this case.
 Collections can be retrieved explicitly:
 
 ```cpp
-    auto& hits = frame.get<HitCollection>("hits");
-    if (hits.hasID()) { ... } // If the collection is not found, an empty one without ID is returned
+    try {
+      auto& hits = frame.get<HitCollection>("hits");
+    } catch (const std::runtime_error&) {
+      // The collection "hits" has not been available as type HitCollection
```
jmcarcell:
```suggestion
      // The collection "hits" does not exist or isn't a HitCollection
```

PR #870, opened 2025-10-23
Pass a unique_ptr to the frame constructor in Python

jmcarcell:
BEGINRELEASENOTES
- Pass a unique_ptr to the frame constructor in Python, making the constructor from any data `FrameData&&` not needed.

ENDRELEASENOTES

First, I'll let CI prove that the constructor is not needed. Then, we can deprecate it if it's not being used anywhere else. The one that takes `unique_ptr` should be preferred I think. `__smartptr__` appears in a tutorial from 2018 in the cppyy docs: https://github.com/wlav/cppyy/blob/master/doc/tutorial/CppyyTutorial.ipynb#L466 so it's probably fine to be used? Otherwise it's possible to do the same with a move and a `cppyy.gbl.unique_ptr` but this may create another `unique_ptr`.

Edit: This works starting with either ROOT 6.36.00 (most likely) or 6.36.02

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
- # podio
+# podio
 
 [![linux](https://github.com/AIDASoft/podio/actions/workflows/test.yml/badge.svg)](https://github.com/AIDASoft/podio/actions/workflows/test.yml)
 [![ubuntu](https://github.com/AIDASoft/podio/actions/workflows/ubuntu.yml/badge.svg)](https://github.com/AIDASoft/podio/actions/workflows/ubuntu.yml)


jmcarcell:
I stopped where I did because adding the include to `Frame.h` feels weird, now it depends on ROOT and logically the include directory of ROOT needs to be added (although it seems to work fine without it in some stacks).

jmcarcell:
I added `PODIO_ROOT_OLDER_6_36` since it is related to the version, then maybe someone finds it when dropping support for ROOT 6.36 in the future?

tmadlener:
It looks like we are loading more libraries again then we should: https://github.com/AIDASoft/podio/actions/runs/19030368924/job/54342600623?pr=870#step:5:1085

tmadlener:
Looks like it might have been somewhat random as it disappeared again. Let's see if this becomes more regular or if this was a one-time thing.

PR #874, opened 2025-11-05
Choose existing function for showing useless state

tmadlener:
hasID does the same thing in this case, but isAvailable is the more appropriate choice, since a default constructed (non-interface) datatype will be available while it will also not have an ID.

BEGINRELEASENOTES
- Use an actually existing member function in documentation of interfaces

ENDRELEASENOTES


jmcarcell (approved):
I think the original won't compile, no? `hasID` only exists for collections, not for objects.

tmadlener:
Ah yes, indeed that as well.

PR #881, opened 2025-11-21
move to tools and rename podio_test_hashes

m-fila:
BEGINRELEASENOTES
- Rename `podio_test_hashes` to `podio-test-hashes` and move to tools directory

ENDRELEASENOTES

This is an executable and probably shouldn't live in `src/` where we library source code is stored.
I'm not sure how "public" is it, do we perhaps want to drop it instead of moving?

- [x] mention in docs

### DIFF
--- a/doc/tools.md
+++ b/doc/tools.md
@@ -107,3 +107,15 @@ Example usage:
 podio-vis --fmt svg my_model.yaml
 firefox gv.svg
 ```
+
+## Testing hash collisions in collection names: `podio-test-hashes`
+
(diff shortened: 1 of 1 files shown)


tmadlener:
> I'm not sure how "public" is it, do we perhaps want to drop it instead of moving?

Essentially public, I would say. I originally wrote it for the check in [this comment](https://github.com/AIDASoft/podio/pull/412#issuecomment-1577050376). IIRC the main reason, I put it into the `src` folder is because `MurmurHash3.h` header is effectively private and I don't think we had any c++ sources yet in `tools`.

Issue #858, opened 2025-10-11
Document basics of subset collections


tmadlener:
Currently subset collections seem to be mostly undocumented (at least in the podio documentation).

PR #885, opened 2025-11-24
Add some documentation about subset collections

jmcarcell:
BEGINRELEASENOTES
- Add some documentation about subset collections

ENDRELEASENOTES

I didn't see a clear place to put it and it is short enough that it doesn't deserve its own section. I don't add an example in examples.md to avoid duplication. The only documentation that seems to exist now is in https://key4hep.github.io/key4hep-doc/main/how-tos/key4hep-tutorials/edm4hep_analysis/edm4hep_api_intro.html#subset-collections.

Close https://github.com/AIDASoft/podio/issues/858.

### DIFF
--- a/doc/examples.md
+++ b/doc/examples.md
@@ -68,6 +68,21 @@ or via direct accessors
 
 If asking for an entry outside bounds, a `std::out_of_range` exception is thrown.
 
+### Subset collections
+Subset collections in podio enable you to create collections whose elements are
+references to objects stored in one or more original collections:
+
+```cpp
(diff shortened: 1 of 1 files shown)


tmadlener:
Thanks a lot. I think it might fit better here: https://github.com/AIDASoft/podio/blob/master/doc/examples.md as *advanced topic* is quite far down the sidebar on the documentation page.

tmadlener:
Fair point. Can we get a short intro and the example into the examples and a link from there to the advanced topics with the limitations?

PR #894, opened 2025-12-01
Make sure to generate Collection headers to beginning

### LINKED ISSUE #892: Undeclared identifiers when including data model headers in cling
We are encountering errors when loading analysis macros when upgrading from podio 1.3 to 1.6 (and keeping root 6.38.00 and EDM4hep 0.99.1 constant). See e.g. https://eicweb.phy.anl.gov/EIC/benchmarks/detector_benchmarks/-/jobs/6875585

We have minimized the issue to the test.C below. This compiles when explicit (`root -l -q -b test.C+`) but not when letting ROOT handle it (no plus sign). Unlikely related to Python.h, I think. Maybe header ordering, but also looks fine to me...

test.C
```cpp
#include "edm4hep/MCParticle.h"

int test(){
  return 0;
}
```

```console
$ root -l -b -q test.C+

Processing test.C+...
Info in <TUnixSystem::ACLiC>: creating shared library /home/wdconinc/./test_C.so
(int) 0
```

```console
$ root -l -b -q test.C 2>&1 | head

Processing test.C...
libedm4hepDict dictionary payload:741:7: error: field has incomplete type '::edm4hep::Vector3f'
      referencePoint{}; ///< Reference point of the track parameters, e.g. the origin at the IP, or the position  of the
      ^
libedm4hepDict dictionary forward declarations' payload:8:26: note: forward declaration of 'edm4hep::Vector3f'
namespace edm4hep{class  Vector3f;}
                         ^
libedm4hepDict dictionary payload:2177:59: error: use of undeclared identifier 'MCParticleObj'
  explicit MutableMCParticle(podio::utils::MaybeSharedPtr<MCParticleObj> obj);
```

The error related to `referencePoint` only shows up in some systems (my local test system) but not in our CI systems linked above. The error for `MCParticleObj` is consistent though.

tmadlener:
Fixes #892 

BEGINRELEASENOTES
- Make sure to generate the Collection headers to the front of the list of header file that is passed to root dictionary generation.

ENDRELEASENOTES

@wdconinc this fixes things for me. Not sure if this points to a more general "instability" in the code generation (e.g. missing headers, or something along those lines) or if this is just a quirk of ROOTs dictionary generation.

I am not entirely sure how to (or at least where) to add a unittest for this. I could not get the original failure reproduced locally within podio and its tests for some reason. We could add it to the downstream tests of EDM4hep, but that would imply some potentially longish and indirect feedback loop.

### DIFF
--- a/python/podio_gen/cpp_generator.py
+++ b/python/podio_gen/cpp_generator.py
@@ -503,9 +503,14 @@ have resolvable schema evolution incompatibilities:"
 
     def _write_cmake_lists_file(self):
         """Write the names of all generated header and src files into cmake lists"""
-        header_files = (f for f in self.generated_files if f.endswith(".h"))
+        header_files = list(f for f in self.generated_files if f.endswith(".h"))
         src_files = (f for f in self.generated_files if f.endswith(".cc"))
         xml_files = (f for f in self.generated_files if f.endswith(".xml"))
+
+        # Sort header files so that Collection headers appear first. This is
+        # necessary for cling to load things in the correct order for some
+        # reason. See https://github.com/AIDASoft/podio/issues/892
+        header_files.sort(key=lambda f: (0 if "Collection.h" in f else 1, f))
 
         def _write_list(name, target_folder, files, comment):
             """Write all files into a cmake variable using the target_folder as path to the


jmcarcell:
Doesn't `clang-format` modify the order?

PR #895, opened 2025-12-02
Fix warnings in the generation of documentation

jmcarcell:
and other things.

Many of the warnings are coming from the myst parser. There was a typo fixed in `conf.py` (`myst_heading_anchor` -> `myst_heading_anchors`), see
https://myst-parser.readthedocs.io/en/latest/syntax/cross-referencing.html#implicit-targets where also it is explained that relying on the implicit conversion is not a good idea and adding anchors is better. Example of warnings about anchors:

```
/podio/doc/collections_as_container.md:7: WARNING: local id not found in doc 'design': 'the-user-layer' [myst.xref_missing]
/podio/doc/collections_as_container.md:7: WARNING: local id not found in doc 'design': 'the-internal-data-layer' [myst.xref_missing]
/podio/doc/examples.md:81: WARNING: local id not found in doc 'advanced_topics': 'subset-collections' [myst.xref_missing]
/podio/doc/links.md:49: WARNING: local id not found in doc 'datamodel_syntax': 'definition-of-links' [myst.xref_missing]
```
Many of those complaining about thing not being at the right level (for example going from `#` to `###` without passing through `##`). There were also many warnings about generated C++ API files not in any TOC, now ignored (should they be in a TOC? These are the files that match `"cpp_api/file/*8h.rst"`).

In `templates.md` many of the paths to files are links but they don't actually take you anywhere, they are not links anymore (there was probably a warning about this).

There were thousands of warnings about C++, now ignored.

A few tabs in `Advanced topics` were not working because `sphinx_design` is needed for that, see: https://myst-parser.readthedocs.io/en/latest/syntax/admonitions.html#other-containers-grids-tabs-cards-etc
This is how it looks after this PR (currently it is empty, after the first paragraph of schema evolution):
2025-12-02-140140_898x163_scrot

After all this, when I build I only find errors like:
```
/podio/doc/cpp_api/class/classpodio_1_1_link_t.rst:4: WARNING: Error when parsing function declaration.
If the function has no return type:
  Error in declarator or parameters-and-qualifiers
  Invalid C++ declaration: Expected identifier in nested name. [error at 46]
    template<typename FromU, typename ToU> inline &&sameTypes< FromU, ToU > operator LinkT< FromU, ToU, false > () const
    ----------------------------------------------^
If the function has a return type:
  Invalid C++ declaration: Expected identifier in nested name. [error at 46]
    template<typename FromU, typename ToU> inline &&sameTypes< FromU, ToU > operator LinkT< FromU, ToU, false > () const
    ----------------------------------------------^
```

BEGINRELEASENOTES
- Fix warnings, tabs, anchors in the documentation

ENDRELEASENOTES

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -57,12 +57,22 @@ source_suffix = {".rst": "restructuredtext", ".md": "markdown"}
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
 # This pattern also affects html_static_path and html_extra_path.
-exclude_patterns = ["_build", "Thumbs.db", ".DS_Store", "cpp_api.rst", "py_api.rst", "ReleaseNotes.md", "contributing.md", "cpp_api/filelist.rst", "cpp_api/file/*8h.rst"]
+exclude_patterns = [
+    "_build",
+    "Thumbs.db",
+    ".DS_Store",
+    "cpp_api.rst",
+    "py_api.rst",
+    "ReleaseNotes.md",
+    "contributing.md",
+    "cpp_api/filelist.rst",
+    "cpp_api/file/*8h.rst",
+]
 
 # Suppress noisy warnings from generated C++ API docs
 suppress_warnings = [
     "duplicate_declaration.cpp",  # duplicate C++ declarations from Breathe
-    "cpp",                        # general C++ domain parse warnings
+    "cpp",  # general C++ domain parse warnings
 ]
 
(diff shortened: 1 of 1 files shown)


tmadlener:
Looks like we need `sphinx_design` to be in the requirements:

https://github.com/AIDASoft/podio/actions/runs/19937177498/job/57165194952#step:4:1051

PR #896, opened 2025-12-03
Fix rpaths in podio, not set correctly because ${LIBDIR} is empty

jmcarcell:
BEGINRELEASENOTES
- Fix rpaths in podio, not set correctly because ${LIBDIR} is empty
- Remove the linker flags that force to use `RPATH` instead of the more modern and recommended `RUNPATH` (the difference being `RPATH` is searched before `LD_LIBRARY_PATH` and `RUNPATH` is searched after `LD_LIBRARY_PATH`).

ENDRELEASENOTES

I have one build where I don't have podio in LD_LIBRARY_PATH and what I get from the library folder (before this PR) is:

```
$ ldd libpodioIO.so
...
	libpodioRootIO.so => not found
	libpodio.so => not found
...

```
and

```
readelf -d libpodio.so | grep RPATH
 0x000000000000000f (RPATH)              Library rpath: [/podio/install/]
```
which is pointing to the installation prefix, and not to the library folder.

This is probably not noticed because everyone has podio under LD_LIBRARY_PATH. I don't change the MacOS stuff since I have not tested it, if everything is installed to /lib it should probably work.

- This says `RPATH` is deprecated: https://libc-help.sourceware.narkive.com/n4inVNdm/deprecated-status-of-rpath (written 14 years ago) but I can't find that in `man ld`.
- See the source of truth of RPATH handling: https://gitlab.kitware.com/cmake/community/-/wikis/doc/cmake/RPATH-handling#always-full-rpath

### DIFF
--- a/cmake/podioBuild.cmake
+++ b/cmake/podioBuild.cmake
@@ -23,7 +23,7 @@ macro(podio_set_rpath)
   elseif(PODIO_SET_RPATH)
     set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}")
     # the RPATH to be used when installing, but only if it's not a system directory
-    list(FIND CMAKE_PLATFORM_IMPLICIT_LINK_DIRECTORIES "${CMAKE_INSTALL_PREFIX}/lib" isSystemDir)
+    list(FIND CMAKE_PLATFORM_IMPLICIT_LINK_DIRECTORIES "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}" isSystemDir)
     if("${isSystemDir}" STREQUAL "-1")
       set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}")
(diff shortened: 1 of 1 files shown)


jmcarcell:
> I think this is the right thing to do. IIUC, this now makes things work by default. But if a user still wants to use their own version of podio in a Key4hep environment they would be able to pull that in by setting the correct `LD_LIBRARY_PATH`, right?

Nothing changes in this case since the local version will point to the local libraries with `RUNPATH` or `LD_LIBRARY_PATH`.

> At least if all of the downstream dependencies are also built with using `RUNPATH` instead of `RPATH` (which I am not sure we actually do, since at least spack is rather aggressive in injecting `RPATH` into cmake builds, I think).

I think this is mostly irrelevant since one doesn't typically use dependencies built with the non-local version of the current package. In any case, `RUNPATH` would allow you to overwrite the existing paths to point to the local one allowing precisely this, which is why it is preferred as it is more flexible than `RPATH`. For spack, indeed it's `RPATH`:
```
readelf -d libedm4hep.so | grep RPATH
 0x000000000000000f (RPATH)              Library rpath: [/cvmfs/sw.hsf.org/contrib/x86_64-almalinux9-gcc11.4.1-opt/gcc/14.2.0-yuyjov/lib64...
```

PR #897, opened 2025-12-03
Make the python writers have the same behavior as the C++ writers

jmcarcell:
BEGINRELEASENOTES
- Make the python writers have the same behavior as the C++ writers to decide what to do when no collections are passed

ENDRELEASENOTES

If no collections are passed then let the C++ decide instead of passing all collections (which is what the C++ writers do). In practice, this only allows to write frames with no collections from python, currently not possible as collections will evaluate to false and all the collections from `frame.getAvailableCollections` will be passed.


m-fila:
Should this have a test to check this behaviour?

Perhaps in longer future python reading and writing could be rewritten to use C++ (type-erased) readers/writers with less wrappers?

PR #905, opened 2025-12-17
Add CMake presets and start to use them in CI

tmadlener:
BEGINRELEASENOTES
- Add a few [cmake presets](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html) to put common build options together. Use `cmake --list-presets` to see the available presets.
- Use presets in CI workflows to reduce some of the duplication that is currently present

ENDRELEASENOTES

- [x] assumes that #902 is already in place as it also uses that cache variable

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,18 +91,37 @@ and `env.sh`.
 
 ## Compiling
 
-If you are using the easy setup from `init.sh` then create separate build
-and install areas, and trigger the build:
+Podio uses CMake presets to simplify the build process. The following presets are available:
 
-    mkdir build
-    mkdir install
-    cd build
-    cmake -DCMAKE_INSTALL_PREFIX=../install ..
-    make -j 4 install
+- `base-build`: Base build with RNTuple backend and DataSource support
+- `full-build`: Full build including SIO and schema evolution tests
+- `ci-build`: Full build with warnings as errors (used for continuous integration)
 
-To see a list of options, do this in the build-directory:
+### Basic build
 
-    cmake -LH ..
+To build podio with the base configuration:
+
+    cmake --preset base-build -DCMAKE_INSTALL_PREFIX=install
+    cmake --build build -j 4
+    cmake --install build
+
+### Full build
+
+For a complete build with all features:
+
+    cmake --preset full-build -DCMAKE_INSTALL_PREFIX=install
+    cmake --build build -j 4
+    cmake --install build
+
+### Custom configuration
+
+To see available CMake options:
+
+    cmake --preset base-build -LH
+
+You can override preset options by adding `-D` flags:
+
+    cmake --preset base-build -DCMAKE_INSTALL_PREFIX=install -DENABLE_SIO=ON
 
 ## Running tests
 


tmadlener:
I have looked into workflow presets a bit and it looks like we cannot pass additional options via `-D` with `cmake --workflow --preset`. See also: https://discourse.cmake.org/t/workflow-usage/8288/9

At least at the moment we use those quite a bit for some configuration on top of the `ci-build` preset.

.github/workflows/test.yml
```diff
@@ -34,33 +34,25 @@ jobs:
...
             -DPODIO_RUN_STRACE_TEST=$([[ ${{ matrix.LCG }} == LCG_104/* ]] && echo "OFF" || echo "ON") \
             -DCMAKE_INSTALL_PREFIX=../install \
-            -DCMAKE_CXX_STANDARD=$([[ ${{ matrix.LCG }} == *-gcc15-* ]] && echo "23" || echo "20") \
-            -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
-            -DCMAKE_CXX_FLAGS=" -fdiagnostics-color=always -Werror -Wno-error=deprecated-declarations " \
             -DUSE_EXTERNAL_CATCH2=AUTO \
-            -DPODIO_USE_CLANG_FORMAT=AUTO \
-            -DPODIO_ENABLE_SCHEMA_EVOLUTION_TESTS=ON \
-            -G Ninja ..
+            -DPODIO_USE_CLANG_FORMAT=AUTO
```
tmadlener: This is already configured like this in the `ci-build` configure preset.

CMakePresets.json
```diff
@@ -0,0 +1,73 @@
...
+                "PODIO_SET_RPATH": "ON"
+            }
+        },
+        {
+            "name": "base-build",
+            "displayName": "Base Build including RNTuple backend and DataSource",
+            "inherits": "common",
+            "generator": "Ninja",
+            "cacheVariables": {
+                "BUILD_TESTING": "ON",
```
tmadlener: Should this be `OFF`?

CMakePresets.json
```diff
@@ -0,0 +1,73 @@
...
+                "CMAKE_CXX_COMPILER_LAUNCHER": "ccache"
+            }
+        },
+        {
+            "name": "full-build",
+            "displayName": "Full build including schema evolution tests and SIO",
+            "inherits": "base-build",
+            "cacheVariables": {
+                "ENABLE_SIO": "ON",
+                "PODIO_ENABLE_SCHEMA_EVOLUTION_TESTS": "ON"
```
tmadlener: This probably should be `OFF` in the `base-build` now that #902 changed defaults.

.github/workflows/sanitizers.yaml
```diff
@@ -41,23 +41,15 @@ jobs:
...
-              -DENABLE_JULIA=OFF \
-              -DENABLE_RNTUPLE=ON \
-              -DENABLE_DATASOURCE=ON \
-              -DPODIO_ENABLE_SCHEMA_EVOLUTION_TESTS=ON \
-              -G Ninja ..
+              -DCMAKE_CXX_STANDARD=$([[ ${{ matrix.compiler }} == gcc15 ]] && echo "23" || echo "20")
             echo "::endgroup::"
             echo "::group::Build"
-            ninja -k0
+            cmake --build build -- -k0
```
jmcarcell:
```suggestion
            cmake --build build
```
`-k0` is in the preset?

CMakePresets.json
tmadlener: No. Let's keep building the tests even in the `base-build`.

.github/workflows/edm4hep.yaml
```diff
@@ -42,43 +42,34 @@ jobs:
...
-              -DPODIO_ENABLE_SCHEMA_EVOLUTION_TESTS=ON \
-              -G Ninja ..
-            ninja -k0
+              -DPODIO_RUN_STRACE_TEST=ON
+            cmake --build --preset ci-build
             echo "::endgroup::"
             echo "::group::Test and install podio"
-            ctest --output-on-failure --schedule-random -j$(nproc) --timeout 600
-            ninja install
+            ctest --preset ci-build -j$(nproc)
```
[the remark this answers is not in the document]
tmadlener: Changed  to use `execute:jobs:0` in the `ci-build` preset. Should default to native parallelism with new enough cmake versions and to `-j1` for older versions.

PR #906, opened 2025-12-19
podio-dump: Make -e -1 dump all of the entries

tmadlener:
Many of our other tools support passing -1 as a value to the number of events signifying to simply do all of them.

BEGINRELEASENOTES
- `podio-dump`: Make `-e`/`--entries` accept `-1` to dump all of the entries of a given category

ENDRELEASENOTES

### DIFF
--- a/tests/CTestCustom.cmake
+++ b/tests/CTestCustom.cmake
@@ -158,6 +158,7 @@ if ((NOT "@FORCE_RUN_ALL_TESTS@" STREQUAL "ON") AND (NOT "@USE_SANITIZER@" STREQ
 
       podio-dump-root
       podio-dump-detailed-root
+      podio-dump-all-events
 
       write_interface_root
       read_interface_root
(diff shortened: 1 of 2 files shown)


tools/src/podio-dump-tool.cpp
```diff
@@ -276,7 +289,14 @@ int main(int argc, char* argv[]) {
     stats = reader.getSizeStats(args.category);
   }
 
-  for (const auto event : args.events) {
+  const auto entryList = [&args, &reader]() {
+    if (!args.events.empty() && args.events[0] == static_cast<std::size_t>(-1)) {
```
tmadlener:
```suggestion
    if (!args.events.size() == 1 && args.events[0] == static_cast<std::size_t>(-1)) {
```

tools/src/podio-dump-tool.cpp
```diff
@@ -276,7 +289,14 @@ int main(int argc, char* argv[]) {
     stats = reader.getSizeStats(args.category);
   }
 
-  for (const auto event : args.events) {
+  const auto entryList = [&args, &reader]() {
+    if (!args.events.size() == 1 && args.events[0] == static_cast<std::size_t>(-1)) {
```
tmadlener:
```suggestion
    if (args.events.size() != 1 && args.events[0] == static_cast<std::size_t>(-1)) {
```

PR #908, opened 2025-12-22
feat: move Pythonizations implementations to source files

wdconinc:
This PR moves the implementation of Pythonizations detail functions into a source file. This allows the headers to be included in C++20 modules (#907).

BEGINRELEASENOTES
- Move Pythonizations detail implementations to separate source file

ENDRELEASENOTES

### DIFF
--- a/src/Pythonizations.cc
+++ b/src/Pythonizations.cc
@@ -1,7 +1,7 @@
 #include "podio/detail/Pythonizations.h"
 
-#include <stdexcept>
 #include <cstring>
+#include <stdexcept>
 
 namespace podio::detail::pythonizations {
 


src/CMakeLists.txt
```diff
@@ -79,6 +80,7 @@ SET(core_headers
 
 PODIO_ADD_LIB_AND_DICT(podio "${core_headers}" "${core_sources}" selection.xml)
 target_compile_options(podio PRIVATE -pthread)
+target_link_libraries(podio PRIVATE Python3::Python)
```
wdconinc: Ah, this probably should have been in the `Python::` namespace since that's what is added as a dependency in the config
https://github.com/AIDASoft/podio/blob/5d827486312154a0b6e2cb095ec28371d6ee5220/cmake/podioConfig.cmake.in#L28-L32
and in the `PODIO_GENERATE_DATAMODEL` module:
https://github.com/AIDASoft/podio/blob/5d827486312154a0b6e2cb095ec28371d6ee5220/cmake/podioMacros.cmake#L158

Issue #918, opened 2026-01-12
Add pythonisation for LinkNavigator


dudarboh:
As the title says.
It would be nice to have this feature in python as well.
As, otherwise one needs to manually loop over link collections

```python
import podio
podio.LinkNavigator
AttributeError: module 'podio' has no attribute 'LinkNavigator'
```

jmcarcell:
There is a way:
``` python
In [1]: import ROOT

In [2]: ROOT.gInterpreter.LoadFile("podio/LinkNavigator.h")
Out[2]: 0

In [3]: from ROOT import podio

In [4]: podio.LinkNavigator
Out[4]: <cppyy.Template 'podio::LinkNavigator' object at 0x7f61c8183c50>
```

PR #919, opened 2026-01-13
Improve usability of LinkNavigator from python

tmadlener:
Attempt for checking whether this is enough to get some parts of the LinkNavigator work more easily from the python side

BEGINRELEASENOTES
- Add dedicated overloads for `Mutable` handles to `LinkNavigator::getLinked` to make python bindings also work with `Mutable` handles.

ENDRELEASENOTES

See #918 for some more discussion and insights.

@dudarboh this currently adds the "easy" part of more overloads that should hopefully help with calling `getLinked` with `Mutable` handles. It doesn't yet add any other python wrapping for improving the other cumbersome parts that you found.

### DIFF
--- a/include/podio/LinkNavigator.h
+++ b/include/podio/LinkNavigator.h
@@ -33,12 +33,20 @@ namespace detail::links {
   struct ReturnToTag {};
 } // namespace detail::links
 
+#if !defined(__CLING__)
+  #define INLCONSTEXPR inline constexpr
+#else
+  #define INLCONSTEXPR
+#endif
 /// Tag variable to select the lookup of *From* objects have links with a *To*
 /// object in podio::LinkNavigator::getLinked
-detail::links::ReturnFromTag ReturnFrom;
+// INLCONSTEXPR detail::links::ReturnFromTag ReturnFrom;
+INLCONSTEXPR detail::links::ReturnFromTag ReturnFrom;
 /// Tag variable to select the lookup of *To* objects that have links with a
 /// *From* object in podio::LinkNavigator::getLinked
-detail::links::ReturnToTag ReturnTo;
+// INLCONSTEXPR detail::links::ReturnToTag ReturnTo;
+INLCONSTEXPR detail::links::ReturnToTag ReturnTo;
+#undef INLCONSTEXPR
 
 /// A helper class to more easily handle one-to-many links.
 ///


tmadlener:
We should add also the python wrapping parts here to make it really easy to use.

tmadlener:
I made Claude come up with some "python bindings" and this looks like a reasonable solution to me. The one thing that doesn't yet transpire through the whole cppyy machinery is the capabilities of structured bindings for the `WeightedObject`, i.e. in c++ I can do

https://github.com/AIDASoft/podio/blob/2a694815e3dda61ade771df7aefa46aa71c5b5ab/tests/unittests/links.cpp#L665

But trying to do the same in python currently fails with 

```
  File "/home/madlener/work/AIDASoft/podio/python/podio/test_LinkNavigator.py", line 85, in test_multiple_links
    for c, w in linked:
        ^^^^
TypeError: WeightedObject<ExampleCluster> object does not support indexing
```

tmadlener:
Thanks :)

@dudarboh do you want to give this another go before we merge to see if it indeed fixes the issues you observed?

include/podio/LinkNavigator.h
```diff
@@ -35,10 +35,10 @@ namespace detail::links {
 
 /// Tag variable to select the lookup of *From* objects have links with a *To*
 /// object in podio::LinkNavigator::getLinked
-inline constexpr detail::links::ReturnFromTag ReturnFrom;
+detail::links::ReturnFromTag ReturnFrom;
```
jmcarcell: There is a problem with this, if included in several translation units, then it won't link since the symbol will be defined several times. Both inline or constexpr trigger the issue of opening all the libraries (just checked).

tmadlener: Thanks for having a look and pushing the changes. Seems like they are working.

One of the key use cases for using the `LinkNavigator` is (IMO) precisely the fact that people don't have to remember which type is `FromT` and which type is `ToT`, so having a `getLinked` that just does the right thing is what we want, I think.

However, looking at the code, it looks like we could still get rid of the tag types and just introduce the `getLinkedTo` and `getLinkedFrom` and keep all the existing functionality because that is what is currently there in effect in any case. I will put that into another PR with a proper deprecation warning.

PR #922, opened 2026-01-14
Make link collection iterators behave consistently with generated collection iterators

tmadlener:
original outset of this PR: ~non-const collections give access to mutable handles via operator[] and at(). Hence, they should also model ranges over mutable handles. The fix is straight forward: Make sure the iterators report the correct value_type.~

See the discussion below for why we now make the link collections behave the same as the generated ones.

BEGINRELEASENOTES
- Make the `LinkCollectionIteratorT::value_type` an immutable `Link` always for consistency with the behavior of generated collection iterators. 

ENDRELEASENOTES

@m-fila am I missing something obvious here?

### DIFF
--- a/include/podio/utilities/TypeHelpers.h
+++ b/include/podio/utilities/TypeHelpers.h
@@ -303,6 +303,9 @@ concept CollectionType = !std::is_abstract_v<T> && std::derived_from<T, Collecti
                             typename T::mutable_type>;
       requires std::same_as<std::remove_cvref_t<decltype(ct.at(std::declval<typename T::size_type>()))>,
                             typename T::value_type>;
+      requires std::same_as<typename T::value_type, typename std::iterator_traits<typename T::iterator>::value_type>;
+      requires std::same_as<typename T::value_type,
+                            typename std::iterator_traits<typename T::const_iterator>::value_type>;
     };
 
 namespace utils {


tests/unittests/std_interoperability.cpp
```diff
@@ -309,7 +309,7 @@ TEST_CASE("Collection container types", "[collection][container][types][std]") {
   DOCUMENTED_STATIC_FAILURE(std::is_convertible_v<CollectionType::iterator, CollectionType::const_iterator>);
   STATIC_REQUIRE(traits::has_value_type_v<CollectionType::iterator>);
   STATIC_REQUIRE(
-      std::is_same_v<CollectionType::value_type, std::iterator_traits<CollectionType::iterator>::value_type>);
+      std::is_same_v<CollectionType::mutable_type, std::iterator_traits<CollectionType::iterator>::value_type>);
```
m-fila: This is a check if we fulfill the requirement of ["Container"](https://en.cppreference.com/w/cpp/named_req/Container.html) saying that the `iterator::value_type` should be the same as `value_type` of container. The standard doesn't mention anything about `mutable_type` ;)

`value_type` should stay as was but `STATIC_REQUIRE` should change to`DOCUMENTED_STATIC_FAILURE` instead.

tmadlener:
Right, so some nuances still here. We don't really change any behavior that allows to break things that weren't already broken before, I think.

Just to confirm, the main thing that now is possible that wasn't before is this:

```cpp
ExampleClusterCollection coll{};
for (auto& elem : coll) {
  coll.setEnergy(42);  // not possible before because elem would not have a setter
}
```
One would have to do this (which is also already possible now)
```cpp
for (size_t i = 0; i < coll.size(); ++i) {
  coll[i].setEnergy(42);  // possible because operator[] returns a mutable handle on non-const collections
}
```

What is broken regardless of these changes is assigning a new handle to the element, I think. Or does that also change with this change? (Do we have a test for that that I simply missed?).


Anyhow, the main reason I wanted to have this feature, is because it makes it behave the same as the `LinkCollection`, where we do:

https://github.com/AIDASoft/podio/blob/35e65ce995180199531336c4e4fb75a010641748/include/podio/detail/LinkCollectionIterator.h#L11-L15

(i.e. we also have the `value_type` be different depending on the iterator type). We should make sure this is consistent. I technically don't need this feature, it just makes for a nice example in the tests for #923.

m-fila:
> Just to confirm, the main thing that now is possible that wasn't before is this:
> 
> ```c++
> ExampleClusterCollection coll{};
> for (auto& elem : coll) {
>   coll.setEnergy(42);  // not possible before because elem would not have a setter
> }
> ```

Does it compile now? I thought it still shouldn't since `operator*` for iterator returns temporary mutable dataobject which can't be bound to non-const l-value reference. (unlike for something like `std::vector::iterator::operator*` which returns real l-value reference)
Using value `for (auto elem : coll)` or forwarding reference `for (auto&& elem : coll)` works now and before

>  What is broken regardless of these changes is assigning a new handle to the element, I think. Or does that also change with this change? (Do we have a test for that that I simply missed?).

This is a separate problem that isn't affected by this PR. Dataobject assignment changes which internal object is held instead of changing members of internal object. For temporaries that was very confusing and made no sense so we disabled it in #743. 

> (i.e. we also have the value_type be different depending on the iterator type). We should make sure this is consistent. I technically don't need this feature, it just makes for a nice example in the tests for https://github.com/AIDASoft/podio/pull/923.

We could also make it consistent in the other direction: have both immutable and mutable iterators to link collection use immutable type as `value_type`. I guess the choice is whether meta-programming becomes easier (in #922) or we want to follow some of the conventions from "Container" even though we can't fulfill all of them (and it seems that actual we have so far no code that really depends whether we follow them or not)

Issue #778, opened 2025-05-11
Intra-frame parallelism and reproducible ordering


wdconinc:
In EIC, we tend to think of collections as sets rather than lists/vectors, with an ordering that's merely a result of the order in which they are filled but cannot be inherently guaranteed since it is not defined in the data model. In our event reconstruction code development we have some infrastructure to identify changes in the outputs between feature and target branch artifacts, and that is of course sensitive to changes in ordering. Due to concurrency in the framework, frames are written out of order, which is of course expected. However, we are also exploring intra-frame parallelism, which leads to vector members and collections being filled in an order that's not necessarily sequential or therefore reproducible (think: `std::for_each(std::execution::unseq_par, hits.begin(), hits.end(), [&c](auto& h){ c.add ToHits(h); })` for hits in a cluster). Is there any way to fill podio vector members and collections out of sequence, yet still end up with a reproducible ordering on disk? I'm thinking orderable data types would need a hash or compare function to define the ordering, manifested in the data model. That could then be used to fill an ordered container, which would be serialized before saving to file.

tmadlener:
> `std::for_each(std::execution::unseq_par, hits.begin(), hits.end(), [&c](auto& h){ c.add ToHits(h); })`

I am not sure this is even possible at the moment. IIRC all the member functions that add relations / vector members are not thread-safe at the moment:

https://github.com/AIDASoft/podio/blob/db109c87bb510d56234de96fcf447af01a63cb1c/python/templates/macros/implementations.jinja2#L138-L141

Where `m_obj->m_{{ relation.name }}` is a `std::vector` and `m_obj->data.{{ relation.name }}_end` is an `unsigned int`, neither of which are protected in any form via a mutex or the use of an atomic.

In general the philosophy we had in mind for intra-event parallelism was that as long as collections are being filled, resp. they are still mutable they live on a single thread only. This is partially inspired by the requirements of Gaudi Functional with a similar philosophy and also because of trying to avoid some of the issues you have described, as now each collection is filled deterministically, even if the collections themselves are potentially not. The latter is (as you know) dealt with by hashing the collection names to get a collection id when putting them into the frame (Watch out for issues similar to https://github.com/key4hep/k4FWCore/issues/311 here in case this is not done immediately).

[...]

PR #923, opened 2026-01-14
Make it possible to construct collections from ranges of handles

tmadlener:
BEGINRELEASENOTES
- Add a `from` factory function to create collections from ranges of (mutable or immutable) handles

ENDRELEASENOTES

- [x] Includes #922 
- [x] Proper documentation (once it's decided that this is useful)

See discussion in #778 for some motivation of this. 

@wdconinc this is a first attempt at addressing at least part of the issue discussed in #778. It allows to construct collections from ranges (satisfying `std::input_range`) of handles, e.g.

```cpp
std::vector<edm4hep::MCParticles> particles; // from wherever
auto mcParticles = edm4hep::MCParticleCollection::from(particles);
```

given that our collections also model a `std::input_range`, it's also possible to construct from them using this approach.

Open questions currently are:
- Does this go into the right direction for your use case?
- Currently ranges of mutable handles will yield "standard" collections, while ranges of immutable handles will yield subset collections. There is currently no way to change this, and trying to do so will yield an `invalid_argument` exception. Is this a problem?

### DIFF
--- a/doc/examples.md
+++ b/doc/examples.md
@@ -130,6 +130,13 @@ normal (owned) collection or a subset collection based on the element type:
 - **Ranges of mutable handles** → Creates a normal collection that owns the objects
 - **Ranges of immutable handles** → Creates a subset collection (non-owning view)
 
+::::{warning}
+Depending on the ownership of the handles in the input range it is possible that
+this constructor throws as it uses `push_back` internally. Concretely, mutable
+handles must not be owned by another collection already and immutable handles
+must be owned by another collection already.
+::::
+
 ### Range Pipeline Integration (C++23)
 
 When compiled with C++23 ranges support, collections support the standard


include/podio/detail/LinkCollectionImpl.h
```diff
@@ -74,6 +76,40 @@ class LinkCollection : public podio::CollectionBase {
...
+    LinkCollection coll;
+    if constexpr (detail::RangeOf<R, value_type>) {
+      coll.setSubsetCollection();
+    }
+
+    // NOTE: We cannot use ranges::copy here, because that will always resolve
+    // to only one of the push_back overloads, specifically the one taking
+    // value_type. See, e.g.: https://godbolt.org/z/seqdh79rr
+    // Additionally, we want no reference in the loop variable to make sure we
+    // get the correct handle type depending on the range we pass in
```
m-fila: This is kinda expected taking into account that `std::back_insert_iterator` assignment takes elements byreferences to container's `value_type`, hence we'd get a conversion to immutable type. https://en.cppreference.com/w/cpp/iterator/back_insert_iterator/operator=.html

include/podio/utilities/TypeHelpers.h
```diff
@@ -240,6 +241,57 @@ namespace detail {
   template <typename T, typename U>
   inline constexpr bool isInterfaceInitializableFrom = InterfaceInitializerHelper<T, U>::value;
 
+  /// A simple check for whether a range R is exactly a range over type T
+  template <typename R, typename T>
+  concept RangeOf = std::ranges::input_range<R> && std::same_as<std::ranges::range_value_t<R>, T>;
```
m-fila:
```suggestion
  concept RangeOf = std::ranges::input_range<R> && std::same_as<std::decay_t<std::ranges::range_ref_t<R>>, T>;
```

Should this be `range_ref_t` (type returned by dereference?), and decayed to remove `const` and `&`?

m-fila:
> Good point. I agree with not having a constructor simply taking a range. Should we already now add that constructor and hide it behind [`#if defined(__cpp_lib_containers_ranges)`](https://en.cppreference.com/w/cpp/experimental/feature_test.html#cpp_lib_containers_ranges) in case we decide to go forward with this approach here?

Yes, let's to that

include/podio/utilities/TypeHelpers.h
```diff
@@ -240,6 +241,32 @@ namespace detail {
...
+
+  /// A simple check for whether a range R is a range of a type that can convert to type T
+  template <typename R, typename T>
+  concept RangeConvertibleTo = std::ranges::input_range<R> && std::convertible_to<std::ranges::range_value_t<R>, T>;
+
+#if defined(__cpp_lib_ranges_to_container)
+  template <typename T, std::ranges::input_range R>
+  auto to_vector(R&& r) {
```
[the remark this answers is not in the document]
jmcarcell: Whoops I deleted before you replied 😂 
Yeah I think it's a bit complicated, maybe one can keep doing it manually or with `std::ranges::to` once it's possible.

PR #925, opened 2026-01-19
Link against `Python3::Module`

paulgessinger:
instead of `Python3::Python`, which according to the documentation is meant for embedding the python interpreter.

BEGINRELEASENOTES
- ...

ENDRELEASENOTES


paulgessinger:
@jmcarcell `Python3::Python` hard-adds `-lintl` in the compiler flags and doesn't propagate correctly in spack, at least on macOS. That breaks the macOS build unfortunately. This change works locally for me on macOS, but clearly has issues in the linux builds. Seems like the spack build produces very different python libraries depending on the platform.

PR #928, opened 2026-01-19
Make pythonization for collection subscript header only again

tmadlener:
BEGINRELEASENOTES
- Make pythonizations for collection subscript header only again to work around a possible issue with linking on macOS

ENDRELEASENOTES

Inspired by #925 and https://github.com/AIDASoft/podio/pull/908#issuecomment-3768945487

It looks like this should keep things building / linking for macOS, while still maintaining the possibility to create c++ modules with it.

@paulgessinger any chance you could give this a go and see if this also works for you? 
@wdconinc Should I rebase #907 or will you do it?

In case this works for macOS, I would merge this and tag v01-07-01 afterwards and add a dedicated `conflict` to the `podio` spack package for mac for `1.7`.


wdconinc:
> Should I rebase #907 or will you do it?

I can rebase it.

tmadlener:
Yeah, also a good point. Since making python an explicit `link` dependency as well here:

https://github.com/spack/spack-packages/blob/8913ea5052d65511819246a6038e4081e133801d/repos/spack_repo/builtin/packages/podio/package.py#L80

seems to fix thins also on macOS, we could keep the current version.

I kind of like the fact that the header-only approach has no traces in the binaries while still working with `cppyy` though. Not sure about how large the impact for compilation is in the end.

Issue #920, opened 2026-01-13
`podio-dump` output of MCParticle table can be improved


dudarboh:
The current output of the MCParticles table from `podio-dump -d` is extremely hard to read compared to the formatting from `dumpevent`.

Here are things that can be improved:
- header alignment with the columns content
- column values formatting
- simulatorStatus should print readable simulation flags [sbvtcls] instead of the integer encoding
- id columns, and ids in "parents" and "daughters" columns should not contain an extra "a1cba250|+" string identical for all entries.
- parents and daughters should be part of the table as separate columns rather than creating independent new lines.
Side to side comparison:

Output from `podio-dump -d`
```
MCParticles
          id:         PDG:generatorStatus:simulatorStatus:      charge:        time:        mass:                   vertex [ x, y, z]:                 endpoint [ x, y, z]:                 momentum [ x, y, z]:       momentumAtEndpoint [ x, y, z]:    helicity:
a1cba250|+0          +11           +3           +0 -1.000000e+00 +0.000000e+00 +0.000000e+00 +0.000000e+00 +0.000000e+00 +0.000000e+00  +0.000000e+00 +0.000000e+00 +0.000000e+00  +0.000000e+00 +0.000000e+00 +4.550000e+01  +0.000000e+00 +0.000000e+00 +0.000000e+00            +0 
      parents : 
      daughters : a1cba250|+2 a1cba250|+3 
a1cba250|+1          -11           +3           +0 +1.000000e+00 +0.000000e+00 +0.000000e+00 +0.000000e+00 +0.000000e+00 +0.000000e+00  +0.000000e+00 +0.000000e+00 +0.000000e+00  +0.000000e+00 +0.000000e+00 -4.550000e+01  +0.000000e+00 +0.000000e+00 +0.000000e+00            +0 
      parents : 
      daughters : a1cba250|+2 a1cba250|+3 
a1cba250|+2           -3           +2           +0 +3.333333e-01 +0.000000e+00 +6.743496e-07 +0.000000e+00 +0.000000e+00 +0.000000e+00  +0.000000e+00 +0.000000e+00 +0.000000e+00  +2.625073e+01 +3.538229e+01 -1.136850e+01  +0.000000e+00 +0.000000e+00 +0.000000e+00            +0 
      parents : a1cba250|+0 a1cba250|+1 
      daughters : a1cba250|+4 
[...]

PR #931, opened 2026-02-12
Make it possible to use objects and collections with fmtlib

tmadlener:
BEGINRELEASENOTES
- Make it possible to use objects and collections with fmtlib by using the `ostream_formatter` adaptor from fmtlib.
- **`fmtlib` is now a transitive dependency of podio**, i.e. EDMs generated with podio will also depend on it. The necessary propagation of this dependency via CMake is handled here, but build environments might need adjustments.

ENDRELEASENOTES

This is an intermediate step towards better support for using `fmt::format` (and eventually hopefully also `std::format`). Currently this simply re-uses the existing `operator<<`. In a next step I will implement proper formatting via the customization points that `fmtlib` provides. Then `operator<<` will be implemented in terms of `fmt::format`. Ideally, that step would be compatible with `<format>` in which case we might be able do drop the `fmtlib` dependency.

See also the discussion in https://github.com/AIDASoft/podio/issues/920#issuecomment-3777230323

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -172,6 +172,8 @@ configure_file(${CMAKE_CURRENT_SOURCE_DIR}/python/__version__.py.in
 install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/LICENSE
         DESTINATION ${CMAKE_INSTALL_DOCDIR})
 
+find_package(fmt 9 REQUIRED)
+
 #--- project specific subdirectories -------------------------------------------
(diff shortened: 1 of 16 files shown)


m-fila:
>Is that possible on a technical level? 🤔

I thought it could be done by adapting the contexts, but couldn't find anyone actually doing this, so perhaps it's not the greatest idea. Never mind

jmcarcell (approved):

PR #933, opened 2026-02-13
[WIP] Add support for customizable formatting

tmadlener:
BEGINRELEASENOTES
- Make it possible to specify a user defined formatting via the `{u}` specifier.

ENDRELEASENOTES

- [ ] Includes #932 
- [ ] Includes #931 
- [ ] Do we need more than one customization hook? (E.g. for EDM4hep to define a better "default" formatting, but still giving the users the possibility to override)
- [ ] Test integration into EDM4hep
- [ ] Better implementation e.g. via class specialization (similar to `fmt::formatter`)?

This follows roughly the approach outlined in https://github.com/AIDASoft/podio/issues/920#issuecomment-3777230323 and is the final piece of the chain towards customizable collection formatting.


tmadlener:
I suppose one of the biggest questions here is whether we are OK with making `fmt` a hard dependency of podio for now or whether we should invest time into making a compatibility layer that switches to `std::format` under the hood if possible / supported by the compiler. All of this should also be possible on top of `std::format`, but there is no mixing & matching between the two libraries, they seem to be mutually exclusive.

The other option is to only have this functionality present in case the compiler has enough `std::format` support, which would put a rather high floor on the compilers. I think for most of the things we would need at least gcc15 (not sure about clang). However, that would probably also mean that we have to maintain dedicated / gated `operator<<` implementations for the builds without `std::format` support.

PR #939, opened 2026-02-24
[WIP] Add a `podio-filter-frames` tool 

tmadlener:
BEGINRELEASENOTES
- Add a `podio-filter-frames` tool that allows to filter whole Frames from files via a user defined `keepFrame` function that is JIT compiled on the fly.

ENDRELEASENOTES

This might simplify things a bit in some cases, e.g. when you want to select files that only contain a certain set of decays and takes off the boiler plate of setting up a reader / writer and an event loop.

- [ ] Documentation
- [ ] Tests

### DIFF
--- a/tools/src/podio-filter-frames.cpp
+++ b/tools/src/podio-filter-frames.cpp
@@ -1,0 +1,117 @@
+#include "argparseUtils.h"
+
+#include "podio/Frame.h"
+#include "podio/Reader.h"
+#include "podio/Writer.h"
+
+#include "TInterpreter.h"
+
+#include <fmt/core.h>
+
+#include <stdexcept>
+#include <string>
+#include <vector>
+
+constexpr auto usageMsg = R"(usage: podio-filter-events [-o outputfile] [-c CATEGORY] filter-def inputfile)";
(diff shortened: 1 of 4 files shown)


tools/src/podio-filter-frames.cpp
```diff
@@ -0,0 +1,117 @@
...
+  if (gInterpreter->LoadFile(filterDefFile.c_str()) != 0) {
+    fmt::println(stderr, "Cannot JIT compile file {}", filterDefFile);
+    std::exit(1);
+  }
+
+  if (const auto filterFunc = gInterpreter->FindSym("_Z9keepFrameRKN5podio5FrameE")) {
+    return reinterpret_cast<FilterFunc*>(filterFunc);
+  }
+
+  fmt::println(stderr, "Could not find a 'keepFrame' function in {}", filterDefFile);
```
andresailer:
```suggestion
  fmt::println(stderr, "Could not find a 'bool keepFrame(const podio::Frame&)' function in {}", filterDefFile);
```

tmadlener:
Probably related to #970 (resp. this might be implementable in terms of it).

PR #952, opened 2026-04-02
Use `TFileMerger` to speed up `podio-merge-files`

jmcarcell:
Currently `podio-merge-files` reads every Frame and writes it back. This is very slow (between 10x and 100x) compared to using `hadd`; I would say unusable. For TTrees and RNTuples, our files can be merged with `hadd`, producing readable files with the only caveat that there are some categories that may be repeated a few times, like metadata. The new implementation is equivalent to what we had before, but much faster.

BEGINRELEASENOTES
- Use `TFileMerger` with ROOT files in `podio-merge-files` to speed it up and add code to make sure the metadata is set correctly. Keep the old behaviour for the other backends (SIO only).

ENDRELEASENOTES

@meleneemil

### DIFF
--- a/python/podio/root_merge.py
+++ b/python/podio/root_merge.py
@@ -1,0 +1,178 @@
+#!/usr/bin/env python3
+"""Utilities for merging podio ROOT files (TTree and RNTuple)."""
+
+import os
+import shutil
+import subprocess
+import tempfile
+
+import ROOT
+
+from podio.frame import Frame
+from podio.utils import convert_to_str_paths
+from podio.root_io import Reader, RNTupleReader, RNTupleWriter, Writer
+
+
+def _get_tree_names(filename):
+    """Return the names of all TTrees in a ROOT file"""
+    f = ROOT.TFile.Open(filename)
+    if not f or f.IsZombie():
+        raise RuntimeError(f"Cannot open file: {filename}")
+    names = {elem.GetName() for elem in f.GetListOfKeys() if elem.GetClassName() == "TTree"}
+    f.Close()
+    return names
+
+
+def _get_rntuple_names(filename):
+    """Return the names of all RNTuples in a ROOT file"""
+    f = ROOT.TFile.Open(filename)
+    if not f or f.IsZombie():
+        raise RuntimeError(f"Cannot open file: {filename}")
+    names = {elem.GetName() for elem in f.GetListOfKeys() if elem.GetClassName() == "ROOT::RNTuple"}
+    f.Close()
+    return names
+
+
+def _hadd_path():
+    """Return the path to the hadd executable, or raise if not found"""
+    path = shutil.which("hadd")
+    if path is None:
+        raise RuntimeError(
+            "hadd not found on PATH. hadd is required by podio.root_merge.merge_files "
+            "and is distributed with ROOT."
+        )
+    return path
+
+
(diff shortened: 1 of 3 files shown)


python/podio/root_merge.py
```diff
@@ -0,0 +1,178 @@
...
+            tmp_writer.write_frame(frame, "metadata")
+        tmp_writer._writer.finish()
+
+        m = ROOT.TFileMerger(ROOT.kFALSE)
+        m.OutputFile(output_file, "UPDATE")
+        m.AddFile(tmp_path)
+        m.SetFastMethod(ROOT.kTRUE)
+        m.AddObjectNames("metadata")
+        if not m.PartialMerge(ROOT.TFileMerger.kAll | ROOT.TFileMerger.kOnlyListed | ROOT.TFileMerger.kIncremental):
+            raise RuntimeError(f"TFileMerger failed adding metadata {fmt_name} to {output_file}")
```
tmadlener:
```suggestion
            raise RuntimeError(f"TFileMerger failed adding metadata category to {output_file}")
```
Would make it possible to drop the `fmt_name` from the arguments with essentially the same information.

python/podio/root_merge.py
```diff
@@ -0,0 +1,178 @@
...
+        src_reader = reader_cls([input_files[0]] if metadata == "first" else input_files)
+        frames = list(src_reader.get("metadata"))
+    else:
+        frames = [Frame()]
+
+    for frame in frames:
+        frame.put_parameter("MergedInputFiles", list(input_files))
+
+    tmp_fd, tmp_path = tempfile.mkstemp(suffix=".root")
+    os.close(tmp_fd)
```
tmadlener:
```python
with tempfile.NamedTemporaryFile(suffix=".root", delete=True) as f:
    tmp_path = f.name
```

that way the cleanup below would be automatically done. Or do we need the file handle closed for ROOT?

tmadlener (commented):
Looks good to me (minor comments below). I suppose we are not yet at a stage where we have less code if we use the root file merger (that `hadd` uses internally) directly instead of going through `hadd` and then "fixing things" afterwards?

python/podio/root_merge.py
jmcarcell: `fmt_name` dropped.

python/podio/root_merge.py
jmcarcell: Ok, but then a weird `tmp_file.close()` is needed to let the reader open it.

jmcarcell:
> Looks good to me (minor comments below). I suppose we are not yet at a stage where we have less code if we use the root file merger (that `hadd` uses internally) directly instead of going through `hadd` and then "fixing things" afterwards?

In the end I tried both with two passes of `hadd` and using `TFileMerger`; for merging files during several minutes it takes the same time. I'll leave the implementation with `TFileMerger` that doesn't have to run subprocesses. And it is quite close in time to running `hadd` directly, plus some constant overhead due to opening files, python and creating the metadata. I ran with thousands of files from EOS, RNTuples and it seems to work fine (and as quickly as `hadd`). `hadd` is also quite fast because it uses a low compression, which I have enabled as a default here but it is configurable. This should be strictly an improvement over what was there before.
Please @meleneemil, have a try again, it will appear in the nightlies tomorrow.

PR #953, opened 2026-04-13
Draft: Lazy Frame reading for ROOT based I/O

tmadlener:
BEGINRELEASENOTES
- ...

ENDRELEASENOTES

This is an alternative approach to lazy reading, outlined in the comments of #949. In short this
- Introduces a new `ROOTLazyReader` that produces `ROOTLazyFrameData` on `readEntry`.
- `ROOTLazyFrameData` retain a reference to the reader that created them and call back to this reader whenever a `getCollectionBuffers` is issued to try and read the collection on demand only. The common shared state also inluces a `mutex` which allows to lock the necessary TTree whenever necessary.

I have lifted some of the adaptions to the `DataSource` from #949 (including the tests) and added a `lazy` flag to `makeReader` instead of using the `readFrameLazy` that was introduced there.

This currently also does some quite extensive refactoring to lift common functionality into a `ReaderCommon` and into a `TTreeReaderCommon` class as well as generalizing some of the utility functionality. If we go for this approach I would potentially lift some of that into a separate PR.

This does **not yet** implement the same functionality for RNTuple, but that can be added in a similar fashion.

The main advantage over #949 is that this approach allows to transparently switch readers (at least when going through the `Reader`), but in principle also when using the low level readers as `podio::Frame(reader.readNextEntry())` will keep working even if the `reader` changes type and `readNextEntry` returns a different frame data type.


tmadlener:
This now also contains a first version for RNTuple with a similar approach as for TTree, i.e. moving common functionality into a base class and only specializing where necessary. There is still some potential for improvement probably, but tests pass.

PR #955, opened 2026-04-15
Add I/O tests that run on multiple threads

tmadlener:
BEGINRELEASENOTES
- Add tests that run I/O on multiple threads

ENDRELEASENOTES

These tests build / read Frames on multiple threads serializing the I/O part (as per design). They effectively check whether the Frame construction as well as the retrival of buffers from the `FrameData` is done properly when running under TSan. They do not check whether accessing a single Frame from multiple threads is thread safe that is covered by unit tests.

This was something that I would have liked to do for quite some time, and due to https://github.com/AIDASoft/podio/pull/940 it has become possible to actually run these under the sanitizers so that there is some actual checks behind it as well.

### DIFF
--- a/tests/CTestCustom.cmake
+++ b/tests/CTestCustom.cmake
@@ -80,6 +80,11 @@ if ((NOT "@FORCE_RUN_ALL_TESTS@" STREQUAL "ON") AND (NOT "@USE_SANITIZER@" STREQ
       read_python_with_rdatasource_root
       param_reading_rdataframe
 
+      write_frame_root_multithreaded
+      read_frame_root_multithreaded
+      write_frame_rntuple_multithreaded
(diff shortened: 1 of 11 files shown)


tmadlener:
I also confirmed that running on 8 threads with 1000 events each doesn't trigger TSan (just in case). Works as well. For that case there is quite a bit of contention on the read / write lock, but getting optimal performance out of this is not the goal of the tests.

jmcarcell (approved):
After the last fixes, CI should pass

jmcarcell:
I checked if adding `PROCESSORS` to CMake to tell CMake the tests are using multiple threads does anything but they are so fast that it seems not so I'll leave as it is.
https://cmake.org/cmake/help/latest/prop_test/PROCESSORS.html

PR #963, opened 2026-04-17
Hash action versions

tmadlener:
BEGINRELEASENOTES
- Pin all used github actions to the commit that corresponds to the latest released tag

ENDRELEASENOTES

Doing this in favor of just bumping versions as originally proposed by dependabot
- #960 
- #961 
- #962
- #959 
- #958

### DIFF
--- a/.github/workflows/sanitizers.yaml
+++ b/.github/workflows/sanitizers.yaml
@@ -32,10 +32,10 @@ jobs:
         #   - compiler: clang10
         #     sanitizer: MemoryWithOrigin
     steps:
-      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
-      - uses: cvmfs-contrib/github-action-cvmfs@10197e000cc0add8e54ac4fb73d3ed44e2de72b4
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - uses: cvmfs-contrib/github-action-cvmfs@10197e000cc0add8e54ac4fb73d3ed44e2de72b4 # v5.5
       - uses: key4hep/key4hep-actions/cache-external-data@main
-      - uses: aidasoft/run-lcg-view@d8db17c61216548c9192889942d46ff2d7c3a3b6
+      - uses: aidasoft/run-lcg-view@d8db17c61216548c9192889942d46ff2d7c3a3b6 # v6
         with:
           release-platform: LCG_108/x86_64-el9-${{ matrix.compiler }}-opt
           ccache-key: ccache-sanitizers-el9-${{ matrix.compiler }}-${{ matrix.sanitizer }}
--- a/.github/workflows/sanitizers.yaml
+++ b/.github/workflows/sanitizers.yaml
@@ -32,7 +32,7 @@ jobs:
         #   - compiler: clang10
         #     sanitizer: MemoryWithOrigin
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
       - uses: cvmfs-contrib/github-action-cvmfs@v5
       - uses: key4hep/key4hep-actions/cache-external-data@main
       - uses: aidasoft/run-lcg-view@v5
--- a/.github/workflows/sanitizers.yaml
+++ b/.github/workflows/sanitizers.yaml
@@ -35,7 +35,7 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
       - uses: cvmfs-contrib/github-action-cvmfs@v5
       - uses: key4hep/key4hep-actions/cache-external-data@main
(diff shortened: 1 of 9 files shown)


jmcarcell:
Dependabot alerts don't work with hashes: https://docs.github.com/en/code-security/concepts/supply-chain-security/about-dependabot-alerts#limitations
Plus anyway if the repository has been compromised how do you know that commit is not? I'm not convinced hashes are much better than versions.

m-fila:
What is the reason to do so? If you want to pin explicit version you can do "v6.3.1"  - the convention is not to update these

jmcarcell:
Security, tags are not immutable. It seems the best practice is to pin to a hash: https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions but then you don't get dependabot security alerts? :thinking:

m-fila:
Ah I see. Then unfortunately no dependabot here :stuck_out_tongue_closed_eyes:
If I'm not mistaken this repo has only official actions, actions by credible sources (cvmfs) or "3rd parties" which are controlled by people with write access to this repo anyway

andresailer:
> Plus anyway if the repository has been compromised how do you know that commit is not? 

At least the content of the commit doesn't change (until someone finds a way for hash collisions...), while it can change for tags. So this is maybe not really needed for where we put them, but we might as well just go for hashes everywhere as a principle.

> If I'm not mistaken this repo has only official actions, actions by credible sources (cvmfs) or "3rd parties" which are controlled by people with write access to this repo anyway

Security in depth!

.github/workflows/sanitizers.yaml
```diff
@@ -32,10 +32,10 @@ jobs:
         #   - compiler: clang10
         #     sanitizer: MemoryWithOrigin
     steps:
-      - uses: actions/checkout@v4
-      - uses: cvmfs-contrib/github-action-cvmfs@v5
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
+      - uses: cvmfs-contrib/github-action-cvmfs@10197e000cc0add8e54ac4fb73d3ed44e2de72b4
       - uses: key4hep/key4hep-actions/cache-external-data@main
```
andresailer, line 37: What about this action?

m-fila:
> Dependabot alerts don't work with hashes: https://docs.github.com/en/code-security/concepts/supply-chain-security/about-dependabot-alerts#limitations

Wait, the updates seem to work with hashes no?  https://docs.github.com/en/code-security/reference/supply-chain-security/supported-ecosystems-and-repositories?utm_source=chatgpt.com#github-actions 

For example, here in TBB they pin by hash and have comment, and they get an update https://github.com/uxlfoundation/oneTBB/pull/2037

The docs you mentioned were about dependabot alerts which are different thing from dependaboty version updates. (which means that if a vulnerability is detected in the version but we pin it with hash not tag then there won't be an alert)

tmadlener:
> Wait, the updates seem to work with hashes no? https://docs.github.com/en/code-security/reference/supply-chain-security/supported-ecosystems-and-repositories?utm_source=chatgpt.com#github-actions

IIUC, we could also do this to track which tag we are tracking and dependabot would automatically update the comment as well(?)
```yaml
- uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd #6.0.2
```

.github/workflows/sanitizers.yaml
tmadlener: Good question (also for other repositories). We currently don't even tag this I think.

m-fila:
That is my understanding as well

jmcarcell:
There is also this in Settings -> Advanced security (disabled by default):

> Dependabot security updates

> Enabling this option will result in Dependabot automatically attempting to open pull requests to resolve every open Dependabot alert with an available patch. If you would like more specific configuration options, leave this disabled and use [Dependabot rules](https://github.com/AIDASoft/podio/settings/dependabot_rules). 

This should be probably updated? But if there are no alerts when using commits does this work? Also alerts is disabled by default

m-fila:
> IIUC, we could also do this to track which tag we are tracking and dependabot would automatically update the comment as well(?)

I checked on some other repo and indeed it seems to work out of the box
- https://github.com/m-fila/coopsync_tbb/pull/77/changes
- https://github.com/m-fila/coopsync_tbb/pull/78/changes

tmadlener:
Dependabot alerts are enabled for podio. We get some fixes for the `doc/requirements.txt` from time to time from that.

jmcarcell:
The security ones are disabled:

2026-04-17-143936_1288x116_scrot

but these will not work anyway with hashes (docs say so).

tmadlener:
Ah, now I get what you mean. But if we get auto-updates in any case, does it matter too much whether the update has been triggered by a security issue or by something else?

edit:
> The security ones are disabled:

The screenshot you put says that they are enabled and that we could disable them with the button.

tmadlener:
I would merge these as is, unless there are more comments. We can decide what to do with our own `key4hep-action` in a separate PR, I think.

m-fila (approved):

PR #966, opened 2026-05-22
feat: Add Apache Arrow schema mapper generation to code generator

arnavdham:
**Description**
This PR introduces the initial foundation for the Apache Arrow interface in PODIO by implementing compile-time schema mapper generation within the python code generator engine.

The generated mapping translates YAML-defined Event Data Models (EDMs) into Arrow schema structures (supporting primitives, components, relations, and frame parameters). For full details on the mapping decisions, please refer to [this](https://docs.google.com/document/d/1vvC0pVLYdkd3_h0zg_pJxQ7msC1myxSdamwZ0Fv_cIM/edit?usp=sharing).

**Unit Test**
Validated correct generator schema output by running the new unit test:
`PYTHONPATH=python:$PYTHONPATH python3 python/podio_gen/test_CPPGenerator.py`

**Manual Verification**
Run the generator script directly:
`python3 python/podio_class_generator.py tests/datalayout.yaml TmpArrow TestDataModel ARROW`
This will output the generated file under TmpArrow/TestDataModel/ArrowMapper.h.

BEGINRELEASENOTES

- Lay the initial foundation for the Apache Arrow backend by implementing compile-time schema mapper generation within the python code generator engine.
- Add support for translating YAML-defined Event Data Models (EDMs) into Apache Arrow schema structures, including support for primitives, components, relations, and frame parameters.

ENDRELEASENOTES


arnavdham:
Related issue: https://github.com/AIDASoft/podio/issues/968

veprbl:
@jmcarcell Hello. This is our first patch towards implementing the Arrow backend. The built "schema" is not practical one, but will give us some foundation to build upon. I've tested that generated code compiles against arrow-cpp. Could you take a look?

jmcarcell:
I can have a look but this is only a starting point. At least these things are missing:
- An option to enable arrow from CMake
- A test that actually uses the generated code (and is disabled otherwise)
- A CI workflow that runs it (I think all the stacks used have arrow)

python/podio_gen/test_CPPGenerator.py
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env python3
```
jmcarcell: I would rather have a test that uses what is produced using arrow that a test that parses the generated file and checks if what's expected is there; actually using it will be more robust.

arnavdham: Sure.

arnavdham: I have added a unittest. Do you want me to remove the existing one?

cmake/podioMacros.cmake
```diff
@@ -197,7 +197,9 @@ function(PODIO_ADD_DATAMODEL_CORE_LIB lib_name HEADERS SOURCES)
 
   # Filter out anything I/O backend related to build the core library
   LIST(FILTER HEADERS EXCLUDE REGEX .*SIOBlock.h)
-  LIST(FILTER HEADERS EXCLUDE REGEX .*ArrowMapper.h)
+  IF(NOT ENABLE_ARROW)
```
veprbl: Looks suspicious. The file can't exist if NOT ENABLE_ARROW. Plus if this is for external use, the option is not even defined.

Should we make this unconditional, since it's already such for SIO?

arnavdham: My bad. It was unconditional before, I will change it back.

python/podio_gen/test_CPPGenerator.py
jmcarcell: Yes, the new test should cover what the old one does (and more) so the existing one can be removed.

arnavdham: The new test was added already. Will remove this existing one today.

arnavdham: @jmcarcell done.

tests/unittests/unittest.cpp
```diff
@@ -1778,3 +1778,103 @@ TEST_CASE("Add type lists", "[basics][code-gen]") {
...
+  auto schema = datamodel::arrow_io::schema();
+  REQUIRE(schema != nullptr);
+  auto field_names = schema->field_names();
+
+  // 1. Verify top-level fields
+  REQUIRE(std::find(field_names.begin(), field_names.end(), "EventInfo") != field_names.end());
+  REQUIRE(std::find(field_names.begin(), field_names.end(), "ExampleHit") != field_names.end());
+  REQUIRE(std::find(field_names.begin(), field_names.end(), "ExampleMC") != field_names.end());
+  REQUIRE(std::find(field_names.begin(), field_names.end(), "ExampleWithArray") != field_names.end());
+  REQUIRE(std::find(field_names.begin(), field_names.end(), "frame_parameters") != field_names.end());
```
jmcarcell:
```suggestion

  // 1. Verify top-level fields
  auto field_names = schema->field_names();
  REQUIRE(std::find(field_names.begin(), field_names.end(), "EventInfo") != field_names.end());
  REQUIRE(std::find(field_names.begin(), field_names.end(), "ExampleHit") != field_names.end());
  REQUIRE(std::find(field_names.begin(), field_names.end(), "ExampleMC") != field_names.end());
  REQUIRE(std::find(field_names.begin(), field_names.end(), "ExampleWithArray") != field_names.end());
  REQUIRE(std::find(field_names.begin(), field_names.end(), "frame_parameters") != field_names.end());
```
Can it be done like in the rest of the file where whatever is obtained before the `REQUIRE`s goes together with the `REQUIRE`s? For consistency and because it makes it easy to have a quick look. In general below it is fine.

tests/unittests/unittest.cpp
```diff
@@ -1778,3 +1778,103 @@ TEST_CASE("Add type lists", "[basics][code-gen]") {
...
+  REQUIRE(parentIndex != nullptr);
+  REQUIRE(parentIndex->type()->id() == arrow::Type::INT32);
+
+  // 6. Verify frame_parameters (Map mapping)
+  auto frameParamsField = schema->GetFieldByName("frame_parameters");
+  REQUIRE(frameParamsField != nullptr);
+  REQUIRE(frameParamsField->type()->id() == arrow::Type::STRUCT);
+  auto frameParamsType = std::static_pointer_cast<arrow::StructType>(frameParamsField->type());
+
+  auto intParamsField = frameParamsType->GetFieldByName("int_params");
```
jmcarcell:
```suggestion

  auto frameParamsType = std::static_pointer_cast<arrow::StructType>(frameParamsField->type());
  auto intParamsField = frameParamsType->GetFieldByName("int_params");
```

python/templates/ArrowMapper.h.jinja2
```diff
@@ -0,0 +1,49 @@
...
+#define {{ package_name.upper() }}_ARROWMAPPER_H
+
+#include <arrow/type.h>
+#include <arrow/util/key_value_metadata.h>
+
+#include <memory>
+
+namespace {{ package_name }}::arrow_io {
+
+inline std::shared_ptr<arrow::DataType> objectRefType() {
```
jmcarcell, line 13: Is `inline` the "default" way of doing this?

jmcarcell:
> @jmcarcell Hello. This is our first patch towards implementing the Arrow backend. The built "schema" is not practical one, but will give us some foundation to build upon. I've tested that generated code compiles against arrow-cpp. Could you take a look?

What do you mean it is not practical?
How would we go about this, once I don't have any more comments I merge?

veprbl:
> What do you mean it is not practical? How would we go about this, once I don't have any more comments I merge?

This builds a plausible schema where each known type is present as a single collection instance under the same name. That demonstrates how we can successfully represent a full EDM, like EDM4hep, in an Arrow schema. Our intent was, once this is merged, the PR to add ArrowWriter will follow with a minor correction to use information from the actual frames being written (this information is not available until we implement a writer). We are open to alternative proposals, of course.

python/templates/ArrowMapper.h.jinja2
arnavdham: Currently, we are using only a ArrowMapper.h file, so inline is needed to avoid multiple re-definition errors. Once we move to a .cc and .h split this would not be needed.(Planned in upcoming PRs).

arnavdham:
@jmcarcell if there are no more comments from your side. Maybe we can merge?

python/templates/ArrowMapper.h.jinja2
```diff
@@ -0,0 +1,49 @@
...
+      arrow::field("string_params", arrow::map(arrow::utf8(), arrow::list(arrow::utf8()))),
+  });
+}
+
+inline std::shared_ptr<arrow::Schema> schema() {
+  return arrow::schema({
+{% for datatype in datatypes %}
+      arrow::field("{{ datatype.class.bare_type }}", arrow::list(arrow::struct_({
+{% for field in datatype.arrow_fields %}
+          {{ field }},
```
jmcarcell, line 34: Formatting of these fields is not great since they all will be on the same line. Maybe one line per field is better so the file is easy to be read for humans? If this is going to be modified later then maybe we don't need to worry about this now and I can merge.

jmcarcell:
> > What do you mean it is not practical? How would we go about this, once I don't have any more comments I merge?
> 
> This builds a plausible schema where each known type is present as a single collection instance under the same name. That demonstrates how we can successfully represent a full EDM, like EDM4hep, in an Arrow schema. Our intent was, once this is merged, the PR to add ArrowWriter will follow with a minor correction to use information from the actual frames being written (this information is not available until we implement a writer). We are open to alternative proposals, of course.



> @jmcarcell if there are no more comments from your side. Maybe we can merge?

I made a last comment about formatting. I think it is OK as it is, given that you plan on making changes in the future, once writing and reading are there we can discuss this in more detail.

python/templates/ArrowMapper.h.jinja2
arnavdham: Yeah this will be fixed later on. Thanks for pointing that out!

jmcarcell (approved):

PR #970, opened 2026-06-10
PoC for consistent collection filtering

paulgessinger:
This is PoC by Claude for filtering collections in a frame based on user-supplied predicates including mark-and-sweep unreachable object (orphan) removal and optional cascading deletes through relationships.

I'm not necessarily proposing to merge this "as is", we might want to change the approach entirely, but I thought it's still useful as the basis for discussion.

The user facing API looks something like this:

```c++
podio::Frame out = podio::FrameFilter{inFrame}
    .keep("MCParticles", [](const ExampleMC& p) { return p.energy() > 1.0; })
    .keepReferenced("hits")
    .cascade("ExampleCluster.Hits")
    .run();
```

- [ ] Think about integration with Gaudi, can only happen at the end
- [ ] Potentially interaction with DataSource, but unclear if relevent
- [ ] Do some basic performance measurements
- [ ] Drop cascade for now

BEGINRELEASENOTES
- ...

ENDRELEASENOTES


paulgessinger:
We had some discussion here: https://indico.cern.ch/event/1711603/contributions/7200389/attachments/3317030/5937793/2026-07-21_podio-filtering.pdf

doc/filtering.md
```diff
@@ -0,0 +1,152 @@
...
+
+- `keep(name, predicate)` — keep only the objects of the named collection for
+  which `predicate` returns `true`. The collection type is deduced from the
+  predicate's argument type.
+- `keepReferenced(name)` — keep an object of the named collection only if some
+  surviving object still references it (an *orphan sweep*).
+- `drop(name)` — remove the named collection entirely; it is omitted from the
+  output Frame. Unlike `keep`, this needs only the name, not the collection's
+  type. (To keep an empty collection instead, use an always-false `keep`
+  predicate.)
```
tmadlener, line 34: This confused me for a bit. There are two ways we can end up with empty collections after the filter has run. Either we have filtered out all elements of a collection, or we started the filtering with an empty collection. Not sure if this needs to be clarified in the end, I just wanted to not I stumbled over this while reading.

tmadlener (commented):
I haven't looked at the details (or the code) yet too much. This is definitely useful though. Conceptually, I think this is a pretty nice approach. Is there any indication already if / how this would carry over to python? One reason to ask is because this might become useful in combination with the proposal in #939 to potentially do some collection level filtering first and then some whole event skimming afterwards. Or potentially that functionality can be entirely absorbed here(?).

I wasn't present when this was discussed, so maybe someone who was can clarify / extend a bit more on the open points? Most importantly about the Gaudi and DataSource interactions.

I suppose the generic functionality of the `FrameFilter` would remain largely in the hands of experts and only some of the functionality would in the end also be exposed via e.g. Gaudi?

Finally, I think I understand what `cascade` is supposed to do, but I am not sure I would have intuitively guessed it from the name.

PR #973, opened 2026-06-17
Remove the deprecated method isValid for Collections

jmcarcell:
BEGINRELEASENOTES
- **Remove** the method **`isValid`** for Collections (generated and library provided). **Use `hasID`** if you want to check if a collection has been assigned a collection id, otherwise assume a collection is valid.

ENDRELEASENOTES

Deprecated since 1.06, removed once we get another tag.

### DIFF
--- a/include/podio/CollectionBase.h
+++ b/include/podio/CollectionBase.h
@@ -48,8 +48,6 @@ public:
 
   virtual bool hasID() const = 0;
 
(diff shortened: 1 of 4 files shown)


jmcarcell:
Ready to be merged

tmadlener (approved):

PR #987, opened 2026-07-15
Implement Arrow-to-Frame Conversion (Deserialization) for Arrow Backend

### LINKED ISSUE #986: Implement Arrow Table-to-Frame Conversion (Deserialization) in PODIO
We need to add support for deserializing Apache Arrow data back into PODIO Frame objects to complete the round-trip workflow for the Apache Arrow backend.

**Key Requirements**

- **convertTableToFrame API**: Reconstruct a Frame from a single row slice of an Arrow Table using an ArrowFrameData class that implements the C++20 FrameDataType concept.
- **Collection & Relation Reconstruction:** Read collections, subset collections, vector members, relations, and Link Collections from Arrow data using schema metadata (coll_id, value_type, is_subset, etc.).
- **Code Generation**: Update cpp_generator.py and templates (ArrowMapper.cc.jinja2) to generate reader functions for all data model types.
- **Generic Parameters**: Reconstruct GenericParameters from the frame_parameters column in the Arrow Table.
- **Testing**: Add unit tests verifying end-to-end round-trips (Frame $\rightarrow$ Table $\rightarrow$ Frame) and multi-row reading.

arnavdham:
This PR implements the deserialization capability for the Apache Arrow backend, completing the round-trip workflow. It adds support for reconstructing PODIO Frame objects from a single-row slice of an Apache Arrow Table.

**Key Features & Changes**

- **convertTableToFrame API**: Introduced a public utility to read and reconstruct a Frame from a specific row of an Arrow Table.
- **Auto-generated Readers**: Updated cpp_generator.py and generator templates (ArrowMapper.cc.jinja2) to generate deserialization/reader functions for all concrete data model types.
- **Complex Type Support:** Handles reconstruction of collections, subset collections, vector members, relations/references, and Link Collections.
- **Generic Parameters**: Reconstructs GenericParameters from the frame_parameters struct column in the Arrow Table.

**Testing**

- Added comprehensive unit tests in test_arrow_converter.cpp covering:
- Full round-trip (Frame $\rightarrow$ Table $\rightarrow$ Frame) checks.
- Multi-row tables and non-zero row index offset mapping.
- Verification of reconstructed relations, subsets, and parameter maps.

Fixes: #986 

BEGINRELEASENOTES

* Add support for Arrow-to-Frame conversion (deserialization) for the Apache Arrow backend to complete the round-trip workflow.
* Introduce the `convertTableToFrame` API to reconstruct `podio::Frame` objects from a specific row of an Arrow `Table`.
* Add support for reconstructing complex types (collections, vector members, relations, subsets) and `GenericParameters`.

ENDRELEASENOTES

### DIFF
--- a/include/podio/utilities/ArrowConverterRegistry.h
+++ b/include/podio/utilities/ArrowConverterRegistry.h
@@ -1,6 +1,8 @@
 #ifndef PODIO_ARROWCONVERTERREGISTRY_H
 #define PODIO_ARROWCONVERTERREGISTRY_H
 
+#include "podio/CollectionBuffers.h"
+#include <cstdint>
 #include <functional>
 #include <memory>
 #include <string>
@@ -20,6 +22,7 @@ class CollectionBase;
 class ArrowConverterRegistry {
 public:
   using CreatorFunc = std::function<std::shared_ptr<arrow::Array>(const podio::CollectionBase*)>;
+  using BufferReaderFunc = std::function<podio::CollectionReadBuffers(std::shared_ptr<arrow::Array>, int64_t, bool)>;
 
   ArrowConverterRegistry(const ArrowConverterRegistry&) = delete;
   ArrowConverterRegistry& operator=(const ArrowConverterRegistry&) = delete;
@@ -44,11 +47,23 @@ public:
    */
   CreatorFunc getConverter(const std::string& typeName) const;
 
+  /**
+   * @brief Register an Arrow array reader callback for a specific type name.
+   */
+  void registerReader(const std::string& typeName, BufferReaderFunc reader);
+
+  /**
+   * @brief Retrieve the Arrow collection reader registered for a specific type name.
+   * @return The registered BufferReaderFunc, or nullptr if not registered.
+   */
+  BufferReaderFunc getReader(const std::string& typeName) const;
+
 private:
   ArrowConverterRegistry() : m_registry() {
   }
 
   std::unordered_map<std::string, CreatorFunc> m_registry;
+  std::unordered_map<std::string, BufferReaderFunc> m_readerRegistry;
 };
 
 } // namespace podio
--- a/include/podio/utilities/ArrowConverterRegistry.h
+++ b/include/podio/utilities/ArrowConverterRegistry.h
@@ -59,7 +59,7 @@ public:
   BufferReaderFunc getReader(const std::string& typeName) const;
 
 private:
-  ArrowConverterRegistry() : m_registry() {
+  ArrowConverterRegistry() : m_registry(), m_readerRegistry() {
   }
 
   std::unordered_map<std::string, CreatorFunc> m_registry;
--- a/include/podio/utilities/ArrowConverterRegistry.h
+++ b/include/podio/utilities/ArrowConverterRegistry.h
@@ -2,9 +2,11 @@
 #define PODIO_ARROWCONVERTERREGISTRY_H
 
 #include "podio/CollectionBuffers.h"
+#include "podio/SchemaEvolution.h"
 #include <cstdint>
 #include <functional>
 #include <memory>
+#include <optional>
 #include <string>
 #include <unordered_map>
 
@@ -22,7 +24,8 @@ class CollectionBase;
 class ArrowConverterRegistry {
 public:
   using CreatorFunc = std::function<std::shared_ptr<arrow::Array>(const podio::CollectionBase*)>;
-  using BufferReaderFunc = std::function<podio::CollectionReadBuffers(std::shared_ptr<arrow::Array>, int64_t, bool)>;
+  using BufferReaderFunc = std::function<std::optional<podio::CollectionReadBuffers>(std::shared_ptr<arrow::Array>,
+                                                                                     int64_t, bool, SchemaVersionT)>;
 
   ArrowConverterRegistry(const ArrowConverterRegistry&) = delete;
   ArrowConverterRegistry& operator=(const ArrowConverterRegistry&) = delete;
--- a/include/podio/utilities/ArrowConverterRegistry.h
+++ b/include/podio/utilities/ArrowConverterRegistry.h
@@ -5,10 +5,10 @@
 #include "podio/SchemaEvolution.h"
 #include <cstdint>
 #include <functional>
+#include <map>
 #include <memory>
 #include <optional>
 #include <string>
-#include <unordered_map>
 
 namespace arrow {
 class Array;
@@ -53,7 +53,7 @@ public:
   /**
    * @brief Register an Arrow array reader callback for a specific type name.
    */
-  void registerReader(const std::string& typeName, BufferReaderFunc reader);
+  void registerReader(const std::string& typeName, BufferReaderFunc&& reader);
 
   /**
    * @brief Retrieve the Arrow collection reader registered for a specific type name.
@@ -65,8 +65,8 @@ private:
   ArrowConverterRegistry() : m_registry(), m_readerRegistry() {
   }
 
-  std::unordered_map<std::string, CreatorFunc> m_registry;
-  std::unordered_map<std::string, BufferReaderFunc> m_readerRegistry;
+  std::map<std::string, CreatorFunc> m_registry;
+  std::map<std::string, BufferReaderFunc> m_readerRegistry;
 };
 
 } // namespace podio
--- a/include/podio/utilities/ArrowFrameConverter.h
+++ b/include/podio/utilities/ArrowFrameConverter.h
@@ -42,9 +42,9 @@ std::shared_ptr<arrow::Table> convertFrameToTable(const podio::Frame& frame,
  *
  * @param table The Arrow Table containing the frame data.
  * @param rowIndex The index of the row to read.
- * @return A unique pointer to the reconstructed Frame.
+ * @return The reconstructed Frame.
  */
-std::unique_ptr<podio::Frame> convertTableToFrame(const std::shared_ptr<arrow::Table>& table, int rowIndex = 0);
+podio::Frame convertTableToFrame(const std::shared_ptr<arrow::Table>& table, int rowIndex = 0);
 
 } // namespace podio
 
--- a/include/podio/utilities/ArrowFrameConverter.h
+++ b/include/podio/utilities/ArrowFrameConverter.h
@@ -35,6 +35,18 @@ std::shared_ptr<arrow::DataType> objectRefType();
 std::shared_ptr<arrow::Table> convertFrameToTable(const podio::Frame& frame,
                                                   const std::vector<std::string>& collsToWrite);
 
+/**
+ * @brief Convert a 1-row slice of an Arrow Table to a PODIO Frame.
+ *
+ * Reconstructs the frame parameters and all collection buffers using registered BufferReaderFuncs.
+ * 
+ * @param table The Arrow Table containing the frame data.
+ * @param rowIndex The index of the row to read.
+ * @return A unique pointer to the reconstructed Frame.
+ */
+std::unique_ptr<podio::Frame> convertTableToFrame(const std::shared_ptr<arrow::Table>& table, int rowIndex = 0);
+
+
 } // namespace podio
 
 #endif // PODIO_ARROWFRAMECONVERTER_H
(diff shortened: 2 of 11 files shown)


arnavdham:
@jmcarcell Please take a look.

tests/unittests/test_arrow_converter.cpp
```diff
@@ -479,3 +485,455 @@ TEST_CASE("ArrowFrameConverter - JSON Equivalence Check", "[arrow][converter][js
   // Compare the JSON representations
   REQUIRE(arrowJson == frameJson);
 }
+
+TEST_CASE("ArrowFrameConverter - convertTableToFrame Verification (Round-Trip)", "[arrow][converter][reader]") {
+  podio::Frame originalFrame;
+
+  // 1. Prepare some hits
```
jmcarcell: We should only have have to set up the tests once (ideally together with the ROOT ones since the setup is independent of the backend, and the same when checking back the values). Isn't it implied that the previous tests have to work for the table conversion to work, and would not be needed? I'm thinking about the json tests a few lines above.

tests/unittests/test_arrow_converter.cpp
```diff
@@ -479,3 +485,455 @@ TEST_CASE("ArrowFrameConverter - JSON Equivalence Check", "[arrow][converter][js
...
+      REQUIRE(mcpRefs[i] == moreMCs[index]);
+    }
+  }
+
+  auto& refs = event.get<ExampleReferencingTypeCollection>("refs");
+  auto ref = refs[0];
+  for (auto cluster : ref.Clusters()) {
+    for (auto hit [[maybe_unused]] : cluster.Hits()) {
+    }
+  }
```
jmcarcell: What is the point of this?

tests/unittests/test_arrow_converter.cpp
arnavdham: I am not sure what the comment means exactly.
1. Should I remove both the ad-hoc round-trip test(which has manually made frames) and the `JSON Equivalence Check`, relying entirely on `Comprehensive Round-Trip (No-UserData)` (which uses `makeFrame(0)` and `verifyEventNoUserData`)?
2. Or keep `convertFrameToTable Verification` / `JSON Equivalence Check`, refactored to use `makeFrame(0)` from `write_frame.h` instead of ad-hoc setup?
I am thinking the JSON equivalence is not really required now that we have a round trip test.

tests/unittests/test_arrow_converter.cpp
arnavdham: Was dead code. I've removed the empty loops and replaced the unused variables with explicit `REQUIRE` assertions..

tests/unittests/test_arrow_converter.cpp
jmcarcell: The round trip test does what the json test does right? Then the json test I think is not needed, I think we talked about this when the json test was introduced. Using `makeFrame(0)` from `write_frame.h` would be great; also using the functionality from `read_test.h` would be great and then we can have a single place where a Frame is created and read.

arnavdham: Removed JSON equivalence test and used `makeFrame(0)`

include/podio/utilities/ArrowConverterRegistry.h
```diff
@@ -44,11 +50,23 @@ class ArrowConverterRegistry {
...
+   */
+  BufferReaderFunc getReader(const std::string& typeName) const;
+
 private:
-  ArrowConverterRegistry() : m_registry() {
+  ArrowConverterRegistry() : m_registry(), m_readerRegistry() {
   }
 
   std::unordered_map<std::string, CreatorFunc> m_registry;
+  std::unordered_map<std::string, BufferReaderFunc> m_readerRegistry;
```
jmcarcell: Maybe these could just be `std::map` since there won't be that many elements there anyway? We may have discussed this before

include/podio/utilities/ArrowFrameConverter.h
```diff
@@ -35,6 +35,17 @@ std::shared_ptr<arrow::DataType> objectRefType();
...
+/**
+ * @brief Convert a 1-row slice of an Arrow Table to a PODIO Frame.
+ *
+ * Reconstructs the frame parameters and all collection buffers using registered BufferReaderFuncs.
+ *
+ * @param table The Arrow Table containing the frame data.
+ * @param rowIndex The index of the row to read.
+ * @return A unique pointer to the reconstructed Frame.
+ */
+std::unique_ptr<podio::Frame> convertTableToFrame(const std::shared_ptr<arrow::Table>& table, int rowIndex = 0);
```
jmcarcell: Would it make sense to return directly a `podio::Frame`? Typically we work with Frames directly and using pointers is not needed that much. It makes it a bit weird to always have to use the `->` and from the code below I haven't seen that this needs to be a `std::unique_ptr`.

src/ArrowConverterRegistry.cc
```diff
@@ -23,4 +23,16 @@ ArrowConverterRegistry::CreatorFunc ArrowConverterRegistry::getConverter(const s
   return nullptr;
 }
 
+void ArrowConverterRegistry::registerReader(const std::string& typeName, BufferReaderFunc reader) {
```
jmcarcell:
```suggestion
void ArrowConverterRegistry::registerReader(const std::string& typeName, BufferReaderFunc&& reader) {
```
shouldn't it be `&&`, otherwise you are moving a copy?

src/ArrowFrameData.cc
```diff
@@ -0,0 +1,204 @@
...
+
+  auto reader = podio::ArrowConverterRegistry::instance().getReader(typeName);
+  if (!reader) {
+    throw std::runtime_error("No Arrow reader registered for type: " + typeName);
+  }
+
+  return reader(chunk, remaining_idx, isSubset, schemaVersion);
+}
+
+std::unique_ptr<podio::GenericParameters> ArrowFrameData::getParameters() {
```
jmcarcell, line 138: Same as with the frame, do you need to return a `std::unique_ptr`?

include/podio/utilities/ArrowConverterRegistry.h
arnavdham: Changed to `std::map` for both m_registry and m_readerRegistry

include/podio/utilities/ArrowFrameConverter.h
arnavdham: Changed to returning a Frame directly

src/ArrowConverterRegistry.cc
arnavdham: Fixed

src/ArrowFrameData.cc
arnavdham: The FrameDataType concept in `Frame.h` forces `getParameters()` to return a `std::unique_ptr`. We'd have to change it everywhere (like in ROOTFrameData and SIOFrameData) if we touch it here.
https://github.com/AIDASoft/podio/blob/master/include/podio/Frame.h#L46

jmcarcell: Ok then nevermind

tests/unittests/test_arrow_converter.cpp
```diff
@@ -168,314 +131,349 @@ TEST_CASE("ArrowFrameConverter - convertFrameToTable Verification", "[arrow][con
...
-  originalFrame.putParameter("anInt", 42);
-  originalFrame.putParameter("someFloats", std::vector<float>{1.23f, 2.34f, 3.45f});
-  originalFrame.putParameter("someStrings", std::vector<std::string>{"hello", "world"});
-
-  std::vector<std::string> colls = {"Hits",       "Clusters",         "OneRelation",  "VectorMember",
-                                    "SubsetHits", "ExampleWithArray", "EmptyCluster", "FixedWidthInts"};
+TEST_CASE("ArrowFrameConverter - Comprehensive Round-Trip (No-UserData)", "[arrow][converter][common]") {
+  auto originalFrame = makeFrame(0);
+
+  const std::vector<std::string> colls = {"mcparticles",
```
jmcarcell: Is it possible to get this from the frame? `getAvailableCollections`?

tests/unittests/test_arrow_converter.cpp
jmcarcell: Looks good

src/ArrowFrameData.cc
```diff
@@ -0,0 +1,204 @@
...
+
+  auto typeNameIdx = metadata->FindKey("value_type");
+  auto isSubsetIdx = metadata->FindKey("is_subset");
+
+  if (typeNameIdx == -1 || isSubsetIdx == -1) {
+    throw std::runtime_error("Missing value_type or is_subset in metadata for: " + name);
+  }
+
+  std::string typeName = metadata->value(typeNameIdx);
+  bool isSubset = (metadata->value(isSubsetIdx) == "1");
```
jmcarcell, line 119: Is this "1" or 1?

src/ArrowFrameData.cc
```diff
@@ -0,0 +1,204 @@
...
+#include <algorithm>
+#include <arrow/api.h>
+#include <stdexcept>
+#include <type_traits>
+
+namespace podio {
+
+namespace {
+
+  struct IntSetter {
```
jmcarcell: What about having a single template for `{Int,Float,Double,String}` setter? and then they're initialized in extractMap that seems to be the only place where they are used?

python/templates/ArrowMapper.cc.jinja2
```diff
@@ -95,12 +161,206 @@ bool registerConverters() {
...
+
+    auto appendRelationRef = [](arrow::StructBuilder* structBuilder, const podio::ObjectID& objId) {
+      checkStatus(structBuilder->Append(), "Failed to append relation struct");
+      checkStatus(static_cast<arrow::UInt32Builder*>(structBuilder->child(0))->Append(objId.collectionID), "Failed to append collectionID");
+      checkStatus(static_cast<arrow::Int32Builder*>(structBuilder->child(1))->Append(objId.index), "Failed to append index");
+    };
+
+    checkStatus(collectionBuilder->Append(), "Failed to append to collectionBuilder");
+    for (size_t i = 0; i < concreteColl->size(); ++i) {
+      auto obj = (*concreteColl)[i];
```
jmcarcell:
```suggestion
    for (const auto& obj: concreteColl) {
```
I think here you can use a range-based for loop since you don't need i

tests/unittests/test_arrow_converter.cpp
jmcarcell:
```suggestion
```
Not used

src/ArrowFrameData.cc
arnavdham: String "1"

python/templates/ArrowMapper.cc.jinja2
arnavdham: Done like this

tests/unittests/test_arrow_converter.cpp
arnavdham: removed

src/ArrowFrameData.cc
arnavdham: I've refactored extractMap into a single template that takes ArrayType and ValueType directly.

tests/unittests/test_arrow_converter.cpp
arnavdham: used `getAvailableCollections` now

jmcarcell (approved):

jmcarcell:
Can you format the PR to have the text between BEGINRELEASENOTES and ENDRELEASENOTES like in other PRs? We have a script that reads those and puts them in a doc file and commits it when tagging. I could do it myself, but it would be good if you do that for this and other PRs.

arnavdham:
> Can you format the PR to have the text between BEGINRELEASENOTES and ENDRELEASENOTES like in other PRs? We have a script that reads those and puts them in a doc file and commits it when tagging. I could do it myself, but it would be good if you do that for this and other PRs.

Done for this PR. Will follow this for any further PRs as well. Do you want me to do this for already merged PRs as well?

jmcarcell:
Yes, please, I may have forgotten to say so in other PRs.

PR #990, opened 2026-07-21
Clean up SIO and Arrow configuration export

arnavdham:
BEGINRELEASENOTES
- Cleans up the package configuration by exporting only the prefixed variables (`PODIO_ENABLE_SIO` and `PODIO_ENABLE_ARROW`) and check both the prefix and the unprefixed ones.

ENDRELEASENOTES

Should fix the CI/CD failures in https://github.com/key4hep/EDM4hep/pull/502


jmcarcell:
I have a different proposal: export in both cases `PODIO_ENABLE_<>` (maybe better all in caps?) and in the functions, check first if `PODIO_ENABLED` is defined, if they are defined then use their value, otherwise use `ENABLE_`. That also allows you from a downstream project to modify the value of `PODIO_ENABLE` to disable the component if you really wanted to (even if podio is built with Arrow, maybe not everyone wants to build EDM4hep with Arrow, for example). This is currently not possible I think. Then the checks for SIO headers can be removed.

cmake/podioMacros.cmake
jmcarcell:
```suggestion
```
This is not needed anymore. Can you add a status message like this one when `PODIO_ENABLE_<>` is not set saying that we are not going to build SIO/Arrow?

jmcarcell:
dev3 failures will be fixed after a rebase (fixed in https://github.com/AIDASoft/podio/pull/993)

cmake/podioMacros.cmake
```diff
@@ -307,6 +307,16 @@ endfunction()
...
+  if(DEFINED PODIO_ENABLE_SIO)
+    if(NOT PODIO_ENABLE_SIO)
+      message(STATUS "Not adding the SIO Blocks library to the targets because PODIO_ENABLE_SIO is not set")
+      return()
+    endif()
+  elseif(NOT ENABLE_SIO)
+    message(STATUS "Not adding the SIO Blocks library to the targets because ENABLE_SIO is not set")
+    return()
+  endif()
+
```
jmcarcell:
```suggestion
if((DEFINED PODIO_ENABLE_SIO AND NOT PODIO_ENABLE_SIO) OR (NOT DEFINED PODIO_ENABLE_SIO AND NOT ENABLE_SIO))
  message(STATUS "Not adding the SIO Blocks library to the targets because SIO support is not enabled")
  return()
endif()

```
Avoid the double message when not enabled

cmake/podioMacros.cmake
```diff
@@ -348,7 +353,13 @@ endfunction()
...
-  if(NOT ENABLE_ARROW)
+  if(DEFINED PODIO_ENABLE_ARROW)
+    if(NOT PODIO_ENABLE_ARROW)
+      message(STATUS "Not adding the Arrow library to the targets because PODIO_ENABLE_ARROW is not set")
+      return()
+    endif()
+  elseif(NOT ENABLE_ARROW)
+    message(STATUS "Not adding the Arrow library to the targets because ENABLE_ARROW is not set")
     return()
   endif()
```
jmcarcell:
```suggestion
if((DEFINED PODIO_ENABLE_ARROW AND NOT PODIO_ENABLE_ARROW) OR (NOT DEFINED PODIO_ENABLE_ARROW AND NOT ENABLE_ARROW))
  message(STATUS "Not adding the Arrow library to the targets because Arrow support is not enabled")
  return()
endif()

```

PR #996, opened 2026-07-30
Expose event weight names through DataSource

ZeynepEdaCabukoglu:
BEGINRELEASENOTES

Expose EventWeightNames metadata through podio::DataSource.
Store the weight names from the metadata frame and make them available to RDataFrame through the _EventWeightNames column.
This allows analyses to access event-weight labels without opening the input file separately.

ENDRELEASENOTES


tmadlener (commented):
I like the idea, but I am wondering whether this can be made more generic on the podio side and put it together specifically for the event weights on the FCCAnalyses side?

Issue #403, opened 2023-04-04
PODIO implicit plug-in mechanism is not robust enough


tmadlener:
PODIO has an implicit plug-in mechanism that is necessary to load I/O functionality for EDMs that have been generated. There are several shared libraries that are loaded at runtime only, and for which different plug-in mechanisms exist within PODIO. All of them essentially search `LD_LIBRARY_PATH` which makes it very easy to end up in a "polluted" environment if one is not careful enough. This bites us regularly in e.g. CI, especially if breaking changes are introduced and the underlying environment has not yet caught up.

Currently the plug-in mechanisms are
- ROOT dictionary loading where we have no real control, since this is all happening in ROOT
- SIO Block library loading. Here we have all control via the [`SIOBlockLibraryLoader`](https://github.com/AIDASoft/podio/blob/09d17d49f434e23663137eadacfea4eaa3d58d48/include/podio/SIOBlock.h#L242-L267), where we currently explicitly search for libraries on `LD_LIBRARY_PATH`:
https://github.com/AIDASoft/podio/blob/09d17d49f434e23663137eadacfea4eaa3d58d48/src/SIOBlock.cc#L163-L178

PR #999, opened 2026-08-04
Refactor backend library loading and rename Arrow suffix to PodioArrow

arnavdham:
Make Arrow mapper registration follow the same runtime-loading model as SIO, while refactoring the underlying library loading logic for both backends into a generalized, thread-safe loader. 
- Consolidates dynamic library loading into a single `podio::utilities::BackendLibraryLoader` used by both SIO and Arrow.
- Ensures the library loader initialization is strictly thread-safe using C++11 magic statics, fixing potential data races.
- Renames the generated Arrow plugin suffix from `Arrow` to `PodioArrow` (e.g. `libedm4hepPodioArrow.so`) to avoid potential clashes with `libarrow.so` during dynamic `LD_LIBRARY_PATH` scanning.

BEGINRELEASENOTES
- **Refactoring:** Generalized dynamic library loading for I/O backends into a shared, thread-safe `BackendLibraryLoader`.
- **Change:** Changed the library suffix generated by `PODIO_ADD_ARROW` from `Arrow` to `PodioArrow` to prevent dynamic loading collisions. Downstream projects will need to update their CMake aliases accordingly.
- **Fix:** Arrow mapper libraries are now loaded dynamically at registry lookup time instead of relying on linker side-effects.

ENDRELEASENOTES


include/podio/utilities/ArrowConverterRegistry.h
```diff
@@ -20,6 +20,10 @@ class CollectionBase;
 /**
  * @brief Global singleton registry mapping PODIO type name strings to their
  * corresponding Apache Arrow array converter callbacks.
+ *
+ * Registration is expected to happen during shared library loading/startup,
+ * before worker threads query the registry. After registration the registry is
```
veprbl: A bit misleading, the loading happens lazily on `getConverter`/`getReader` calls, so users must not use those in multi-threaded context.

src/ArrowConverterRegistry.cc
```diff
@@ -28,11 +34,74 @@ void ArrowConverterRegistry::registerReader(const std::string& typeName, BufferR
...
+  std::string dir;
+  std::istringstream stream(ldLibPath);
+  while (std::getline(stream, dir, ':')) {
+    if (not fs::exists(dir)) {
+      continue;
+    }
+
+    for (auto& lib : fs::directory_iterator(dir)) {
+      const auto filename = lib.path().filename().string();
+      if (filename.find("Arrow") != std::string::npos && filename.find("libarrow") == std::string::npos) {
```
veprbl: Why can it be either one? `libarrow` must come from arrow-cpp, which `libpodioArrow` is meant to link against, so it gets loaded as a dependency.

include/podio/utilities/ArrowConverterRegistry.h
arnavdham: But this initialization is thread safe. Once lazy loading is complete MT is completely fine.

veprbl: Right, so it does matter when lazy loading happens. Races can occur if several threads start up and simultaneously try to initialize reader.

src/ArrowConverterRegistry.cc
jmcarcell: Maybe check for `ArrowMapper`?

src/ArrowConverterRegistry.cc
```diff
@@ -28,11 +34,74 @@ void ArrowConverterRegistry::registerReader(const std::string& typeName, BufferR
...
   }
   return nullptr;
 }
 
+ArrowConverterLibraryLoader::ArrowConverterLibraryLoader() {
+  for (const auto& [lib, dir] : getLibNames()) {
+    const auto status = loadLib(lib, dir);
+    switch (status) {
+    case LoadStatus::Success:
+      break;
```
jmcarcell: Logging here is different compared to SIO, is there any reason for that?

src/ArrowConverterRegistry.cc
```diff
@@ -28,11 +34,74 @@ void ArrowConverterRegistry::registerReader(const std::string& typeName, BufferR
...
+    case LoadStatus::AlreadyLoaded:
+      break;
+    case LoadStatus::Error:
+      std::cerr << "ERROR while loading Arrow library \'" << lib << "\' (from " << dir << ")" << std::endl;
```
jmcarcell: SIO doesn't do it but `dlerror()` gives you the reason why it failed so it could be included in the message.

src/ArrowTypeRegistry.cc
```diff
@@ -16,6 +17,7 @@ void ArrowTypeRegistry::registerType(const std::string& typeName, std::shared_pt
 }
 
 std::shared_ptr<arrow::DataType> ArrowTypeRegistry::getType(const std::string& typeName) const {
+  ArrowConverterLibraryLoader::instance();
```
jmcarcell: Compared to SIO, this runs on every lookup, also in `getConverter` and `getReader`. Shouldn't it only run once in the entrypoint (`getReader` I think should be the equivalent for SIO)?

src/ArrowConverterRegistry.cc
veprbl: I believe, we agreed not to name the library Mapper, but simply Arrow.

include/podio/utilities/ArrowConverterRegistry.h
arnavdham: We can discuss this in our meet but I think the arrowConverterLibraryLoader will only be initialsed once and as it is a local static variable it is thread safe. https://stackoverflow.com/questions/8102125/is-local-static-variable-initialization-thread-safe-in-c11

veprbl: This makes sense now. We agreed, the text can be improved to specify which library this sentence is talking about (libpodioArrow or libedm4hepArrow etc)

arnavdham: Better comment added.

src/ArrowConverterRegistry.cc
arnavdham: Only checking for Arrow now.

src/ArrowConverterRegistry.cc
arnavdham: Made the logging consistent

src/ArrowConverterRegistry.cc
arnavdham: Added `dlerror()`

src/ArrowTypeRegistry.cc
arnavdham: It guarantees the plugins are loaded even if a user or script queries the ArrowTypeRegistry or ArrowConverterRegistry directly before performing a conversion. Also,the overhead is negligible even if it runs multiple times.

src/ArrowConverterRegistry.cc
```diff
@@ -28,11 +34,79 @@ void ArrowConverterRegistry::registerReader(const std::string& typeName, BufferR
...
+  std::string dir;
+  std::istringstream stream(ldLibPath);
+  while (std::getline(stream, dir, ':')) {
+    if (not fs::exists(dir)) {
+      continue;
+    }
+
+    for (auto& lib : fs::directory_iterator(dir)) {
+      const auto filename = lib.path().filename().string();
+      if (filename.find("Arrow") != std::string::npos) {
```
jmcarcell: I think this is ok because of the capital `A`; in a view the directories in `LD_LIBRARY_PATH` may contain lots of libraries, and in particular `libarrow.so`, which doesn't match. But if there were multiple podio datamodels that means all the arrow libraries are opened. This is just a comment, since I had to think about this case and check which names are used in LCG stacks (there doesn't seem to be anything else with `Arrow` in their name in the top level folders).

jmcarcell:
Any more comments @veprbl? Looks good to me as it is

include/podio/utilities/ArrowConverterRegistry.h
veprbl: I think this one is not correct, libpodioArrow.so is not a data model, so registry does not get updated when that one is loaded (it may contain the main symbol for the registry).

arnavdham: So I just remove libpodioArrow.so as example?

src/ArrowConverterRegistry.cc
tmadlener: You could also give this a slightly less generic suffix that is less likely to clash. That might require to change parts of the code gen in some cmake macro I would assume (without having checked).

tmadlener (commented):
At least the library loading part looks like an almost copy of what we do for SIO (i.e. `SIOBlockLibraryLoader` is essentially `ArrowConverterLibraryLoader` with some string replacements). Can those two be generalized into one configurable library loader that works for both? Configuration would probably take the env variable name, a library pattern and probably some designator for the log messages.

include/podio/utilities/ArrowConverterRegistry.h
arnavdham: Removed.

src/ArrowConverterRegistry.cc
arnavdham: For this to work, I think we will have to make changes downstream(edm4hep) as well so I would suggest keeping it as Arrow for simplicity.

arnavdham:
> At least the library loading part looks like an almost copy of what we do for SIO (i.e. `SIOBlockLibraryLoader` is essentially `ArrowConverterLibraryLoader` with some string replacements). Can those two be generalized into one configurable library loader that works for both? Configuration would probably take the env variable name, a library pattern and probably some designator for the log messages.

I could do this.

include/podio/utilities/ArrowConverterRegistry.h
veprbl: This looks good, thank you.

src/ArrowConverterRegistry.cc
veprbl: A more impactful improvement would be to load only a relevant library for the model(s) used in a file, which would likely also resolve this.

arnavdham: I don't think getting the library would be easy. Also we don't have files right now as everything is in memory

tmadlener: > For this to work, I think we will have to make changes downstream(edm4hep) as well so I would suggest keeping it as Arrow for simplicity.

I don't think there has been a podio release yet with the functionality, so the most we would break at this point is some nightly builds for which we don't make too many guarantees. So if we want to make this change, now is essentially the time.

> A more impactful improvement would be to load only a relevant library for the model(s) used in a file, which would likely also resolve this.

There is generally room for improvement here, see also: https://github.com/AIDASoft/podio/issues/403 I don't think we need to worry too much about performance here in general. There have been some interesting interactions with cvmfs and root dictionary loading, but realistically people will have only a (very) small number of datamodels to consider and even the fallback of going through `LD_LIBRARY_PATH` will only do a scan for matching libraries once.

include/podio/utilities/LibraryLoader.h
```diff
@@ -0,0 +1,32 @@
...
+
+#include <map>
+#include <string>
+#include <tuple>
+#include <vector>
+
+namespace podio {
+namespace utilities {
+
+  class LibraryLoader {
```
tmadlener: Maybe `BackendLibraryLoader` to make it slightly more specific?

src/ArrowConverterRegistry.cc
```diff
@@ -1,4 +1,10 @@
 #include "podio/utilities/ArrowConverterRegistry.h"
+#include "podio/utilities/LibraryLoader.h"
+#include <cstdlib>
+#include <dlfcn.h>
+#include <filesystem>
+#include <iostream>
+#include <sstream>
```
tmadlener: Some of these could now be obsolete as they moved to the `LibraryLoader` part, right?

src/SIOBlock.cc
```diff
@@ -1,9 +1,11 @@
 #include "podio/SIOBlock.h"
 
+#include "podio/utilities/LibraryLoader.h"
 #include <algorithm>
 #include <cstdlib>
 #include <dlfcn.h>
 #include <filesystem>
+#include <iostream>
 #include <map>
 #include <sstream>
```
tmadlener: Similar comment as for the Arrow part; I don't think we need all of these now.

tmadlener (commented):
Thanks for pulling out the generic bits, looks good to me.

I am wondering whether we want the fallback to `LD_LIBRARY_PATH` for Arrow as well. For SIO it's essentially mainly there because we need it for some backwards compatibility, which is not really a concern yet for Arrow.

src/ArrowConverterRegistry.cc
veprbl: > I don't think there has been a podio release yet with the functionality, so the most we would break at this point is some nightly builds for which we don't make too many guarantees. So if we want to make this change, now is essentially the time.

Would you like to suggest a name? We had `ArrowMapper` which I did not like at the time.

> 
> There is generally room for improvement here, see also: #403 I don't think we need to worry too much about performance here in general.

Right, my point was that loading an irrelevant Arrow library is on the same scale of a problem.

> I don't think getting the library would be easy. Also we don't have files right now as everything is in memory

A library name can go to Table metadata.

tmadlener: > Would you like to suggest a name? We had `ArrowMapper` which I did not like at the time.

I don't have a strong opinion on this one. One could even do `PodioArrow` (or any variation of that) to be very specific. Given that all the builds and environment building are pretty well automated, I don't think too many people will look at the generated library names in any case.

src/SIOBlock.cc
arnavdham: Removed.

src/ArrowConverterRegistry.cc
arnavdham: Removed.

include/podio/utilities/LibraryLoader.h
arnavdham: Renamed.

arnavdham:
> Thanks for pulling out the generic bits, looks good to me.
> 
> I am wondering whether we want the fallback to `LD_LIBRARY_PATH` for Arrow as well. For SIO it's essentially mainly there because we need it for some backwards compatibility, which is not really a concern yet for Arrow.

So no scanning if podio_arrow_path is not set? Maybe bool allowFallbackToLDLibraryPath can be added to the constructor of BackendLibraryLoader set to true for SIO and false for arrow? @veprbl @tmadlener

tmadlener:
> So no scanning if podio_arrow_path is not set?

Yes, exactly. I have no strong opinion here.

>  Maybe bool allowFallbackToLDLibraryPath can be added to the constructor of BackendLibraryLoader set to true for SIO and false for arrow?

If we decide to not have the fallback for arrow that would be the technical way of doing it probably.

veprbl:
I'm not in favor of removing LD_LIBRARY_PATH. If anything, non-standard PODIO_ARROW_PATH should be retired.

arnavdham:
> I'm not in favor of removing LD_LIBRARY_PATH. If anything, non-standard PODIO_ARROW_PATH should be retired.

So I keep the current implementation and just change `Arrow `to `PodioArrow ` to avoid clashes and similar change in edm4hep and its should be fine? @veprbl @tmadlener

veprbl:
> > I'm not in favor of removing LD_LIBRARY_PATH. If anything, non-standard PODIO_ARROW_PATH should be retired.
> 
> So I keep the current implementation and just change `Arrow `to `PodioArrow ` to avoid clashes and similar change in edm4hep and its should be fine? @veprbl @tmadlener

Let's do that, please.

tmadlener:
> So I keep the current implementation and just change `Arrow` to `PodioArrow`  to avoid clashes and similar change in edm4hep and its should be fine?

I don't think any changes in edm4hep would be necessary? Shouldn't a corresponding change in the `PODIO_ADD_ARROW` macro already propagate through all the machinery transparently?

arnavdham:
> > So I keep the current implementation and just change `Arrow` to `PodioArrow`  to avoid clashes and similar change in edm4hep and its should be fine?
> 
> I don't think any changes in edm4hep would be necessary? Shouldn't a corresponding change in the `PODIO_ADD_ARROW` macro already propagate through all the machinery transparently?

https://github.com/key4hep/EDM4hep/pull/502/changes the changes made here would have to be renamed from edm4hepArrow to edm4hepPodioArrow

arnavdham:
Have renamed to PodioArrow

tmadlener:
Ah yes, missed the fact that we use the target names directly there. In that case this would need a change in EDM4hep as well.

arnavdham:
https://github.com/key4hep/EDM4hep/pull/509 Change in edm4hep

src/ArrowConverterRegistry.cc
```diff
@@ -28,11 +30,16 @@ void ArrowConverterRegistry::registerReader(const std::string& typeName, BufferR
 }
 
 ArrowConverterRegistry::BufferReaderFunc ArrowConverterRegistry::getReader(const std::string& typeName) const {
+  loadArrowLibraries();
   auto it = m_readerRegistry.find(typeName);
   if (it != m_readerRegistry.end()) {
     return it->second;
   }
   return nullptr;
 }
```
tmadlener, line 39: Sorry for being rather slow here, but IIUC, this might happen on multiple threads concurrently here. The `getReader` (potentially also the `getConverter`, but there I haven't fully understood the logic yet, I think) is called in a section which podio assumes is safe to be called from multiple threads. In SIO we load the libraries during the construction of the reader. This can't be placed into the `ArrowConverterRegistry` constructor because we will actually try to call into that from the loaded libraries, I think. But maybe there is some place where we have slightly less potential problems with trying to load from multiple threads.

veprbl: That was the question we had too. In the end, it seems like Arnav's implementation relies on a guarantee that static local variable initialization is thread safe in C++.

tmadlener: Ah yes, good point. I missed the `static` assignment in the `loadXYZLibraries`. In that case I think threading should pose no issues and, I think this also gives us a guarantee of happening exactly once.

arnavdham: So ig this can be merged now if there are no more comments?

tmadlener (approved):

PR #1001, opened 2026-08-08
Optimize Arrow Backend Read Path

arnavdham:
This PR introduces two performance optimizations to the Arrow backend read path (`registerBufferReaders`), collectively improving the Arrow stream rate and significantly dropping the duration of the `Materialize` stage (the final stage of the read path where raw Arrow arrays are converted back into usable C++ PODIO collections).

* **Fix 1: Add `.reserve()` to vectors** 
  We were doing `push_back`s into vectors, causing reallocation overhead as PODIO reconstructed collections. We now `.reserve()` the exact capacity using `arr->values()->length()`. This bumped the overall throughput by **~5%**.
  
* **Fix 2: Hoist Arrow schema lookups** 
  `GetFieldByName` performs expensive `std::string` schema searches and was mistakenly placed inside the per-element `for` loop. We hoisted all Arrow array lookups completely outside the loop so they are only called once per collection. This dropped the `Materialize` execution time by an additional **~10-11%**, as C++ objects can now be materialized without redundant string allocations.

BEGINRELEASENOTES
- Optimized the Arrow backend read path by eliminating vector reallocation overhead and redundant O(N) schema lookups.

ENDRELEASENOTES

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -87,3 +87,6 @@ cmake-build*/
 
 # Created by tests
 /tools/Digraph.gv
+
+# Benchmark outputs
+/perf/
--- a/.gitignore
+++ b/.gitignore
@@ -87,6 +87,3 @@ cmake-build*/
 
 # Created by tests
 /tools/Digraph.gv
-
-# Benchmark outputs
-/perf/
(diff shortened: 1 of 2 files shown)


arnavdham:
Please take a look @jmcarcell

arnavdham:
As juan is on leave, if you could take a look @tmadlener

tmadlener (approved):
So this is a roughly 15% improvement over previously, IIUC? Nice. I have no comments, this all looks like it does what is described in the PR.

Issue #945, opened 2026-03-12
Inconsistent use of strings and string_view in Categories


wdconinc:
There is a bit of an inconsistency on how one works with Categories on Frames:
- `getAvailableCategories` returns `std::vector<std::string_view>`: https://github.com/AIDASoft/podio/blob/2c2a0f5d698f7bbe45b648c1cb25546eb2742b53/include/podio/Reader.h#L32
- but then `readFrame` and `getEntries` on the category requires a `string` anyway: https://github.com/AIDASoft/podio/blob/2c2a0f5d698f7bbe45b648c1cb25546eb2742b53/include/podio/Reader.h#L27-L29

We're not clear on what the motivation for this choice is, but it forces us to do an explicit conversion where we feel we could just as well continue using `string_view`:
```cpp
    for (const auto& _category : podio_source->getAvailableCategories()) {
      std::string category = _category;
      std::size_t n = podio_source->getEntries(category);
    }
```

jmcarcell:
Fixed with https://github.com/AIDASoft/podio/pull/946.

PR #1003, opened 2026-08-15
fix: avoid dangling string_view keys in RNTupleReader maps

wdconinc:
## Dependencies
- [ ] #1004

## Summary

Several `std::unordered_map` members in `RNTupleReader` use `std::string_view` keys (introduced in #946). Maps populated in `openFiles()` are safe because their keys point into the stable `m_availableCategories` vector. However, `m_collectionInfo`, `m_idTables`, and `m_entries` were populated lazily from `initCategory()` and `readEntry()` using the **caller's transient `string_view category` parameter** as the map key.

When the caller's backing string is destroyed (e.g. a local `std::string` at the end of a loop iteration, ironically the example in #945), the stored key becomes dangling. This produces UB in the `unordered_map` — ultimately `ROOTFrameData::getIDTable()` crashes when dereferencing a null ID table created from a skipped `initCategory`.

BEGINRELEASENOTES
- fix: avoid dangling string_view keys in RNTupleReader maps

ENDRELEASENOTES

## Backtrace
```
0: podio::CollectionIDTable::CollectionIDTable(std::vector<uint32_t> const&, std::vector<std::string> const&)
1: podio::ROOTFrameData::getIDTable() const
2: podio::Frame::FrameModel<podio::ROOTFrameData>::FrameModel(std::unique_ptr<podio::ROOTFrameData>)
3: podio::Reader::ReaderModel<podio::RNTupleReader>::readFrame(std::string_view, unsigned long, std::vector<std::string> const&)
```

## Fix

- `initCategory()`: retain the iterator from `std::ranges::find` and derive a stable `string_view` into `m_availableCategories` to use as the map key.
- `readEntry()`: reuse the `m_collectionInfo.find()` iterator to obtain the stable key without a second search.
- `readNextEntry()`: use `m_totalEntries.find()` (O(1), stable keys) instead of a linear scan over `m_availableCategories`.

## Depends on

This branch is rebased on #1004, which adds the ASAN regression test that demonstrates the bug. Once #1004 merges to master, this PR's diff reduces to only `src/RNTupleReader.cc` changes.


wdconinc:
Closing — not ready for upstream review yet.

wdconinc:
This PR fixes the failure in https://eicweb.phy.anl.gov/EIC/benchmarks/detector_benchmarks/-/jobs/8292678#L3468, as seen in the CI run that uses this branch: https://eicweb.phy.anl.gov/EIC/benchmarks/detector_benchmarks/-/jobs/8296023

tmadlener (approved):

PR #1005, opened 2026-08-15
fix(cmake): typo USE_SANITIZE => USE_SANITIZER

wdconinc:
This just fixes a typo in a CMake condition.

BEGINRELEASENOTES
- fix(cmake): typo USE_SANITIZE => USE_SANITIZER

ENDRELEASENOTES


jmcarcell (approved):

jmcarcell:
Checked that the typo doesn't happen in other places

wdconinc:
> Checked that the typo doesn't happen in other places

There is the CMake option `-warn-uninitialized` which should flag this, but it didn't seem to pick up this instance for some reason when I was testing it. That would be a systemic way to find any incorrect usage of CMake variables.

PR #1007, opened 2026-08-21
Add a comparison operator for ObjectID

jmcarcell:
BEGINRELEASENOTES
 - Add a comparison operator for ObjectID, allowing ObjectIDs to be used in `std::map` containers and not only `std::unordered_map`.
 - Add a test for `ObjectID` ordering and `std::map` usage.

ENDRELEASENOTES

Allows also to do several things for debugging, like filling a map to do something with the elements with deterministic ordering (and ordered within a collection).

### DIFF
--- a/include/podio/ObjectID.h
+++ b/include/podio/ObjectID.h
@@ -31,8 +31,7 @@ public:
   constexpr bool operator==(const ObjectID&) const noexcept = default;
 
   /// Provide an order solely for use in ordered containers.
-  /// Order ObjectIDs by collection first and then by their index within the collection.
-  /// Only comparisons between ObjectIDs from the same collection make sense.
+  /// Order ObjectIDs by collectionID first and then by their index within the collection.
   constexpr std::strong_ordering operator<=>(const ObjectID& other) const noexcept {
     if (const auto comparison = collectionID <=> other.collectionID; comparison != 0) {
       return comparison;
--- a/include/podio/ObjectID.h
+++ b/include/podio/ObjectID.h
@@ -30,7 +30,9 @@ public:
   /// this operator is necessary for meaningful comparisons in python
   constexpr bool operator==(const ObjectID&) const noexcept = default;
 
+  /// Provide an order solely for use in ordered containers.
   /// Order ObjectIDs by collection first and then by their index within the collection.
+  /// Only comparisons between ObjectIDs from the same collection make sense.
   constexpr std::strong_ordering operator<=>(const ObjectID& other) const noexcept {
     if (const auto comparison = collectionID <=> other.collectionID; comparison != 0) {
       return comparison;
--- a/include/podio/ObjectID.h
+++ b/include/podio/ObjectID.h
@@ -27,7 +27,6 @@ public:
   uint32_t collectionID{static_cast<uint32_t>(untracked)};
 
   /// index and collectionID uniquely defines the object.
-  /// this operator is necessary for meaningful comparisons in python
   constexpr bool operator==(const ObjectID&) const noexcept = default;
 
   /// Provide an order solely for use in ordered containers.
--- a/include/podio/ObjectID.h
+++ b/include/podio/ObjectID.h
@@ -1,6 +1,7 @@
 #ifndef PODIO_OBJECTID_H
 #define PODIO_OBJECTID_H
 
+#include <compare>
 #include <cstdint>
 #include <functional>
 #include <iomanip>
@@ -28,6 +29,14 @@ public:
   /// index and collectionID uniquely defines the object.
   /// this operator is necessary for meaningful comparisons in python
   constexpr bool operator==(const ObjectID&) const noexcept = default;
+
+  /// Order ObjectIDs by collection first and then by their index within the collection.
+  constexpr std::strong_ordering operator<=>(const ObjectID& other) const noexcept {
+    if (const auto comparison = collectionID <=> other.collectionID; comparison != 0) {
+      return comparison;
+    }
+    return index <=> other.index;
+  }
 };
 
 inline std::ostream& operator<<(std::ostream& os, const podio::ObjectID& id) {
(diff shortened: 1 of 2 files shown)


include/podio/ObjectID.h
```diff
@@ -28,6 +29,14 @@ class ObjectID {
   /// index and collectionID uniquely defines the object.
   /// this operator is necessary for meaningful comparisons in python
   constexpr bool operator==(const ObjectID&) const noexcept = default;
+
+  /// Order ObjectIDs by collection first and then by their index within the collection.
```
tmadlener: Maybe add here that this is mainly meant to provide ordering for certain containers, and that comparisons of object ids are about equally meaningless in terms of data as comparisons between handles? Arguably ObjectIDs **from the same collection** might be a bit better, but I think we should avoid that people start to rely on anything here.

jmcarcell: Done

tmadlener (dismissed):

include/podio/ObjectID.h
m-fila: I think I know what you meant but the phrasing now is confusing. 
>  Provide an order solely for use in ordered containers.
> Only comparisons between ObjectIDs from the same collection make sense.

reads as if the contract was that only objects from the same collection can be compared  and put in an associative container

tmadlener: What about simply dropping that sentence? *Order ObjectIDs by collectionID first and then by their index within the collection.* describes pretty well what is happening, I think.

m-fila: Yes, I think without that sentence it's less confusing

include/podio/ObjectID.h
```diff
@@ -28,6 +29,16 @@ class ObjectID {
   /// index and collectionID uniquely defines the object.
   /// this operator is necessary for meaningful comparisons in python
   constexpr bool operator==(const ObjectID&) const noexcept = default;
+
+  /// Provide an order solely for use in ordered containers.
+  /// Order ObjectIDs by collection first and then by their index within the collection.
+  /// Only comparisons between ObjectIDs from the same collection make sense.
```
tmadlener:
```suggestion
  /// Order ObjectIDs by collectionID first and then by their index within the collection.
```

m-fila (approved):

include/podio/ObjectID.h
```diff
@@ -28,6 +29,15 @@ class ObjectID {
...
   constexpr bool operator==(const ObjectID&) const noexcept = default;
+
+  /// Provide an order solely for use in ordered containers.
+  /// Order ObjectIDs by collectionID first and then by their index within the collection.
+  constexpr std::strong_ordering operator<=>(const ObjectID& other) const noexcept {
+    if (const auto comparison = collectionID <=> other.collectionID; comparison != 0) {
+      return comparison;
+    }
+    return index <=> other.index;
+  }
```
tmadlener, line 39:
```suggestion
  constexpr auto operator<=>(const ObjectID& other) const noexcept = default;
```
Does this not do the exact same thing, actually?

jmcarcell: `index` is declared before `collectionID` so no? Or we can change the order

tmadlener: Aah, I missed that. Does it matter that we order with a different priority? Not sure we can change the order without breaking things in other places.

include/podio/ObjectID.h
```diff
@@ -28,6 +29,15 @@ class ObjectID {
   /// index and collectionID uniquely defines the object.
   /// this operator is necessary for meaningful comparisons in python
   constexpr bool operator==(const ObjectID&) const noexcept = default;
```
tmadlener, line 30: This can go once we have `operator<=>`

m-fila: Non-defaulted operator `<=>` doesn't generate `==`. The part about Python should go

jmcarcell: Removed the sentence about python, it was probably the reason why `==` was added.

include/podio/ObjectID.h
jmcarcell: https://en.cppreference.com/cpp/language/default_comparisons#Default_comparison_order it's there "in declaration order" for non-static. For the order I would imagine it doesn't change much since the bytes are the same (although no one should be using the bytes) and then the name doesn't change for TTrees and RNTuples. I'm not sure if it's worth it.

tmadlener: Given that the ordering is introduced for essentially only technical reasons, I would prefer to just go with a defaulted `operator<=>` without changing the order of the members. We would have to adjust the docstring and could then also remove the `operator==` declaration since that will be covered by the defaulted `operator<=>` as well.

m-fila: I'm not sure if cppyy will be able to figure out what to do with just `<=>`

tmadlener: I would have liked to say, we will see in CI, but I can't find any code path in python that covers comparison of object ids at all.

jmcarcell: > Given that the ordering is introduced for essentially only technical reasons

Iterating over a sorted set of ObjectID means you iterate over one collection at a time, which is a nice property.

> I'm not sure if cppyy will be able to figure out what to do with just `<=>`

In python, comparisons with `<=>` raise a `NotImplementedError` (so all except `==` and `!=`). Given that the purpose of this is to fill `std::map` or `std::set` and they don't need the comparison in Python it may not be that bad. But sorting any number of `ObjectID` will fail.

So we can keep `<=>` and if it's ever needed we implement all the comparison operators that are needed in Python (`<` only?) or we go directly to implement all the operators for Python.

tmadlener: > Iterating over a sorted set of ObjectID means you iterate over one collection at a time, which is a nice property.

Fair enough. So no we go with what is currently present, right?

> if it's ever needed we implement all the comparison operators that are needed in Python (`<` only?) or we go directly to implement all the operators for Python.

I would go with do nothing now and only if these are ever requested / needed do something.

jmcarcell: Then this is good to go on my side. THere is also the possibility cppyy (ROOT's version) catches up with `<=>` in the meantime.

tmadlener (approved):

PR #1011, opened 2026-08-26
Add a create factory method for link collections that take objects to link

### LINKED ISSUE #1008: Variadic create for Link collections
Datatype collections have a variadic `create` function. LinkCollections do not.

This would allow it to turn this
```cpp
auto coll = SomeLinkCollection{};
auto from = MutableFromType{};
auto to = MutableToType{};
auto link = coll.create();
link.set(from);
link.set(to);
```

into this

```cpp
auto coll = SomeLinkCollection{};
auto from = MutableFromType{};
auto to = MutableToType{};
auto link = coll.create(from, to); // potentially create({from, to});
```

For cases where the `FromT` and `ToT` are different `create` should probably accept any order, for the rest I would go with the "canonical order" of _from first, to second_.

tmadlener:
BEGINRELEASENOTES
- Add a `LinkCollection::create` overload that takes the objects that should be linked to facilitate creating links.

ENDRELEASENOTES

Fixes #1008

### DIFF
--- a/doc/links.md
+++ b/doc/links.md
@@ -87,6 +87,21 @@ auto recoP = link.getTo();
 auto weight = link.getWeight();
 ```
 
+The objects that should be linked (and optionally the weight) can also be
+passed directly to `create`
+```cpp
+auto link = mcRecoLinks.create(mcParticle, recoParticle, 1.0);
+```
+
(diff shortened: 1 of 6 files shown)


tmadlener:
Unless there are comments, I will merge this later today to have it in the next tag.

PR #1015, opened 2026-08-31
fix: propagate dependency versions downstream via CMake

wdconinc:
This PR fixes a minor issue reported in https://github.com/acts-project/acts/pull/5983#issuecomment-5476193435, where nlohmann-json defines a compatibility target if no explicit version is passed in `find_dependency`.

BEGINRELEASENOTES
- fix: add nlohmann_json_VERSION to find_dependency

ENDRELEASENOTES


wdconinc:
FYI @andiwand

tmadlener (commented):
Since we are at it (and it's just adjacent). Do we need the same procedure for Arrow & Parquet?

wdconinc:
> Since we are at it (and it's just adjacent). Do we need the same procedure for Arrow & Parquet?

It probably wouldn't be bad to add a min version (I don't know which one) and then propagate the found version as a dependency minimum, but arrow and parquet don't have the compatibility fallback that nlohmann-json has so it is not leading to downstream issues.

wdconinc:
@arnavdham do you know what minimum versions of arrow and parquet are required?

tmadlener:
I think as it is done now, this will typically resolve to a minimal version. It's all a bit dependent how each dependency sets its version compatibility themselves (hopefully via [`write_basic_package_version_file`](https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html#generating-a-package-version-file).

So, I think what we pass on from here at least makes sure that whatever version podio was built with, we get a compatible version if someone builds against podio. And then it's effectively up to podio / arrow / nlohmann::json to do the right thing.

jmcarcell:
Looks good to me. In LCG stacks I haven't seen this issue (or in ACTS) appear.

tmadlener:
I have applied the same treatment for Arrow and Parquet and confirmed that the resulting `podioConfig.cmake` contains the expected versions.

Unless there are complaints, I will merge this later today.

PR #1016, opened 2026-09-01
Report podio version during CMake package discovery

jmcarcell:
BEGINRELEASENOTES
 - Report the installed podio version when found by CMake.

ENDRELEASENOTES


tmadlener (approved):
Confirmed that this is properly picked up via the EDM4hep workflows which now report the version of podio.

PR #1018, opened 2026-09-03
Switch python tooling to ruff

### LINKED ISSUE #703: Harmonize python tool configuration
Currently we have a `.flake8`, a `setup.cfg` and a `pyproject.toml` with partially contradictory settings. These should all be in one place only and consistent.

mohitt31:
The python tool configuration was spread over three files that disagreed with each other. `.flake8` set the line length to 99, `setup.cfg` carried a second `[flake8]` section setting it to 120 plus `[pycodestyle]` and `[pep8]` sections for tools that aren't run at all, and `pyproject.toml` configured black at 99. Since the pre-commit hook passes `--config=.flake8`, the whole setup.cfg block was dead config. Nothing outside those files referenced them and there's no setup.py, so they're removed and the configuration is one `.ruff.toml`, copied from the one EDM4hep already uses.

A few things turned up on the way. `# noqa: 402` in arrow_io.py and sio_io.py is missing the E, so flake8 was reading it as a blanket noqa and it only worked by accident. The `# pylint: disable=import-outside-toplevel` suppressions aren't readable by ruff, so they became `# noqa: PLC0415`. PLW2901 fires on two loop variables that pylint's redefined-loop-name doesn't flag, and since both are deliberate they got a noqa instead of a rewrite. The long lines in test_MemberParser.py are carried over as a per-file ignore, the same exemption `.flake8` already had.

Six files reformat, because podio pins black 23.11.0 and ruff format follows the newer black style. Two things worth flagging: dropping pylint loses the import-error check, since ruff doesn't resolve imports, and there are stale `# pylint: disable` comments in about ten files that are inert now. I left those for a separate PR rather than growing this one.

Ran `ruff check` and `ruff format --check` over all 55 files pre-commit looks at, including the extensionless scripts in tools/. Both clean.

BEGINRELEASENOTES
- Replace black, flake8 and pylint with ruff and ruff-format
- Move the python tool configuration into a single `.ruff.toml` and remove `.flake8`, `setup.cfg` and `pyproject.toml`
ENDRELEASENOTES

Fixes #703


tmadlener:
Can the format only changes be split off into a separate commit, such that we can more easily ignore them in `git blame`?

Compared to the changes in #1020 it looks like some of the checks from newer versions of flake are not yet taken up by this, or why do we not need annotations like `#noqa: A003` here (for example)?

mohitt31:
Split now, the reformatting is its own commit so its hash can go in `.git-blame-ignore-revs`.

A001 and A003 are flake8-builtins and I didn't select that group, I just took the list EDM4hep uses. Adding `A` finds one thing, the `copyright` in `doc/conf.py` that #1020 annotates, and A003 finds nothing at all. Say the word and I'll add it.

python/podio/arrow_io.py
```diff
@@ -4,10 +4,10 @@
 from ROOT import gSystem
 
 if gSystem.DynamicPathName("libpodioArrow.so", True):
-    gSystem.Load("libpodioArrow")  # noqa: 402
+    gSystem.Load("libpodioArrow")  # noqa: E402
 else:
     raise ImportError("Error when importing libpodioArrow")
-from ROOT import podio  # noqa: 402 # pylint: disable=wrong-import-position
+from ROOT import podio  # noqa: E402 # pylint: disable=wrong-import-position
```
tmadlener:
```suggestion
from ROOT import podio  # noqa: E402
```
Since we are no longer running pylint, we can probably also get rid of ignore statements.

.pre-commit-config.yaml
```diff
@@ -17,14 +13,14 @@ repos:
         exclude: (tests/(datamodel|src|extra_code)/.*(h|cc)$|podioVersion.in.h)
         types: [c++]
         language: system
-      - id: pylint
-        name: pylint
-        entry: 'pylint --rcfile=.github/scripts/pylint.rc --msg-template="{path}:{line}: [{msg_id}({symbol}), {obj}] {msg}"'
+      - id: ruff
+        name: ruff
+        entry: ruff check --force-exclude
```
tmadlener: Should the `force-exclude` go to `.ruff.toml`?

tmadlener (commented):
There are still a few minor things. In general, I think it's a good idea to homogenize here.

@jmcarcell any concerns from the point of view of the whole stack? I tihnk we mostly use ruff in any case already, right?

jmcarcell:
> There are still a few minor things. In general, I think it's a good idea to homogenize here.
> 
> @jmcarcell any concerns from the point of view of the whole stack? I tihnk we mostly use ruff in any case already, right?

No problem for the stack. It typically is at the latest version.

mohitt31:
Both done. `force-exclude` is in `.ruff.toml` now and out of the hook entries, and the pylint disables are gone across the tree, keeping the noqa where ruff still needs it. Two of them had a useful note attached, like gbl being a dynamic cppyy module, so I left that as a plain comment.

Removing the comments shortened some lines enough that ruff format reflowed them, that's in the same commit rather than the format-only one.

.pre-commit-config.yaml
```diff
@@ -17,14 +13,14 @@ repos:
...
+        name: ruff
+        entry: ruff check
         types: [python]
         language: system
-      - id: flake8
-        name: flake8
-        entry: 'flake8 --config=.flake8'
+      - id: ruff-format
+        name: ruff-format
+        entry: ruff format
```
tmadlener, line 23: Formatting should run before the linting, as otherwise format changes might move lint directives and break linting.

.ruff.toml
```diff
@@ -0,0 +1,19 @@
...
+
+[format]
+# Make things format the same way as black
+quote-style = "double"
+indent-style = "space"
+skip-magic-trailing-comma = false
+line-ending = "auto"
+
+[lint]
+select = ["F", "E", "W", "PLE", "PLW", "PLC"]
```
tmadlener, line 16: Does this truly cover everything that pylint did before?

We are dropping quite a fiew `pylint: disable-xxx` in the changes here and I am wondering whether we now just ignore them because ruff is more lenient, or whether we have dropped the checks entirely.

tmadlener (commented):
https://github.com/AIDASoft/podio/blob/master/cmake/pythonFormat.cmake also needs to be adapted, as it still uses the tools that we are removing here.

python/podio/arrow_io.py
```diff
@@ -4,14 +4,14 @@
 from ROOT import gSystem
 
 if gSystem.DynamicPathName("libpodioArrow.so", True):
-    gSystem.Load("libpodioArrow")  # noqa: 402
+    gSystem.Load("libpodioArrow")  # noqa: E402
```
tmadlener, line 7: Some of these `noqa: E402` should no longer be necessary, as `ruff` does allow imports after `if`s. Using `RUF100` for the link check should flag the ones that are not necessary.

mohitt31:
Sorry for the slow reply. Rebased onto master now that #1020 is in, and:

- pythonFormat.cmake uses ruff now (its flake8 target was pointing at the deleted .flake8)
- ruff-format runs before ruff check
- RUF100 is on and the unused noqa are gone, including the A003 ones from #1020. ruff's A003 doesn't fire on those lines even without the comment, so they'd stay unused if A were turned on.

On pylint: it doesn't fully cover it, some checks are dropped rather than just looser. E402 and PLC0415 are covered. Turning on the rest would give roughly naming (N) 20, docstrings (D) 19, import order (I001) 36, refactor (PLR) 28, protected-access (SLF001) 3, builtins (A) 1. import-error has no ruff equivalent.

I kept the select list matching EDM4hep. If you want any of those back, tell me which and I'll add them and fix what they flag.

PR #1019, opened 2026-09-04
fix: take the upstream EDM into account when reading old schemas

veprbl:
This should make OLD_DESCRIPTIONS usable for downstream models.

BEGINRELEASENOTES
- Take the upstream EDM into account when re-reading the old and current schemas, so that `OLD_DESCRIPTIONS` can be used together with `UPSTREAM_EDM`.

ENDRELEASENOTES


jmcarcell:
Can you expand on what is failing and how?

tmadlener (commented):
There is a bit of complexity here that I think needs some discussion and maybe a decision now. Which changes are allowed in this case for the upstream model? Does it have to be the same version for both versions of the consuming EDM? Or would one allow different versions also there? (That could potentially get arbitrarily complex).

PR #1020, opened 2026-09-06
Update pre-commit checks for latest LCG tools

jmcarcell:
The recent LCG stack update brings newer Black and Cppcheck versions.

BEGINRELEASENOTES
 - Fix pre-commit checks for the updated LCG tool stack.

ENDRELEASENOTES


include/podio/detail/LinkCollectionImpl.h
```diff
@@ -148,8 +148,8 @@ class LinkCollection : public podio::CollectionBase {
   /// @tparam FromU the type of the first object (**inferred!**)
   /// @tparam ToU   the type of the second object (**inferred!**)
   ///
-  /// @param from   the object to link (the *from* object in canonical order)
-  /// @param to     the object to link (the *to* object in canonical order)
+  /// @param source      the object to link (the *from* object in canonical order)
+  /// @param destination the object to link (the *to* object in canonical order)
```
tmadlener: Not sure I like this naming here. There is not really a *source* and a *destination* in the links here. Also at least to me they suggest that something is being either moved or copied from a *source* to a *destination*.

.cppcheck-suppressions.txt
```diff
@@ -6,6 +6,14 @@ missingIncludeSystem:*
 *:*Dict.cxx
 *:MurmurHash3.cpp
 preprocessorErrorDirective:cpython/
+syntaxError:RNTupleReader.h
+syntaxError:RNTupleWriter.h
```
tmadlener: What are the `syntaxError` that we are suppressing here?

.git-blame-ignore-revs
```diff
@@ -1 +1,2 @@
 518dae4cbcb411f1e0c1590edaf8919b3ce132f4
+22351f2b60cd0b46f4a686c730599a43e18324af
```
tmadlener, line 2: I think this one can only be done once we have landed these changes on the main branch. Otherwise we will just be updating this needlessly.

tmadlener (commented):
Since we are touching tooling here, should we consider #1018 fully and switch to ruff for the python parts?

jmcarcell:
> Since we are touching tooling here, should we consider #1018 fully and switch to ruff for the python parts?

Up to you, I'm fine with that, then the first commit can be removed.

.git-blame-ignore-revs
jmcarcell: Well you need two commits, assuming the formatting one is correct, then having this in the second commit makes it correct if not squashing.

.cppcheck-suppressions.txt
jmcarcell:
```
  failed to evaluate #if condition, undefined function-like macro invocation: ROOT_VERSION( ... )
```
because of 
```
#if ROOT_VERSION_CODE < ROOT_VERSION(6, 35, 0)
```
I don't know if there is a better way of fixing this, this requires reading `RVersion.h` from ROOT.

jmcarcell: Apparently you can pass macro definitions to cppcheck and this is now gone.

include/podio/detail/LinkCollectionImpl.h
jmcarcell: Suppression added instead

.git-blame-ignore-revs
m-fila: Doesn't github change SHA even with rebase and merge (from their [docs](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/about-merge-methods-on-github#rebasing-and-merging-your-commits))
> The rebase and merge behavior on GitHub deviates slightly from git rebase outside of GitHub. Rebase and merge on GitHub:
>    Always updates the committer information and creates new commit SHAs, whereas git rebase does not change the committer information when the rebase happens on top of an ancestor commit.

jmcarcell: "on top of an ancestor commit" but CI is not green if the branch is not up to date. I'll just tell AI to put the right one when bringing it up to date.

m-fila: "on top of an ancestor commit" was regarding the `git rebase` not the "github" rebase.

jmcarcell: I was sure I have done this in the past, maybe it was in Gitlab.

tmadlener: I will just do this once the format commit has landed on the main branch and we can keep this out of this PR.

.pre-commit-config.yaml
```diff
@@ -29,7 +29,7 @@ repos:
         language: system
       - id: cppcheck
         name: cppcheck
-        entry: cppcheck --project=build/compile_commands.json --enable=all --std=c++20 --cppcheck-build-dir=build --config-exclude="build/src/" --suppressions-list=.cppcheck-suppressions.txt -i "build/_deps/*" --error-exitcode=1 --check-level=exhaustive -j 4
+        entry: cppcheck --project=build/compile_commands.json --enable=all --std=c++20 --cppcheck-build-dir=build --config-exclude="build/src/" --suppressions-list=.cppcheck-suppressions.txt -DROOT_VERSION(major,minor,patch)=(((major)<<16)+((minor)<<8)+(patch)) -i "build/_deps/*" --error-exitcode=1 --check-level=exhaustive -j 4
```
tmadlener, line 32:
```suggestion
        entry: cppcheck --project=build/compile_commands.json --enable=all --std=c++23 --cppcheck-build-dir=build --config-exclude="build/src/" --suppressions-list=.cppcheck-suppressions.txt -DROOT_VERSION(major,minor,patch)=(((major)<<16)+((minor)<<8)+(patch)) -i "build/_deps/*" --error-exitcode=1 --check-level=exhaustive -j 4
```

Or do we for now have to go with the lowest version that we support?

jmcarcell: No preference, I don't know how different the checks with a different standard. I guess it suggests much more of `std::transform` instead of loops or I remember that from other repositories.

tmadlener: It looks like there is no really easy way to have *optional* hooks with pre-commit. We could have a second `cppcheck` hook that runs with `c++23`, so that we could at least see in CI what the things are that it will flag.

One could do something like `entry: cppcheck [...]; exit 0` to force pre-commit to essentially ignore it. Not sure if want to go that way.

For now, I think we should keep c++20 as the default version for `cppcheck` and switch to c++23 when we also make that the minimal version of c++ that we support.

tmadlener (approved):

PR #1021, opened 2026-09-06
Use the portable CMake Threads target

jmcarcell:
Related to #907.

BEGINRELEASENOTES
 - Use CMake's portable Threads target for thread support.

ENDRELEASENOTES


tmadlener (commented):
With the `find_package(Threads)` in the top level, it could be removed here

https://github.com/AIDASoft/podio/blob/d5c9be3404782e61da5fe9a063c3980db5938211/tests/unittests/CMakeLists.txt#L43

jmcarcell:
> With the `find_package(Threads)` in the top level, it could be removed here
> 
> https://github.com/AIDASoft/podio/blob/d5c9be3404782e61da5fe9a063c3980db5938211/tests/unittests/CMakeLists.txt#L43

Done

tmadlener (approved):

PR #1023, opened 2026-09-13
feat: allow for no ROOT dependency builds

wdconinc:
Now that we have the possibility for backends like Arrow, there is no absolute requirement for ROOT in podio (was technically already the case with SIO). Default remains to require ROOT.

The python interfaces in podio use cppyy and ROOT dictionaries for introspection, so all python interfaces (even for SIO and Arrow) require ROOT. This would need something like pybind11 or nanobind to avoid the ROOT dependency.

Add one test matrix entry with ROOT off.

Needs:
- [ ] https://github.com/key4hep/EDM4hep/pull/521
- [x] https://github.com/eic/EDM4eic/pull/181

BEGINRELEASENOTES
- feat: allow for no ROOT dependency builds

ENDRELEASENOTES

Assisted by ~~Claude~~ actually Gemini this time.


python/CMakeLists.txt
```diff
@@ -37,40 +37,46 @@ install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/templates
...
+    )
 
-  option(PODIO_RUN_STRACE_TEST "Run strace test" OFF)
-  if(PODIO_RUN_STRACE_TEST)
-    if (NOT strace_program)
-      message(WARNING "strace not found, skipping test_strace")
-    else()
-      message(STATUS "strace found: ${strace_program}")
-      # Test if too many files are being opened (https://github.com/AIDASoft/podio/issues/770)
-      # When the issue happened, it was around 7k with the Key4hep stack and 20k with the LCG stack
```
wdconinc, line 56: Keep this comment in place.

tests/schema_evolution/code_gen/test_utilities.cmake
```diff
@@ -58,39 +58,41 @@ function(GENERATE_DATAMODEL test_case model_version)
...
   )
+  set_target_properties(${model_base} PROPERTIES LIBRARY_OUTPUT_DIRECTORY ${output_base})
   PODIO_ADD_ROOT_IO_DICT(${model_base}Dict ${model_base} "${headers}" ${output_base}/src/selection.xml
     OUTPUT_FOLDER ${output_base}
   )
 
-  # Make sure that each model can be "toggled" at runtime separately.
-  # Effectively amounts to moving the byproducts of the dictgen above to the
-  # appropriate place and making sure libraries are built into the right output
-  # directory
```
wdconinc, line 68: Keep this comment in place.

tests/schema_evolution/code_gen/test_utilities.cmake
```diff
@@ -58,39 +58,41 @@ function(GENERATE_DATAMODEL test_case model_version)
...
+    set_target_properties(${model_base}Dict PROPERTIES LIBRARY_OUTPUT_DIRECTORY ${output_base})
+    add_custom_command(TARGET ${model_base}Dict
+      POST_BUILD
+      BYPRODUCTS
+        ${output_base}/lib${model_base}Dict_rdict.pcm
+        ${output_base}/${model_base}DictDict.rootmap
 
-  # Make sure cmake is aware of how these files came to their final destination
-  # so that there is a chance of dependency tracking and we might avoid too
-  # frequent triggering of the dictgen step
```
wdconinc, line 86: Keep this comment in place.

tests/unittests/unittest.cpp
tmadlener, line 1: Looking at the different `#if`s in here, I think we should split this up into a smaller file and conditionally compile it rather than having this done via pre-processor directives. But that is also possible in another PR.

tests/unittests/CMakeLists.txt
```diff
@@ -42,7 +42,10 @@ endif()
 
 find_package(Threads REQUIRED)
 add_executable(unittest_podio unittest.cpp frame.cpp buffer_factory.cpp interface_types.cpp std_interoperability.cpp links.cpp test_json_output.cpp)
-target_link_libraries(unittest_podio PUBLIC TestDataModel ExtensionDataModel InterfaceExtensionDataModel PRIVATE Catch2::Catch2WithMain Threads::Threads podio::podioRootIO)
+target_link_libraries(unittest_podio PUBLIC TestDataModel ExtensionDataModel InterfaceExtensionDataModel PRIVATE Catch2::Catch2WithMain Threads::Threads)
```
tmadlener:
```suggestion
target_link_libraries(unittest_podio PUBLIC TestDataModel ExtensionDataModel InterfaceExtensionDataModel PRIVATE Catch2::Catch2WithMain Threads::Threads podio::podioIO)
```
should also work, I think. That would bring the things along it was compiled with.

.github/workflows/test.yml
tmadlener, line 1: Maybe a cmake pre-set would be easier to handle this?

cmake/podioMacros.cmake
```diff
@@ -232,6 +232,11 @@ endfunction()
 #    OUTPUT_FOLDER        OPTIONAL: The folder in which the output files have been placed by PODIO_GENERATE_DATAMODEL. Defaults to ${CMAKE_CURRENT_SOURCE_DIR}
 #---------------------------------------------------------------------------------------------------
 function(PODIO_ADD_ROOT_IO_DICT dict_name CORE_LIB HEADERS SELECTION_XML)
+if((DEFINED PODIO_ENABLE_ROOT AND NOT PODIO_ENABLE_ROOT) OR (NOT DEFINED PODIO_ENABLE_ROOT AND NOT ENABLE_ROOT))
+  message(STATUS "Not adding the ROOT dictionaries for ${CORE_LIB} because ROOT support is not enabled")
```
tmadlener:
```suggestion
  message(STATUS "Not adding the ROOT dictionaries for ${CORE_LIB} because ROOT support is not enabled for podio")
```
To make it a bit more obvious where this is coming from?

python/podio/__init__.py
```diff
@@ -2,17 +2,22 @@
...
-# Try to load podio, this is equivalent to trying to load libpodio.so and will
+# Try to load podio via ROOT, this is equivalent to trying to load libpodio.so and will
 # error if libpodio.so is not found but work if it's found
 try:
     from ROOT import podio  # noqa: F401
 except ImportError:
-    print("Unable to load podio, make sure that libpodio is in LD_LIBRARY_PATH")
-    raise
+    print("Unable to load podio via ROOT (ROOT may not be installed or enabled)")
+    pass
```
tmadlener, line 11: Is there really anything we can meaningfully do if we have no ROOT dictionaries? I don't fully get, why this should become a `pass` here, rather than still being a hard error

jmcarcell: If it's a `pass` it will fail in the next line since `frame.py` imports `cppyy` and `ROOT`.

tmadlener: Depending on the environment you try to do this, that might work though, right?

The current print statement is also at best misleading, because it points towards ROOT as an issue, where as more likely it's the podio build that is.

.github/workflows/test.yml
wdconinc: Added preset. Debatable if it's much easier since it requires RNTuple config avoidance for 104...

CMakeLists.txt
```diff
@@ -75,53 +76,62 @@ option(ENABLE_DATASOURCE "Build podio's ROOT DataSource" OFF)
...
-  list(APPEND root_components_needed ROOTNTuple)
-  set(root_min_version 6.32)
+if(ENABLE_RNTUPLE AND NOT ENABLE_ROOT)
+  message(FATAL_ERROR "ENABLE_RNTUPLE requires ENABLE_ROOT=ON")
 endif()
-if(ENABLE_DATASOURCE)
-  list(APPEND root_components_needed ROOTDataFrame)
+if(ENABLE_DATASOURCE AND NOT ENABLE_ROOT)
+  message(FATAL_ERROR "ENABLE_DATASOURCE requires ENABLE_ROOT=ON")
 endif()
```
tmadlener, line 84: This should probably be done via `cmake_dependent_option` (thanks Claude for pointing out that that exists). That would also make the handling in the workflows much easier, because we can just use `ENABLE_ROOT=OFF` to consistently also turn these two off simultaneously.

wdconinc: Yes, I considered this but we don't include the module yet, so I didn't want to change it too much. If you are ok with the added module include, I'll add it.

wdconinc: See 4f131793.

tmadlener:
It looks like one of the root dictionaries is still trying to be installed:

https://github.com/AIDASoft/podio/actions/runs/35885912750/job/107265873642?pr=1023#step:5:271
``` 
install TARGETS given target "podioArrowDict" which does not exist.
```

similar guards as for the rest should do the trick when applied here, I think

https://github.com/AIDASoft/podio/blob/65a6811bef57bad219545c7165dfe0d3fc1bdd59/src/CMakeLists.txt#L202-L204

CMakeLists.txt
wdconinc: Ah, and 370cc9b.

tests/unittests/unittest.cpp
wdconinc: Yeah, I'll punt this to a follow-up PR. It requires pulling out some helper functions into headers since they then are used in both `unittest.cpp` and `unittest_root.cpp`.

wdconinc: https://github.com/AIDASoft/podio/issues/1026

python/podio/__init__.py
```diff
@@ -2,17 +2,22 @@
...
 
-# Try to load podio, this is equivalent to trying to load libpodio.so and will
+# Try to load podio via ROOT, this is equivalent to trying to load libpodio.so and will
 # error if libpodio.so is not found but work if it's found
 try:
     from ROOT import podio  # noqa: F401
 except ImportError:
-    print("Unable to load podio, make sure that libpodio is in LD_LIBRARY_PATH")
-    raise
+    print("Unable to load podio via ROOT (ROOT may not be installed or enabled)")
```
tmadlener:
```suggestion
    print("Unable to load podio via ROOT. podio may have built without ROOT support or libpodio is not on LD_LIBRARY_PATH.")
```

python/podio/__init__.py
wdconinc: Thanks to @tmadlener for the error message suggestion. Are we ok then with pass and fail in next?

